PR #16, opened 2019-01-08
Compare parameter string to arg instead of args

HarrisonMc555:
Unless I'm mistaken, you should be comparing all parameter strings to `arg` (a String), not `args` (a String Array).


clavin-xlnx:
Thank you for your pull request!  It does appear to be a bug you have found and I appreciate your efforts to fix it.  We still need to get a CLA in place before we can accept pull requests, but I hope to remedy that problem soon.

PR #43, opened 2019-08-14
Update PerformanceExplorer.java

LJChen98:
In Windows OS, running PerformanceExplorer would encounter a problem; for file separator in Windows is "\" while in Vivado Tcl is “/”. Therefore, the Tcl script created by PerformanceExplorer in Windows OS would fail in Vivado. 
I suppose the simplest way to solve this problem would be making a replacement of all "\" in the Tcl script to "/".


clavin-xlnx:
Thank you for the fix.  I can see that one of the parameters to the function could potentially have '\\' in it as well and this is a way to ensure they all get switched to '/'.  Since this is a trivial fix, I don't think we need to require you to sign the contributor license agreement.

PR #57, opened 2020-02-27
TimingGroup: Make 'add' functions public

daveshah1:
I am working on integrating RapidWright timing with nextpnr. AIUI, there is no way at the moment to work out timing for an incomplete part of routing, which would be needed to provide some kind of timing model in the nextpnr database for the router to use and these functions seem to be helpful for. I appreciate this may not end up being as accurate as doing it on full nets, but this is only to provide a guide for the nextpnr router with Vivado still useful for signoff timing.

### DIFF
--- a/src/com/xilinx/rapidwright/timing/TimingGroup.java
+++ b/src/com/xilinx/rapidwright/timing/TimingGroup.java
@@ -200,7 +200,7 @@ public class TimingGroup implements Comparable<TimingGroup> {
      * @param n Node to be added.
      * @param c IntentCode is the Vivado-assigned Type for the given node.
      */
-    protected void add(Node n, IntentCode c) {
+    public void add(Node n, IntentCode c) {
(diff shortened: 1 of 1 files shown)


clavin-xlnx:
Thank you for the pull request. Yes your assessment is spot on.  The timing model is the ground work and a next step would be to build a router timing estimation database/cache to be used to estimate the delay of the partially routed endpoint to the un-routed sink while routing a net.  This is on the roadmap, but I can't say when it will be available quite yet.

daveshah1:
As a follow-up, what do you think about making DelayModel public too? If you'd rather sort out the interface issue mentioned in the TODO first (getting a list of arcs rather than having to know what arcs exist a priori), I can have a crack at that?

clavin-xlnx:
I went ahead and made DelayModel public.  If you'd like to take a look at improving the interface, we welcome your input and support!

PR #58, opened 2020-03-03
Fixed file naming issues when having multiple instances of an IP

ElaHobby:
The issue occurs in case of designs having more instances for an IP, when RW copies into IP_CACHE the already generated synthesis files. The order in which these files are copied determine the name of the .dcp files. This order does not correspond with the one that RW expects when generating opt runs.


src/com/xilinx/rapidwright/design/blocks/PBlockGenerator.java
```diff
@@ -68,7 +68,7 @@
 	
 	/** X dimension over Y dimension */
 	public float ASPECT_RATIO = 0.125f;//1.5f;	
-	public float OVERHEAD_RATIO = 1.5f;//1.25f;
+	public float OVERHEAD_RATIO = 4.5f;//1.25f;
```
clavin-xlnx: @ElaHobby  - I think we probably want to keep the OVERHEAD_RATIO default to 1.5.  I'm sure you have found instances where it has needed to be increased, but we probably aren't ready to change it for everyone.

src/com/xilinx/rapidwright/design/blocks/PBlockGenerator.java
```diff
@@ -940,16 +940,57 @@ public static void main_testing(String[] args) {
...
+                
 			}
 			if(numBRAMColumns > 0){
 				int pIdx = 0;
 				for(int i=0; pIdx < p.size(); i++){
 					TileTypeEnum t = dev.getTile(row, col+i).getTileTypeEnum();
 					if(Utils.isBRAM(t)){
 						for(Site s : dev.getTile(row, col+i).getSites()){
-							if(s.getSiteTypeEnum() == SiteTypeEnum.RAMBFIFO36) upperLeft = s;
+							if((s.getSiteTypeEnum() == SiteTypeEnum.RAMBFIFO36)||(s.getSiteTypeEnum() == SiteTypeEnum.RAMBFIFO36E1)) upperLeft = s; // Update. Goal: support for 7 series
```
clavin-xlnx: Thanks, this change looks good.  I will follow up later with a better check for BRAMs.

src/com/xilinx/rapidwright/design/blocks/PBlockGenerator.java
```diff
@@ -940,16 +940,57 @@ public static void main_testing(String[] args) {
 					}
 					if(p.get(pIdx) == t) pIdx++;
 				}
-				sb.append("SLICE_X" + upperLeft.getInstanceX() + "Y" + (upperLeft.getInstanceY()-(numSLICERows-1)) + 
-						 ":SLICE_X" + (upperLeft.getInstanceX()+(numSLICEColumns+numSLICEMColumns)-1) + "Y" + upperLeft.getInstanceY());
+
```
clavin-xlnx: This update is clearly attempting to address the issue of the pblock colliding with the PS which is not a homogeneous--this is a problem.  Since we have a number of situations similar to this, perhaps its best if we hold off on these point solutions and work together on a more holistic approach.

tcl/rapidwright.tcl
```diff
@@ -53,7 +53,12 @@ proc compile_block_dcp { dcpFile } {
     puts "RAPIDWRIGHT_PATH=$rwpath"
     puts "CLASSPATH=$cpath"
     exec java com.xilinx.rapidwright.util.Unzip ${dcpFile} ${unzipDir}
-    read_xdc ${unzipDir}/${rootDcpFileName}_in_context.xdc
+    # Updated. Goal: avoid naming problems caused by the fact that the files copied into IP_CACHE have different names as the ones expected by RW. Error appears only in designs with multiple IPs with the same ID
+	# Line bellow removed and replaced by the following 2 lines
+    #read_xdc ${unzipDir}/${rootDcpFileName}_in_context.xdc
+	set file_name_xdc [glob -directory ${unzipDir} *_in_context.xdc]
```
clavin-xlnx: This looks like it should work.  I'm ok with this change, you can delete the old line as it stays in the history of git.

tcl/rapidwright.tcl
```diff
@@ -662,3 +673,25 @@ proc offset_dsps { count } {
...
+    set existingIP [string map {"</spirit:instanceName>" ""} $existingIP]
+    set existingIP [string trim $existingIP]
+	set ip_return [get_ips $existingIP]
+	if {$ip_return!={}} {
+		return $ip_return
+	} else {
+		return $ip
+	}
+}
+# End of updated
```
clavin-xlnx: Looks good, perhaps just add a newline at the end of the file.

PR #60, opened 2020-03-27
Horizontal density

ElaHobby:
Added the following features: 
- algorithm for higher horizontal density = pblocks will try to use different columns
- algorithm for computing more compact pblocks (for exp. clbs closer to brams in case bigger patterns are chosen)
- new algorithm selects the slice from a tile depending on whether it is a _R or _L clb
- new algorithm selects the pblock width having in mind also 7. series devices which have 2 slices/tile

### DIFF
--- a/src/com/xilinx/rapidwright/design/blocks/PBlockGenerator.java
+++ b/src/com/xilinx/rapidwright/design/blocks/PBlockGenerator.java
@@ -926,31 +926,38 @@ public class PBlockGenerator {
 		ArrayList<String> pBlocks = new ArrayList<String>(PBLOCK_COUNT);
 		
 		// Code inserted to obtain the pblock pattern having the highest number of free resources on the device
-		boolean do_horiz_dens = !GLOBAL_PBLOCK.contentEquals("");
-		TreeMap<Float,TileColumnPattern> store_best_pattern =new TreeMap<Float,TileColumnPattern> () ; // Store an ordered list of the patterns. The order is given by the number of free resources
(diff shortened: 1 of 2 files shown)


clavin-xlnx:
Thank you for the updates and features added.  To work towards accepting the pull request, one test that would need to pass is to run the IP Integrator tutorial:

http://www.rapidwright.io/docs/IPI_PreImpl_Tutorial.html

I tried running it and it failed with something like:

`can't read "global_pblock_command": no such variable`

Also, if you could refactor (rename) your methods and variables added to follow a lowerCamelCase style, it would be much appreciated (see https://github.com/Xilinx/RapidWright/blob/master/CODING_STYLE.md).

Thank you!

ElaHobby:
The mentioned error is described in the initial mail. It is about choosing how to insert the new feature in the flow (rapidwright.tcl). Bellow again its content:
-	Rapidwright.tcl: how should we insert the new feature in the flow? I was thinking of checking the value of a GLOBAL_PBLOCK env. variable. If this exists, than the horiz_density algorithm shall be run, else, the original one. This is the commented code in "compile_block_dcp" function, before calling the PBlockGenerator. While testing the functionality of the new algorithm, the -p option was inserted manually. However this shall not remain like this in the merged version. This aspect shall be updated before runing the code, otherwise the code won't run!
 -	In function “create_allCLB_PBlocks”, when I select the slice of the tile, I wrote the code having in mind 7. Series. This might not work for Ultrascale though, didn't check

PR #61, opened 2020-04-02
Fix jar build issues

moxyll:
[This post](https://discuss.gradle.org/t/gradle-hangs-and-create-inifinitely-long-jar/653) indicated that the build hang might be because multiple sections had the same output directories. I changed the java compilation to go into `compiled` instead of `bin`, and the jar file builds instantly!

I can easily change the target directory name if you don't like `compiled`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ project.buildDir = 'bin'
 
 sourceSets {
   main {
-    java.outputDir = file('compiled')
+    java.outputDir = file('bin/compiled')
     java {
       srcDirs = ['src']
     }    
--- a/build.gradle
+++ b/build.gradle
@@ -1,11 +1,11 @@
 apply plugin: 'java'
-jar.enabled = false
+jar.enabled = true
 
 project.buildDir = 'bin'
 
 sourceSets {
   main {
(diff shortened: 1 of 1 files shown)


clavin-xlnx:
Thank you for the work-around!  Yes, gradle had worked to build a jar in the past with a common `bin` directory, but at some version, the behavior changed (or a bug was introduced).  It seems gradle depends on a certain directory structure which RapidWright does not adhere to.  

I would pull your request, except I it would break a few other places that depend on the compiled class files being in `bin` rather than `compiled`.  If we can find a way where all class files from the Java source stay in the `bin` directory, that would be ideal.  

I am open to suggestions.

moxyll:
Would it work for you if the compiled classes were in `bin/compiled`? That seems to work. The paths end up as `bin/compiled/com/xilinx/rapidwright/...`

clavin-xlnx:
I'm sorry, the CLASSPATH for compiled RapidWright code in a several places points to ``bin``.  Perhaps we can explore another option for generating jar files for RapidWright.  I am currently exploring other tools for compilation that perhaps could help close the gap here.

PR #62, opened 2020-04-23
Ensure that highlighted tile numbers are drawn above tile highlighting

mortbopet:
Even if the text item is added to the scene after 'this' (the tile
highlighting), It may happen that the text item is drawn before the tile
highlighting. As such, the text will not be visible.

Usually, this would be automatically handled if the text item is created
as a child of 'this' (the tile highlighting).
However, this does not seem to be working in QtJambi and the text item
has to be created without a parent and manually added to the scene.
Instead, we manually set the Z value of the text item to be greater than
the highlighting, to ensure that it is always drawn above the highlighting.

Signed-off-by: Morten Borup Petersen <morten_bp@live.dk>


clavin-xlnx:
Thanks for the fix!

PR #68, opened 2020-06-25
Fixed getLUTSize. Returns actual size instead of ASCII value

jojokl:
Instead of the actual size the ASCII value of the digit was used (for instance 51 instead of 3). So the configuration of LUTs was very slow because the equation was evaluated 2^51 times instead of 2^3 times.
This was fixed by not using the ASCII character value but to parse the value correctly.


clavin-xlnx:
Thanks for the fix!

PR #81, opened 2020-10-06
Add constant net check.

litghost:
@clavin-xlnx Can you run this through your test set, and see if it explodes.  I've done some initial tests and it seems to be catch the problem I wanted to detect.

### DIFF
--- a/src/com/xilinx/rapidwright/interchange/PhysNetlistReader.java
+++ b/src/com/xilinx/rapidwright/interchange/PhysNetlistReader.java
@@ -5,7 +5,10 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.ArrayDeque;
+import java.util.Queue;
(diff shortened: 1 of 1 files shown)


clavin-xlnx:
@litghost, I ran through the test set, no explosions so looks good.  One request, would you mind renaming variables to use camelCase?  If not, I can easily do it after the merge.  Thank you!

litghost:
> One request, would you mind renaming variables to use camelCase? If not, I can easily do it after the merge. Thank you!

Ya sure.

litghost:
> @litghost, I ran through the test set, no explosions so looks good. One request, would you mind renaming variables to use camelCase? If not, I can easily do it after the merge. Thank you!

This has been fixed.  Something to consider is makefile targets for auto-formatting and checking that formatting.  Does Java have a good auto-formatter?

clavin-xlnx:
I need to look into finding a compatible auto-formatter, that is a good suggestion.

PR #83, opened 2020-10-14
Add some simple error checking on PIP path.

litghost:
@clavin-xlnx Is there a better way to detect a bad PIP besides catching the NPE from `new PIP`?

### DIFF
--- a/src/com/xilinx/rapidwright/interchange/PhysNetlistReader.java
+++ b/src/com/xilinx/rapidwright/interchange/PhysNetlistReader.java
@@ -359,7 +359,25 @@ public class PhysNetlistReader {
                 Tile tile = device.getTile(strings.get(pReader.getTile()));
                 String wire0 = strings.get(pReader.getWire0());
                 String wire1 = strings.get(pReader.getWire1());
-                PIP pip = new PIP(tile, wire0, wire1);
+                if(tile == null) {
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/interchange/PhysNetlistReader.java
```diff
@@ -359,7 +359,19 @@ private static void readRouteBranch(RouteBranch.Reader branchReader, Net net, De
                 Tile tile = device.getTile(strings.get(pReader.getTile()));
                 String wire0 = strings.get(pReader.getWire0());
                 String wire1 = strings.get(pReader.getWire1());
-                PIP pip = new PIP(tile, wire0, wire1);
+                if(tile == null) {
```
clavin-xlnx, line 362: @litghost  - Yes, this set of APIs will get around that issue:
```
        Integer wire0Idx = tile.getWireIndex(wire0);
        Integer wire1Idx = tile.getWireIndex(wire1);
        PIP pip = tile.getPIP(wire0Idx, wire1Idx);
```

clavin-xlnx: See also: http://www.rapidwright.io/javadoc/com/xilinx/rapidwright/device/Tile.html#getPIP-int-int-

litghost: Right, fixed.

clavin-xlnx (approved):

PR #86, opened 2020-10-21
Fix various bugs in device resource generation

litghost:
The schema indicates that the wire table is StringIdx, StringIdx: 

https://github.com/Xilinx/RapidWright/blob/58f91e011e449be78c09b3f4302b341396c7b57c/interchange/DeviceResources.capnp#L105-L108

This fixes the DeviceResourceWriter to match the schema.

### DIFF
--- a/src/com/xilinx/rapidwright/interchange/DeviceResourcesVerifier.java
+++ b/src/com/xilinx/rapidwright/interchange/DeviceResourcesVerifier.java
@@ -17,6 +17,7 @@ import org.capnproto.TextList;
 import org.capnproto.StructList.Reader;
 
 import com.xilinx.rapidwright.design.Design;
+import com.xilinx.rapidwright.design.SiteInst;
 import com.xilinx.rapidwright.device.BEL;
 import com.xilinx.rapidwright.device.BELPin;
 import com.xilinx.rapidwright.device.Device;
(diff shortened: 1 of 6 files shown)


litghost:
As part of expanding the FPGA interchange Python library, I identified several issues with the device resources database.

src/com/xilinx/rapidwright/interchange/DeviceResourcesWriter.java
```diff
@@ -241,8 +241,8 @@ public static void writeAllSiteTypesToBuilder(Device device, DeviceResources.Dev
                 SitePin.Builder pin = pins.get(j);
                 pin.setName(allStrings.getIndex(pinNames.get(j)));
                 pin.setDir(j <= highestIndexInputPin ? Direction.INPUT : Direction.OUTPUT);
-                Integer siteWireIdx = site.getSiteWireIndex(pinNames.get(j));
-                pin.setSitewire(siteWireIdx == null ? -1 : siteWireIdx);
+                String tileWireName = site.getTileWireNameFromPinName(pinNames.get(j));
```
clavin-xlnx, line 245: You are correct in that this is not writing out the site wire name associated with the site pin.  However, the `Site.getTileWireNameFromPinName()` method will return the tile wire from the context of the site which does not match the schema in this situation.  Because Vivado (and thus RapidWright) do not have SiteType objects, we are using Site objects.  But we are writing out a SiteType object.  We want the site wire name, which could be gotten by:

```
                Integer siteWireIdx = site.getSiteWireIndex(pinNames.get(j));
                String siteWireName = siteWireIdx == null ? null : site.getSiteWireName(siteWireIdx);
                pin.setSitewire(siteWireName == null ? -1 : allStrings.getIndex(siteWireName));
```

src/com/xilinx/rapidwright/interchange/DeviceResourcesVerifier.java
```diff
@@ -224,8 +224,8 @@ private static boolean verifySiteType(Device device, DeviceResources.Device.Read
             if( (isInput != (dir == Direction.INPUT)) || (isOutput != (dir == Direction.OUTPUT)) ){
                 throw new RuntimeException("ERROR: Mismatch on site pin direction");
             }
-            Integer siteWireIndex = site.getSiteWireIndex(pinName);
-            expect(siteWireIndex == null ? -1 : siteWireIndex, spReader.getSitewire());
+            String tileWireName = site.getTileWireNameFromPinName(pinName);
```
clavin-xlnx, line 227: See my other comment from DeviceResourcesWriter.java:

Corresponding changes to match suggested changes would be:
```
            Integer siteWireIndex = site.getSiteWireIndex(pinName);
            String siteWireName = siteWireIndex == null ? null : site.getSiteWireName(siteWireIndex);
            int siteWireNameIdx = spReader.getSitewire();
            expect(siteWireName == null ? null : siteWireName, siteWireNameIdx == -1 ? null : allStrings.get(siteWireNameIdx));
```
Also, because the siteWire associated with a site pin can be null, we have to accommodate that scenario.  The `expect()` method would also need to be modified by checking for null equality:

```
    private static boolean expect(String gold, String test) {
        if(gold == null && test == null) return true;
        //...
```

src/com/xilinx/rapidwright/interchange/DeviceResourcesWriter.java
```diff
@@ -238,11 +238,13 @@ public static void writeAllSiteTypesToBuilder(Device device, DeviceResources.Dev
             siteType.setLastInput(highestIndexInputPin);
             StructList.Builder<SitePin.Builder> pins = siteType.initPins(pinNames.size());
             for(int j=0; j < pinNames.size(); j++) {
+                com.xilinx.rapidwright.device.SitePin site_pin = new com.xilinx.rapidwright.device.SitePin(site, pinNames.get(j));
```
clavin-xlnx, line 241: This could be simplified with:
```
                BELPin belPin = site.getBELPin(pinNames.get(j));
                pin.setBelpin(allBELPins.getIndex(belPin));
```

clavin-xlnx (commented):
@litghost - Thank you for identifying these issues and working to resolve them.  I believe one can move forward without any changes, but the other two will need some adjustment.

src/com/xilinx/rapidwright/interchange/DeviceResourcesWriter.java
litghost: Ya, fixed by removing Sitewire, as it can be reached `SiteWire`.

src/com/xilinx/rapidwright/interchange/DeviceResourcesVerifier.java
litghost: Fixed

src/com/xilinx/rapidwright/interchange/DeviceResourcesWriter.java
litghost: Thanks, done.

litghost:
I gave this PR a big iteration to fix alternative site types.

litghost:
I've tested these changes in the FPGA interchange library using the device resources to verify routing trees successfully.  We might need to iteration the structure on the alternative site types site.

src/com/xilinx/rapidwright/interchange/DeviceResourcesWriter.java
```diff
@@ -49,14 +51,36 @@
     private static Enumerator<String> allStrings;
     private static Enumerator<String> allSiteTypes;
 
-    private static HashMap<TileTypeEnum,Tile> tileTypes; 
+    private static HashMap<TileTypeEnum,Tile> tileTypes;
     private static HashMap<SiteTypeEnum,Site> siteTypes;
-    
-    public static void populateEnumerations(Device device) {
+
+    public static void populateSiteEnumerations(SiteInst site_inst, Site site) {
```
clavin-xlnx, line 57: @litghost - Thank you for all the updates, looks good.  

Not a big issue, but its nice to match the camelCase variable naming convention.   

At your suggestion, I did try looking into code style checkers.  The [Google code style checker for Java](https://github.com/google/google-java-format) seemed popular, but it doesn't check/correct naming schemes/conventions.  Are there tools that enforce/report variable naming conventions (camelCase vs. underscore) in any language?   

Eventually, I need to update the coding style to disallow tabs and purge the codebase of them.

litghost: Ya, I'll fix up the under_score.  My muscle memory is `_`, but I don't mind fixing it.  This is one place where either a linting pass to detect or even an auto-format if possible is handy.

litghost: `_` fixed

PR #87, opened 2020-10-21
7 series PBlock generation

ElaHobby:
The pull request: 
-> takes into account that 7 series and Ultrascale devices have a different number of CARRY_PER_CLE, RAMLUTS_PER_CLE, 	SLICES_PER_TILE.
-> considers that FF & Carry can be placed in a SLICE together with LUTS (including LUTRAM) => different method of computing required nr. of Slices
-> in case of horizontal density algorithm, it adds some checks to avoid endless loop in special cases. For example, if provided patterns are unfeasible (unfeasible meaning cases like dev.getSite("SLICE_X"+xr+"Y"+yd)=null)

### DIFF
--- a/src/com/xilinx/rapidwright/design/blocks/PBlockGenerator.java
+++ b/src/com/xilinx/rapidwright/design/blocks/PBlockGenerator.java
@@ -164,17 +164,14 @@ public class PBlockGenerator {
 			RAMLUTS_PER_CLE = 4;
 			SLICES_PER_TILE = 2;
 		}
-		//if(debug){ // ELA
+		if(debug){ 
 			System.out.println("Parsed report: " + reportFileName);
 			System.out.println("  Device: " + dev.getName());
-			System.out.println("Ela    CARRY: " + carryCount);
-			System.out.println("Ela    lutRAMCount: " + lutRAMCount);
-			System.out.println("Ela  regCount:  " + regCount);
 			System.out.println("    LUTs: " + lutCount);
 			System.out.println("    DSPs: " + dspCount);
 			System.out.println("18kBRAMs: " + bram18kCount);
 			System.out.println("36kBRAMs: " + bram36kCount);
-		//} // ELA
+		} 
 		if(lutCount < LUTS_PER_CLE) lutCount = LUTS_PER_CLE;
 	}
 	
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/design/blocks/PBlockGenerator.java
```diff
@@ -157,6 +158,10 @@ private void getResourceUsages(String reportFileName){
 			}
 		}
 		
+		if (dev.getName().contains("xc7")) {
```
clavin-xlnx, line 162: Some Series 7 devices don't actually match.  Try this instead:

`if (dev.getSeries() == Series.Series7) {`

src/com/xilinx/rapidwright/design/blocks/PBlockGenerator.java
```diff
@@ -974,38 +979,44 @@ public static void main_testing(String[] args) {
...
-					// Also, current solution computes free resources only in case of CLB pblocks. The BRAMS and DRAMs columns are selected to match the most suitable clb pblock
-					List<String> WritePBlocks = new ArrayList<String>();
-					for (Integer i: CLBPBlock.keySet()) {
-						WritePBlocks.add("SLICE_X" + CLBPBlock.get(i)[0] + "Y" + CLBPBlock.get(i)[2] + ":SLICE_X" + CLBPBlock.get(i)[1] + "Y" + CLBPBlock.get(i)[3]);
-					} 
-					if(IP_NR_INSTANCES==0) {
-						if (debug)
-							System.out.println(" CRITICAL WARNING: IP_NR_INSTANCES is 0! Default value 1 was set.");
-						IP_NR_INSTANCES = 1;
+					if(CLBPBlock.size()>0) { // If no feasible CLB PBlock found for this pattern, skip it!
```
clavin-xlnx, line 994: Is there a difference here?  If it is just whitespace difference, do you mind reverting this block to the original code?

clavin-xlnx (changes requested):
Hi @ElaHobby - thank you for the pull request.  I've made a few comments, please take a look and I believe we can move forward after just a few minor changes.  Thanks!

src/com/xilinx/rapidwright/design/blocks/PBlockGenerator.java
ElaHobby: I added an if-clause (line 994 + 1025-1028) to avoid error if no feasible* PBlocks are found for a pattern. Thus, the indentation is different and probably that's why github signals this as different code.
* - probability is really low. However, I had one design where I got error when reading CLBPBlock.get(0)[0];. The createAllPBlocks() function checks if PBlock corners.getSite return null. When I got the error, the corresponding pattern had only 1 possible PBlock and one of the corners were not a valid Site.  As said, probability is quite low in my opinion, but I introduced the additional if-clauses as safety net.

src/com/xilinx/rapidwright/design/blocks/PBlockGenerator.java
```diff
@@ -159,7 +159,7 @@ private void getResourceUsages(String reportFileName){
 			}
 		}
 		
-		if (dev.getName().contains("xc7")) {
+		if (if (dev.getSeries() == Series.Series7) {
```
clavin-xlnx, line 162: @ElaHobby - Sorry, but it looks like this change does not compile.

PR #92, opened 2020-11-10
Draft changes for pseudo PIP blockage data.

clavin-xlnx:
Minimal pseudo PIP blockage map data.

### DIFF
--- a/interchange/DeviceResources.capnp
+++ b/interchange/DeviceResources.capnp
@@ -61,10 +61,11 @@ struct Device {
   }
 
   struct TileType {
-    name      @0 : StringIdx;
-    siteTypes @1 : List(SiteTypeInTileType);
-    wires     @2 : List(StringIdx);
-    pips      @3 : List(PIP);
+    name       @0 : StringIdx;
+    siteTypes  @1 : List(SiteTypeInTileType);
+    wires      @2 : List(StringIdx);
+    pips       @3 : List(PIP);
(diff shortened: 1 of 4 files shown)


clavin-xlnx:
@litghost - please take a look when it is convenient.  This is a minimal pseudo PIP blockage information map.  Perhaps there are other details you'd like to see added?

interchange/DeviceResources.capnp
```diff
@@ -61,10 +61,11 @@ struct Device {
...
   struct TileType {
-    name      @0 : StringIdx;
-    siteTypes @1 : List(SiteTypeInTileType);
-    wires     @2 : List(StringIdx);
-    pips      @3 : List(PIP);
+    name       @0 : StringIdx;
+    siteTypes  @1 : List(SiteTypeInTileType);
+    wires      @2 : List(StringIdx);
+    pips       @3 : List(PIP);
+    pseudoPIPs @4 : List(PseudoPIP);
```
litghost, line 68: Are all `PseudoPIP` duplicated in the `PIP` array?  Does that make sense?  Should the `PseudoPIP` just be part of `PIP` as a `union { void / pseudoCells }`?

clavin-xlnx: Thanks @litghost , that is a good point.  I have moved PseudoCells under PIP combined with a union with void.

litghost (approved):
LGTM

PR #94, opened 2020-11-11
Add Cell BEL pin mapping schema.

litghost:
I'm working on adding logic to `DeviceResourceWriter` / `DeviceResourceVerifier` to populate this.


clavin-xlnx:
@litghost - Not quite sure how all the entry/mapping relationships work, but seems flexible enough.

PR #98, opened 2020-11-18
Add schema for describing local site inverters.

litghost:
This adds a schema description for local site inverters, how to control them, and how to route them.  This is a peer to the inverting property on `SitePips` in the `PhysicalNetlist`.

### DIFF
--- a/interchange/DeviceResources.capnp
+++ b/interchange/DeviceResources.capnp
@@ -25,7 +25,7 @@ struct Device {
   primLibs        @7 : Dir.Netlist; # Netlist libraries of Unisim primitives and macros
   exceptionMap    @8 : List(PrimToMacroExpansion); # Prims to macros expand w/same name, except these
   cellBelMap      @9 : List(CellBelMapping);
-  cellInversions @10 : List(CellInversions);
+  cellInversions @10 : List(CellInversion);
   packages       @11 : List(Package);
   constants      @12 : Constants;
 
@@ -335,7 +335,7 @@ struct Device {
     inverting    @2 : CellPinInversionParameter;
   }
 
-  struct CellInversions {
+  struct CellInversion {
     # Which cell is being described?
     cell     @0 : StringIdx;
 
--- a/interchange/DeviceResources.capnp
+++ b/interchange/DeviceResources.capnp
@@ -25,7 +25,7 @@ struct Device {
   primLibs        @7 : Dir.Netlist; # Netlist libraries of Unisim primitives and macros
   exceptionMap    @8 : List(PrimToMacroExpansion); # Prims to macros expand w/same name, except these
   cellBelMap      @9 : List(CellBelMapping);
-  cellInversions @10 : Void;
+  cellInversions @10 : List(CellInversions);
   packages       @11 : List(Package);
   constants      @12 : Constants;
 
@@ -76,11 +76,20 @@ struct Device {
   #######################################
   # Placement instance objects
   #######################################
+  struct BELInverter {
+    nonInvertingPin @0 : BELPinIdx;
+    invertingPin    @1 : BELPinIdx;
+  }
+
   struct BEL {
     name      @0 : StringIdx;
     type      @1 : StringIdx;
     pins      @2 : List(BELPinIdx);
     category  @3 : BELCategory; # This would be BELClass/class, but conflicts with Java language
+    union {
+        nonInverting @4 : Void;
+        inverting    @5 : BELInverter;
+    }
   }
 
   enum BELCategory {
@@ -293,5 +302,44 @@ struct Device {
     # nodeSources should be used to explicitly list nodes that fall into this
     # case.
     nodeSources            @3 : List(NodeConstantSource);
+  }
+
+  ######################################
+  # Inverting pins description
+  #
+  # This block describes local site wire
+  # inverters, site routing BELs, and
+  # parameters.
+  ######################################
+  struct CellPinInversionParameter {
+    union {
+      # This inverter cannot be controlled by parameter, only via tool merging
+      # of INV cells or other means.
+      invOnly      @0 : Void;
+      # This inverter can be controlled by a parameter.
+      # What parameter value configures this setting?
+      parameter    @1 : Dir.Netlist.PropertyMap.Entry;
+    }
+  }
+
+  struct CellPinInversion {
+    # Which cell pin supports a local site inverter?
+    cellPin      @0 : StringIdx;
+
+    # What parameters are used for the non-inverting case, and how to route
+    # through the inversion routing bels (if any).
+    notInverting @1 : CellPinInversionParameter;
+
+    # What parameters are used for the inverting case, and how to route
+    # through the inversion routing bels (if any).
+    inverting    @2 : CellPinInversionParameter;
+  }
+
+  struct CellInversions {
+    # Which cell is being described?
+    cell     @0 : StringIdx;
+
+    # Which cell have site local inverters?
+    cellPins @1 : List(CellPinInversion);
   }
 }
--- a/src/com/xilinx/rapidwright/interchange/DeviceResources.java
+++ b/src/com/xilinx/rapidwright/interchange/DeviceResources.java
@@ -5,7 +5,7 @@ package com.xilinx.rapidwright.interchange;
 
 public final class DeviceResources {
   public static class Device {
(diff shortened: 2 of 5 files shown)


litghost:
Example data, xc7:
```
- cell: BUFGCTRL
 - cellPins:
  - cellPin: CE0
    notInverting:
     parameter:
      key: IS_CE0_INVERTED
      textValue: 0
     routingBels:
      - routingBel: CE0INV
        belPin: CE0
    inverting:
     parameter:
      key: IS_CE0_INVERTED
      textValue: 1
     routingBels:
      - routingBel: CE0INV
        belPin: CE0_B
- cell: RAMB36E1
 - cellPins:
  - cellPin: ENBWREN
    notInverting:
     parameter:
       key: IS_ENBWREN_INVERTED
       textValue: 0
     routing_bels:
      - routingBel: ENBWRENUINV
        bel_pin: ENBWRENU
      - routingBel: ENBWRENLINV
        bel_pin: ENBWRENL
    inverting:
     parameter:
       key: IS_ENBWREN_INVERTED
       textValue: 1
     routing_bels:
      - routingBel: ENBWRENUINV
        bel_pin: ENBWRENU_B
      - routingBel: ENBWRENLINV
        bel_pin: ENBWRENL_B
```

xcu:
```
- cell: BUFGCTRL
 - cellPins:
  - cellPin: CE0
    notInverting:
     parameter:
      key: IS_CE0_INVERTED
      textValue: 0
     routingBels:
      - routingBel: CE0INV
        belPin: CE0
    inverting:
     parameter:
      key: IS_CE0_INVERTED
      textValue: 1
     routingBels:
      - routingBel: CE0INV
        belPin: CE0
- cell: RAMB36E1
 - cellPins:
  - cellPin: ENBWREN
    notInverting:
     parameter:
       key: IS_ENBWREN_INVERTED
       textValue: 0
     routing_bels:
      - routingBel: OPTINV_ENBWREN_U
        bel_pin: I
      - routingBel: OPTINV_ENBWREN_L
        bel_pin: I
    inverting:
     parameter:
       key: IS_ENBWREN_INVERTED
       textValue: 1
     routing_bels:
      - routingBel: OPTINV_ENBWREN_U
        bel_pin: I
      - routingBel: OPTINV_ENBWREN_L
        bel_pin: I
```

interchange/DeviceResources.capnp
```diff
@@ -203,5 +204,48 @@ struct Device {
...
+
+  struct CellPinInversionParameter {
+    # What parameter value configures this setting?
+    parameter    @0 : Dir.Netlist.PropertyMap.Entry;
+
+    # How are the routing BELs setup in this state?
+    routingBels  @1 : List(InversionRoutingBel);
+  }
+
+  struct CellPinInversion {
```
daveshah1, line 231: Looking at ensuring this is generic enough:

In Nexus (and IIRC ECP5) devices, a subset of a bus is sometimes invertible - ADA[3..0] and ADB[1..0] of the block RAMs (both are 14-bit buses).

It's also notable that in the vendor flow there is never an inversion parameter, just that an `INV` primitive created by synthesis is folded into the cell so I'm not sure if this how this would apply either.

Should this also consider pins that can be muxed with a constant inside a site - ECP5 and Nexus have quite a few of these too

litghost: So when the vendor tool "folds" the inversion, how does it mark it?  In my neive assumption, the cell would be modified to reflect the folding so that simulation tools/etc can reflect the new state.  Concretely:

```
wire sig1, sig1_inv;
INV inv(.I(sig1), .O(sig1_inv));

wire sig2, sig2_inv;
INV inv(.I(sig2), .O(sig2_inv));

EXAMPLE_CELL cell(.ADA({sig1_inv, sig2_inv}));
```

When folded, the INV cells will be gone, yes?  So how does the post folding netlist reflect the inversion?  I would expect something like:


```
EXAMPLE_CELL #(.IS_ADA_INVERTED(1)) cell(.ADA({sig1, sig2}));
```

etc.

litghost: > Should this also consider pins that can be muxed with a constant inside a site - ECP5 and Nexus have quite a few of these too

Not sure what you mean?  Do you mean something akin to a routing BEL with `SIG`, `INVSIG`, `CONST0`, `CONST1`?  Or concretely, something like the `SLICEL/PRECYINIT` with the ability to invert the CIN or CYINIT signals?

daveshah1: The inversion is a special 'property' stored on the packed/placed/routed cell - see screenshot:

Screenshot from 2020-11-18 21-07-38

There is never any parameter, in any netlist it is stored as an `INV` cell driving the pin.

Internally 'NeoCAD' text representation (I think ISE also used something like this) is also sometimes used, which looks like (example for a SLICE because it's what I have to hand but EBR is similar):

```
DP16K_MODE:::<regular parameters go here>:ADA0=#INV,CLKR=#SIG,ADB1=INV,...
```

noting the colon that delimits these from regular parameters.

In terms of constant-only options without an inversion choice, an example is DIRMUX in https://daveshah1.github.io/prjoxide-html/LIFCL/tilehtml/SYSIO_B5_1.html which can either be set to the DIR input or constant 0, inside the site. Similar for LSRINMUX.

litghost: The other option for absorbing `INV` is to explicitly add them to the site.  For example, this is how the `OBUFDS` and IOBUFDS` cells handle this.

litghost: > In terms of constant-only options without an inversion choice, an example is DIRMUX in https://daveshah1.github.io/prjoxide-html/LIFCL/tilehtml/SYSIO_B5_1.html which can either be set to the DIR input or constant 0, inside the site. Similar for LSRINMUX etc in this tile although this does have inversion too.

Ok, what you are describing is something completely different, but does need to be described.  The current routing BELs in the device resources can already express LSRINMUX/PRECYINIT/CEUSEDMUX/SRUSEDMUX style routing, but there is not currently a description of where VCC/GND sources can originate.  So what is missing around this is:
 - Where are constant sources in the fabric, both in the routing graph (think TIEOFF in 7-series) and site local (LSRINMUX, 7-series PRECYINIT/CEUSEDMUX/SRUSEDMUX)?
 - What is the "default" routable signal type?  In 7-series and US+, routing a VCC is much easier to route than a GND for most site pins.

daveshah1: > What is the "default" routable signal type? In 7-series and US+, routing a VCC is much easier to route than a GND for most site pins.

Same for most Lattice FPGAs (ECP5 and Nexus). For iCE40 LUT inputs, SR signals and equivalent float low whereas CE signals float high.

litghost: In cases where there is not an option to express the inversion as a parameter, we can simply let the parameter become "Void".  In terms of the routing BEL formulation, I believe the current formulation still handles what is needed.  In the `LSRINMUX` case, that could be expressed as a routing BEL with either 2 or 3 inputs.  One of the inputs would go to a GND source (e.g. `CYINITGND`).  The signal could either be `LSROUT` or `LSROUT`/`INV`.  In the first case (which is akin to US+), then "encoding" of the inversion is attached to the routing BEL in the physical net description: https://github.com/Xilinx/RapidWright/blob/1b315a103fc941f58319411c37f18a69121ad505/interchange/PhysicalNetlist.capnp#L112-L115

litghost: For clarity, I'm planning on another seperate PR to describe the constant source and default constant description.  I believe it is separate from the local site inverter case.

daveshah1: Great, that all makes sense to me

interchange/DeviceResources.capnp
```diff
@@ -203,5 +204,54 @@ struct Device {
...
-}
 
+  ######################################
+  # Inverting pins description
+  #
+  # This block describes local site wire
+  # inverters, site routing BELs, and
+  # parameters.
+  ######################################
+  struct InversionRoutingBel {
```
litghost, line 215: TODO(@litghost) - Add site type

litghost:
Make need to rework to seperate inverter BELs in the site, and invertable pins on a Cell.

litghost:
@clavin-xlnx This is ready for review.  I've fixed some bugs in the primitives and macros export, and integrated the new inversion methods to populate the schema.

PR #100, opened 2020-11-19
Add initial constants schema.

litghost:


daveshah1:
LGTM.

Although somewhat separate from this, something else that would be useful is information on the constant that a cell pin should be connected to, if it is left unconnected in the netlist.

litghost:
> LGTM.
> 
> Although somewhat separate from this, something else that would be useful is information on the constant that a cell pin should be connected to, if it is left unconnected in the netlist.

That is actually part of the Cell <-> BEL pin mapping.

litghost:
Remaining work on this PR:
 - [ ] Add wires using https://www.rapidwright.io/javadoc/com/xilinx/rapidwright/device/Node.html#isTied-- and friends
 - [ ] Add BELs matching on BEL type == "HARD0" and == "HARD1"
  - [ ] Use API once added

litghost:
This PR is blocked by a fix for #104

litghost:
The current tied wires encoding isn't good enough, or there is a database error.  Looking into it.

litghost:
Ok, this is ready for review.

clavin-xlnx (approved):
Looks good to me.

PR #112, opened 2020-12-17
Add metadata annotations to enable automated textual format conversion.

litghost:
This is an initial attempt at using capnp annotations to enable automated textual format conversions.  I've implemented a prototype that converts to/from capnp, YAML, and JSON, all as defined by the capnp schema.  Prototype is here: https://github.com/SymbiFlow/python-fpga-interchange/pull/7


litghost:
@clavin-xlnx I believe this is ready for review now.

PR #113, opened 2021-01-05
BlockCreator: fix loading wrong pblock

7FM:
Without this change all module implementations load the same pblock file and thus have all the same pblock (at least metadata wise) instead of their implemented one.


clavin-xlnx:
Thanks for the fix!

PR #116, opened 2021-01-08
Bugfixes for migrateCellAndSubCells

jakobwenzel:
This is a collection of issues I encountered while merging some EDIFNetlists.

Without this PR, `migrateCellAndSubCells` leaves the design in a somewhat inconsistent state:
* If a cell is newly inserted to the netlist, CellInsts will refer to it
* If a cell of the same name is already part of the design, CellInsts will refer to the old Cell that is not part of the Netlist

This PR tries to fix that by always having CellInsts refer to the cell that is part of the netlist.

Some discussion pointers:
* `migrateCellAndSubCells` gets called from closed source code. Therefore I cannot change its function signature. I had to introduce a helper function. This is not needed when you are able to recompile the closed source. Should I leave it as is, or replace with a version that will only work with recompiled closed source binaries?
* I had to change the function to be recursive. Is this a problem?

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFLibrary.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFLibrary.java
@@ -193,12 +193,11 @@ public class EDIFLibrary extends EDIFName {
 	 */
 	public void uniqueifyCellsWithPrefix(String name){
 		ArrayList<EDIFCell> renamedCells = new ArrayList<>(getCells());
-		String validEDIFPrefix = EDIFTools.makeNameEDIFCompatible(name);
 		cells.clear();
 		for(EDIFCell c : renamedCells){
 			c.setName(name + c.getName());
 			if(c.getEDIFName() != null){
-				c.setEDIFRename(validEDIFPrefix + c.getEDIFName()); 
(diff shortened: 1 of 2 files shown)


clavin-xlnx:
Thanks @jakobwenzel for the pull request and addressing these issues!  Unfortunately, the RapidWright project is in between contributor guidelines at the moment and I'm working with our legal team to formulate a process to accept contributions soon.  

I think this fix should be merged, I'll let you know when we have the contributor process updated. I'll probably just need one more step from you and we can accept the change.  Thanks for your help and patience!

clavin-xlnx:
@jakobwenzel - I've updated RapidWright's [CONTRIBUTING.md](https://github.com/Xilinx/RapidWright/blob/master/CONTRIBUTING.md) which you'll want to review to make sure things look good.  Principally, we'll just need you peform a sign-off of the DCO in your commit messages.  Please see the bottom of CONTRIBUTING.md for [instructions on how to update the commit with the sign-off message](https://github.com/Xilinx/RapidWright/blob/master/CONTRIBUTING.md).  Thanks!

jakobwenzel:
I updated all three PRs to include a signoff.

PR #118, opened 2021-01-12
detect NPE from resource not found

jakobwenzel:
When neither RAPIDWRIGHT_PATH is set, nor the resources are present on the classpath, a NullPointerException is thrown. This PR detects this case and prints an error message.

Normally, I would throw an exception here. However, exceptions seem to be frowned upon in RapidWright, so this PR just prints to `System.err`.

Kind Regards,
Jakob


clavin-xlnx:
Thanks @jakobwenzel, this looks good.  Just see my note from #116 about adding a sign-off for the DCO.

PR #120, opened 2021-01-12
PartNameTools: Normalize Part Names to lower case

jakobwenzel:
`PartNameTools.getPart` should find parts regardless of upper-/lowercase input.

This PR accomplishes this by normalizing all names to lowercase both before storing them and before looking them up.


clavin-xlnx:
This looks like a good change that will be more robust.  Again, same thing with the sign-off for the DCO requirement mentioned in #116.

PR #122, opened 2021-01-20
Corrected pblockCLEHeight for 7 series

ElaHobby:
Updated the computation of pblockCLEHeight. In original code, sliceMsRequired was divided by SLICES_PER_TILE. However, in case of 7 series, an M-type tile has only one M slice.


clavin-xlnx:
I think this change looks ok.  Although our contribution guidelines have changed, would you be willing to sign off on the change?  Please see our [contributing page ](https://github.com/Xilinx/RapidWright/blob/master/CONTRIBUTING.md)for details.

PR #132, opened 2021-02-19
Fix bug in alternative site pin directions for DeviceResources

litghost:


litghost:
@clavin-xlnx Please review when you get a chance!

clavin-xlnx (approved):
Thanks for catching this!

PR #133, opened 2021-03-11
Altered update instructions to match main README

reillymck:
Signed-off-by: Reilly McKendrick <romckend@byu.edu>


reillymck:
It appears that `make update_jars` does not do a git pull from what I understand. This tripped me up to be honest. Should the git pull be implied, should it be added to the makefile or should it be added as a step before calling `make update_jars`?

clavin-xlnx:
Thanks for the pull request.  Let's keep the `git pull` separate, that way users have more control over updating.  There might be situations where they might want to update the jars (and device data), but not get the latest code revisions (like incompatible feature, etc).

Issue #136, opened 2021-03-11
NullPointerException when running LogicalNetlistExample


reillymck:
This may be related to the other issue I opened concerning PhysicalNetlistExample. However, I have more information on this one.
The exception is cause by `n.getDevice()` returning null on [line 35](https://github.com/Xilinx/RapidWright/blob/master/src/com/xilinx/rapidwright/interchange/LogicalNetlistExample.java). I am going to open a pull request that I think partially solves the issue: I noticed that the getters for the device and design members of an EDIFNetlist object do not return correctly. However, this does not solve the problem. Upon examining the EDIF parser, I do not think that there is any place where the device for the edif object is actually set (although I believe that the parser accurately picks up the device name string from the first line of the EDIF file). I noticed that in parseEDIFNetlist(), on [line 398](https://github.com/Xilinx/RapidWright/blob/master/src/com/xilinx/rapidwright/edif/EDIFParser.java) the design is set. I think that this function would be the appropriate place to call setDevice() as well. I am not very sure specifically how/where to add that, but I think it will fix the problem.

clavin-xlnx:
The problem is that RapidWright cannot auto detect the device being targeted by the EDIF file and looks for the device partname attribute.  It is likely that the EDIF file does not contain this attribute and thus doesn't know what device to load.  For example, in pblock0.dcp (from #134), the EDIF has the following at the end of the file:

```
  (design picoblaze_top
    (cellref picoblaze_top (libraryref work))
    (property XLNX_PROJ_DIR (string "/home/clavin/testRW/picoblaze"))
    (property PART (string "xcvu3p-ffvc1517-2-i"))
  )
)
```

Check your EDIF file to see if it has the PART property and if it is set.  If not, you could try adding it.  Unfortunately, the PART property is not a mandatory field and is not always written by Vivado.

reillymck:
Thanks for pointing this out. The EDIF that I want to use with this example does not have that PART property set. However, before I raised this issue, I did try running this with the pblock.edf file that your provide for the PhysicalNetlistExample, and it still gets the NullPointerException.

reillymck:
At the very least, maybe there should be a conditional that sees if the EDIF has the part declared, and if it does, updates the device property of the netlist? It looks like the example expects this property to be set on the EDIFNetlist object, but that is not happening, even with an EDIF file that has the PART declared.

litghost:
[...]

PR #137, opened 2021-03-11
Fixed small getter bug for device/design in EDIFNetlist

reillymck:
See issue #136 

Signed-off-by: ReillyMcK <romckend@byu.edu>


clavin-xlnx:
Are you seeing different behavior based on this change?  My understanding is that the code change should not have any effect.

reillymck:
I am not seeing different behavior, like I pointed out in the issue, I thought the main problem was that the EDIF parser never sets the device member for the EDIFNetlist object. Although, I felt like it was safe to assume that these getters are supposed to return _this_.device and _this_.design instead of device/design. I scanned the document and from what I see global device/design variables are never set, but there are setters for these members for the EDIFNetlist object, so I assumed these are supposed to get what was previously set.

Issue #143, opened 2021-03-24
Missing VCC/GND ties on some Cell BEL pin maps


litghost:
I'm currently debugging the Cell BEL pin maps for the RAMB18E1 on 7-series.  The current mapping appears to be missing some VCC/GND ties on pins in some configurations.  

For example, the pin map for `WRITE_WIDTH_B` = `9` is currently:
```
      - parametersSiteTypes:
          - bel: RAMB18E1
            siteType: RAMB18E1
            parameter:
              key: WRITE_WIDTH_B
              textValue: 1
          - bel: RAMB18E1
            siteType: RAMB18E1
            parameter:
              key: WRITE_WIDTH_B
              textValue: 2
          - bel: RAMB18E1
            siteType: RAMB18E1
            parameter:
              key: WRITE_WIDTH_B
              textValue: 4
          - bel: RAMB18E1
            siteType: RAMB18E1
            parameter:
              key: WRITE_WIDTH_B
              textValue: 9
        pins:
          - cellPin: 'WEBWE[0]'
            belPin: WEBWE0
          - cellPin: 'WEBWE[0]'
            belPin: WEBWE1
          - cellPin: 'WEBWE[0]'
            belPin: WEBWE2
          - cellPin: 'WEBWE[0]'
            belPin: WEBWE3
```

The `WEBWE[0]` 1 to many relationship is correct.  However, I believe that there should be additional lines to tied `WEBWE4` to `WEBWE7` to `GND`.  The current `DeviceResourceWriter` does not do this.  So I believe the correct `pins` block would be:
```
        pins:
          - cellPin: 'WEBWE[0]'
            belPin: WEBWE0
          - cellPin: 'WEBWE[0]'
            belPin: WEBWE1
          - cellPin: 'WEBWE[0]'
            belPin: WEBWE2
          - cellPin: 'WEBWE[0]'
            belPin: WEBWE3
          - cellPin: GND
            belPin: WEBWE4
          - cellPin: GND
            belPin: WEBWE5
          - cellPin: GND
            belPin: WEBWE6
          - cellPin: GND
            belPin: WEBWE7
```

There also appears to be an issue where `REGCLKARDRCLK` and `REGCLKB` should be tied to `VCC`, but I haven't isolated where in the map that should be happening.

I'm going to continue to investigate to determine if I can find out where the issue is.

litghost:
[...]

PR #144, opened 2021-03-24
Add PinMapTester tool to check P2L mapping from RapidWright.

litghost:
This is a little tool to explore where the problems seen in #143 are coming from.


clavin-xlnx:
Thanks @litghost , I am also investigating in parallel.

PR #145, opened 2021-03-25
WIP: Default input pins to GND.

litghost:
This is based on conversion from #143 .


litghost:
Per discussion in #143, this approach will not work. Closing.

PR #156, opened 2021-04-13
allow ports to have multiple pins

jakobwenzel:


jakobwenzel:
I might have found a bug in this PR, please don't merge yet. I'll investigate.

PR #160, opened 2021-04-23
interchange: Use fixed names for macro and primitive libraries

gatecat:
This is the complement to SymbiFlow/fpga-interchange-schema#43.

The renaming is done while reading/writing the logical netlist structure itself, as the `hdi_primitives` name is expected to be used for the primitives library in a lot of the EDIF related code. Hopefully this is reasonable.

### DIFF
--- a/src/com/xilinx/rapidwright/interchange/DeviceResourcesWriter.java
+++ b/src/com/xilinx/rapidwright/interchange/DeviceResourcesWriter.java
@@ -56,6 +56,7 @@ import com.xilinx.rapidwright.edif.EDIFLibrary;
 import com.xilinx.rapidwright.edif.EDIFNetlist;
 import com.xilinx.rapidwright.edif.EDIFPort;
 import com.xilinx.rapidwright.edif.EDIFPortInst;
+import com.xilinx.rapidwright.edif.EDIFTools;
 import com.xilinx.rapidwright.interchange.EnumerateCellBelMapping;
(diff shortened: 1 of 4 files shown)


src/com/xilinx/rapidwright/interchange/DeviceResourcesWriter.java
```diff
@@ -348,7 +349,11 @@ public static void writeDeviceResourcesFile(String part, Device device, CodePerf
 
         Netlist.Builder netlistBuilder = devBuilder.getPrimLibs();
         netlistBuilder.setName(netlist.getName());
-        LogNetlistWriter writer = new LogNetlistWriter(allStrings);
+        LogNetlistWriter writer = new LogNetlistWriter(allStrings, new HashMap<String, String>() {{
+                    put(EDIFTools.EDIF_LIBRARY_HDI_PRIMITIVES_NAME, "primitives");
+                    put(device.getSeries()+"_"+EDIFTools.MACRO_PRIMITIVES_LIB, "macros");
```
clavin-xlnx, line 354: @gatecat Perhaps the `"macros"` and `"primitives"` strings should be declared as a `public static final String` at the top of one of the LogNetlist classes and named such that they indicate a standardized Interchange library name.

gatecat: Good idea, updated

PR #165, opened 2021-04-27
Gradle: Allow Building as Application

jakobwenzel:
This PR allows building RapidWright as a Gradle Application. These new tasks are available:

* `run`: Directly run RapidWright
* `installDist`: Create a Distribution containing all Java Code as well as Start Scripts. This task is confusingly named for people coming from make. It does not install anything. It only creates a distribution inside the build directory.
* `distZip`/`distTar`: Create a Zip/Tar from the Distribution.

Since we have many Main-Files in RapidWright, this PR also creates a new Single Entrypoint that delegates to the other main classes by looking at the first argument.

I just set the arguments to be the same as the class names. This is good for most classes, but may require some more thought for some classes like [RapidWright](https://github.com/Xilinx/RapidWright/blob/b6ba4cf6d702ae6e826bea73723c4d0abcc25126/src/com/xilinx/rapidwright/util/RapidWright.java).

### DIFF
--- a/src/com/xilinx/rapidwright/MainEntrypoint.java
+++ b/src/com/xilinx/rapidwright/MainEntrypoint.java
@@ -1,0 +1,181 @@
+package com.xilinx.rapidwright;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import com.xilinx.rapidwright.debug.DesignInstrumentor;
+import com.xilinx.rapidwright.debug.ILAInserter;
+import com.xilinx.rapidwright.debug.ProbeRouter;
+import com.xilinx.rapidwright.design.MetadataParser;
+import com.xilinx.rapidwright.design.blocks.ImplGuide;
+import com.xilinx.rapidwright.design.blocks.PBlock;
+import com.xilinx.rapidwright.design.blocks.PBlockGenerator;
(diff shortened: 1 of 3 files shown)


src/com/xilinx/rapidwright/MainEntrypoint.java
```diff
@@ -0,0 +1,182 @@
...
+        addFunction("EDIFParser", EDIFParser::main);
+        addFunction("EDIFPropertyValue", EDIFPropertyValue::main);
+        addFunction("EDIFTools", EDIFTools::main);
+        addFunction("EnumerateCellBelMapping", EnumerateCellBelMapping::main);
+        addFunction("ExampleNetlistCreation", ExampleNetlistCreation::main);
+        addFunction("FileTools", FileTools::main);
+        addFunction("HandPlacer", HandPlacer::main);
+        addFunction("ILAInserter", ILAInserter::main);
+        addFunction("ImplGuide", ImplGuide::main);
+        addFunction("Installer", Installer::main);
```
clavin-xlnx, line 113: @jakobwenzel Let's remove the Installer, I can't think of any useful cases where we would want to run this from an existing install.

jakobwenzel: Done

PR #172, opened 2021-05-04
create single method for selecting Job Types

jakobwenzel:
Instead of deciding between LocalJob and LSFJob in all places where a job is required, this PR provides a factory method that does encapsulates the checking.

Since `JobQueue.isLSFAvailable` will be run more often, this PR introduces some caching there.


clavin-xlnx:
These are probably good changes to make.  Ideally, others with different cluster management systems (Sun Grid Engine, etc) could extend the `Job` class and potentially modify the `createJob()` method accordingly.

Issue #162, opened 2021-04-26
Add macro parameter mapping rules


gatecat:
See SymbiFlow/fpga-interchange-schema#42 for more background

At the moment, the way that parameters are split up for primitives like distributed RAM and `LUT6_2` is not output into the resources. 

This will either need to be hardcoded into RapidWright, or included in the Vivado-derived database. I don't know how feasible the latter option would be.

clavin-xlnx:
This appears to have been resolved per #213.

gatecat:
Thanks, yes, this should be fine now!

PR #180, opened 2021-05-12
[RFC] Adding macro parameter mapping rules

gatecat:
This is an attempt to work on #162. Don't merge it yet as it still depends on SymbiFlow/fpga-interchange-schema#42 being merged.

It adds hardcoded macro parameter mapping rules for a few 7-series macros; and exports them to the device resources. Some feedback on this approach would be appreciated before I work on this more.

### DIFF
--- a/src/com/xilinx/rapidwright/interchange/MacroParamRule.java
+++ b/src/com/xilinx/rapidwright/interchange/MacroParamRule.java
@@ -1,0 +1,35 @@
+package com.xilinx.rapidwright.interchange;
+
+public class MacroParamRule {
+    public String primParam;
+    public String instName;
+    public String instParam;
+    public int[] bitSlice;
+    public ParameterMapEntry[] tableLookup;
+
+    public MacroParamRule(String primParam, String instName, String instParam) {
+        this.primParam = primParam;
+        this.instName = instName;
+        this.instParam = instParam;
+        this.bitSlice = null;
+        this.tableLookup = null;
+    }
+
+    public static MacroParamRule copy(String primParam, String instName, String instParam) {
+        return new MacroParamRule(primParam, instName, instParam);
+    }
+
+    public static MacroParamRule bit_range(String primParam, String instName, String instParam, int start, int end) {
+        MacroParamRule result = new MacroParamRule(primParam, instName, instParam);
+        result.bitSlice = new int[(end-start) + 1];
+        for (int i = 0; i <= (end-start); i++)
+            result.bitSlice[i] = start + i;
+        return result;
+    }
+
+    public static MacroParamRule table(String primParam, String instName, String instParam, ParameterMapEntry[] table) {
+        MacroParamRule result = new MacroParamRule(primParam, instName, instParam);
+        result.tableLookup = table;
+        return result;
+    }
+}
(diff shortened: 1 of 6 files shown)


src/com/xilinx/rapidwright/interchange/MacroParamRule.java
```diff
@@ -0,0 +1,68 @@
...
+        this.instParam = instParam;
+        this.bitSlice = null;
+        this.tableLookup = null;
+    }
+
+    public static MacroParamRule copy(String primParam, String instName, String instParam) {
+        return new MacroParamRule(primParam, instName, instParam);
+    }
+
+    public static MacroParamRule bit_range(String primParam, String instName, String instParam, int start, int count) {
```
clavin-xlnx, line 24: Would it be ok to switch to camelCase? (`bit_range() --> bitRange()`)

src/com/xilinx/rapidwright/interchange/MacroParamRule.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public static MacroParamRule table(String primParam, String instName, String instParam, MacroParamTableEntry[] table) {
+        MacroParamRule result = new MacroParamRule(primParam, instName, instParam);
+        result.tableLookup = table;
+        return result;
+    }
+
+    public static MacroParamRule[] getRules(Unisim prim) {
+        switch (prim) {
```
clavin-xlnx, line 39: Thanks @gatecat for the work.  I am wondering if we could avoid hardcoding the data.  I am looking at extracting it from the Macro EDIFLibrary netlist which seems to have mostly all the information (easy way to dump to a file):
```
Design.getMacroPrimitives(Series.Series7).exportEDIF(new FileWriter("Series7_macros.edf"));
```
Based on the examples, the only thing I can't seem to derive is the `start` of `bitRange`.  Perhaps you could describe how you inferred those values for your examples and we could device a scheme to derive or augment existing RapidWright data to supply them.

src/com/xilinx/rapidwright/interchange/MacroParamRule.java
gatecat: sure, happy to fix if we proceed with this

src/com/xilinx/rapidwright/interchange/MacroParamRule.java
gatecat: As an example of the way to infer these, consider the following Tcl:

```tcl
link_design -part xc7a35tftg256-1
create_cell -reference RAM128X1D ram_i
set_property INIT 128'hFFEEDDCCBBAA99887766554433221100 [get_cells ram_i]
puts "LOW: [get_property INIT [get_cells ram_i/DP.LOW]]"
puts "HIGH: [get_property INIT [get_cells ram_i/DP.HIGH]]"
```

prints
```
LOW: 64'h7766554433221100
HIGH: 64'hFFEEDDCCBBAA9988
```

it would definitely be even better to have this info available directly in RapidWright! I'd also be curious if the placement rules for macros are something that could be included as part of the RapidWright in the longer term (I think Vivado calls these "shapes" based on the error messages when you get it wrong.)

PR #185, opened 2021-05-19
Clear Cached Image in TileScene

jakobwenzel:
This fixes the issue raised at https://groups.google.com/g/rapidwright/c/YyEYI0Mebu8

Also initialize the image to black before drawing on it. For me, there were some artifacts in the image otherwise.


clavin-xlnx:
Thanks Jakob!

PR #198, opened 2021-07-02
correctly close leaked file descriptors

jakobwenzel:
This PR addresses multiple instances of File Descriptors not being closed. It includes simple testcases for writing out and reading back in both EDIF and DCP files without leaking file descriptors. As we don't have a way to test designs for equality, I just check for very obvious problems.

This PR is dependent on #197. Github's view for this PR will include all of #197's diff until that one is merged.


jakobwenzel:
The Github Action was running out of Memory. Switched to a very small device to compensate.

More details about the current failure should be visible once #200 is merged.

PR #204, opened 2021-07-14
interchange: Fix wire type categories for UltraScale+ clocks

gatecat:
These are all intents that are needed to connect between pins and global routing, so they must be 'special' rather than 'general'.


clavin-xlnx:
LGTM

PR #213, opened 2021-07-31
Builds and adds onto #180.  Discovers data from Vivado and generates …

clavin-xlnx:
…static code to make available in RapidWright.


clavin-xlnx:
@gatecat - this is an attempt to build on #180.  I harvest the information from Vivado and then automatically create a static class with the rule mappings.  It unfortunately doesn't handle some complex mappings (RAM32M, for example), but provides support for most other macros.

gatecat (approved):
LGTM, thanks!

PR #224, opened 2021-08-17
Add method to uncrypt EDIF inside Checkpoints via Vivado

jakobwenzel:


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2225,4 +2231,50 @@ public static void makePhysNetNamesConsistent(Design design) {
...
+		}
+	}
+
+	/**
+	 * Use Vivado to create a readable version of the EDIF file inside a Checkpoint. Do nothing if the file already exists.
+	 * @param dcp the checkpoint
+	 * @param edfFileName filename to use or null if we should select a filename
+	 * @return the output edif filename
+	 */
+	public static Path genReadableEdif(Path dcp, Path edfFileName) {
```
clavin-xlnx, line 2269: Perhaps to be consistent with the other method being added, it should be called `genReadableEDIF()`?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2225,4 +2231,50 @@ public static void makePhysNetNamesConsistent(Design design) {
...
+
+
+			return job;
+		} catch (IOException e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	/**
+	 * Use Vivado to create a readable version of the EDIF file inside a Checkpoint. Do nothing if the file already exists.
```
clavin-xlnx, line 2264: The Javadoc specifies that it will do nothing if the EDIF file already exists, however I did not see any checking logic to support that statement.  In a quick test, it launched Vivado despite an EDIF file already being available.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2225,4 +2231,50 @@ public static void makePhysNetNamesConsistent(Design design) {
...
+	 * @param edif the output EDIF
+	 * @return the created Job
+	 */
+	public static Job generateReadableEDIFJob(Path checkpoint, Path edif) {
+		try {
+
+			final Job job = new LocalJob();
+			job.setCommand(FileTools.getVivadoPath() + " -mode batch -source readable.tcl");
+
+			final Path runDir = Files.createTempDirectory(edif.toAbsolutePath().getParent(),edif.getFileName()+"_readable_edif_");
```
clavin-xlnx, line 2248: It appears that the temporary directory of files does not get deleted after things finish.  I can see a case to keeping it if EDIF conversion fails, however, it should probably be deleted if the job completes successfully.

PR #229, opened 2021-09-17
RWRoute source code

YunxZhou:


clavin-xlnx:
Thanks @YunxZhou for the pull request.  One high level comment, I see a DCP and a number of text files in the request:

•	A tcl/rwroute/test_data/dump_all_dsp_delay/bd_0_i-hls_inst-inst-call_ret_projection_fu_3263-mul_ln1371_1_fu_108_p2.txt (57) 
•	A tcl/rwroute/test_data/dump_all_dsp_delay/bd_0_i-hls_inst-inst-call_ret_projection_fu_3263-mul_ln1371_2_fu_128_p2.txt (57) 
•	A tcl/rwroute/test_data/dump_all_dsp_delay/bd_0_i-hls_inst-inst-call_ret_projection_fu_3263-mul_ln1371_fu_88_p2.txt (57) 
•	A tcl/rwroute/test_data/dump_all_dsp_delay/bd_0_i-hls_inst-inst-grp_rasterization1_fu_3239-mul_ln214_fu_468_p2.txt (329) 
•	A tcl/rwroute/test_data/dump_all_dsp_delay/bd_0_i-hls_inst-inst-grp_rasterization1_fu_3239-ret_V_28_fu_206_p2.txt (2029) 
•	A tcl/rwroute/test_data/dump_all_dsp_delay/bd_0_i-hls_inst-inst-grp_rasterization1_fu_3239-ret_V_31_fu_244_p2.txt (364) 
•	A tcl/rwroute/test_data/dump_all_dsp_delay/bd_0_i-hls_inst-inst-grp_rasterization1_fu_3239-ret_V_32_fu_254_p2.txt (144) 
•	A tcl/rwroute/test_data/dump_all_dsp_delay/bd_0_i-hls_inst-inst-grp_rasterization2_fu_3192-grp_fu_468_p2.txt (147) 
•	A tcl/rwroute/test_data/dump_all_dsp_delay/bd_0_i-hls_inst-inst-grp_rasterization2_fu_3192-grp_fu_475_p2.txt (147) 
•	A tcl/rwroute/test_data/dump_all_dsp_delay/bd_0_i-hls_inst-inst-grp_rasterization2_fu_3192-grp_fu_482_p2.txt (147) 
•	A tcl/rwroute/test_data/dump_all_dsp_delay/bd_0_i-hls_inst-inst-grp_rasterization2_fu_3192-ret_V_11_reg_628_reg.txt (148) 
•	A tcl/rwroute/test_data/dump_all_dsp_delay/bd_0_i-hls_inst-inst-grp_rasterization2_fu_3192-ret_V_18_reg_633_reg.txt (148) 
•	A tcl/rwroute/test_data/dump_all_dsp_delay/bd_0_i-hls_inst-inst-grp_rasterization2_fu_3192-ret_V_25_reg_638_reg.txt (148) 
•	A tcl/rwroute/test_data/dump_all_dsp_delay/routed_3d-rendering_4.0_post-vr.dcp (0) 
•	A tcl/rwroute/test_data/find_clock_route_template/X36Y157-X36Y157_X0Y58_X44Y150-X52Y270_2020.1.txt (31) 
•	A tcl/rwroute/test_data/find_clock_route_template/X36Y157-X36Y157_X0Y58_X44Y150-X52Y270_2020.2.txt (31) 
•	A tcl/rwroute/test_data/find_clock_tree_template/X2Y2-X3Y3_BUFG_X4Y0_2020.1.txt (45) 
•	A tcl/rwroute/test_data/find_clock_tree_template/X2Y2-X3Y3_BUFG_X4Y0_2020.2.txt (45) 

Were these intended to be included?

YunxZhou:
@pongstorn-xilinx Please see Chris's comments above regarding the files.

eddieh-xlnx (commented):
Great job building such complex machinery! Here are some initial comments.

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -0,0 +1,855 @@
...
+	}
+	
+	/**
+	 * A lightweight useful class for different simple routing-related scenarios, 
+	 * each {@link RoutingNode} Object is associated to a {@link Node} Object.
+	 */
+	static class RoutingNode{
+		private Node node;
+		private RoutingNode prev;
+		private boolean isTareget;
```
eddieh-xlnx, line 807:
```suggestion
		private boolean isTarget;
```
(and also its uses and methods below)

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -0,0 +1,855 @@
...
+		if(usedRoutingNodes != null && usedRoutingNodes.contains(routingNode))
+			return true;
+		Node node = routingNode.getNode();
+		return isNodeUsableStaticSource(node, type, design);
+	}
+	
+	/**
+	 * This method handles queries during the static source routing process. 
+	 * It determines if the node in question can be used as a source for the current NetType.
+	 * @param routingNode The ResourceNode in question.
```
eddieh-xlnx, line 774:
```suggestion
	 * @param node The Node in question.
```

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -0,0 +1,855 @@
...
+				lutOutputPinNames.add("CLE_CLE_"+cle+"_SITE_0_"+pin+"_O");
+			}
+		}
+	}
+	
+	/**
+	 * Routes a static net (GND or VCC).
+	 * @param currNet The current static net to be routed.
+	 * @param unavailableNodes A set of unavailable nodes.
+	 * @param routethruHelper Instantiated RoutethruHelper to exclude routethru nodes.
```
eddieh-xlnx, line 642:
```suggestion
```

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -0,0 +1,1843 @@
...
+			this.assignNodesToConnections();
+			// fix routes with cycles and / or multi-driver nodes
+			Set<NetWrapper> routes = this.fixRoutes();
+			if(this.config.isTimingDriven()) this.updateTimingAfterFixingRoutes(routes);
+		}
+	}
+	
+	/**
+	 * Checks if a connection should be routed.
+	 * @param connection The connection in question.
```
eddieh-xlnx, line 842:
```suggestion
	 * @param con The connection in question.
```

src/com/xilinx/rapidwright/rwroute/NetWrapper.java
```diff
@@ -0,0 +1,165 @@
...
+		return id;
+	}
+	
+	public void addCons(Connection c){
+		this.connections.add(c);	
+	}
+	
+	public List<Connection> getConnection(){
+		return this.connections;
+	}
```
eddieh-xlnx, line 118:
```suggestion
	public void addConnection(Connection c){
		this.connections.add(c);	
	}
	
	public List<Connection> getConnections(){
		return this.connections;
	}
```

src/com/xilinx/rapidwright/rwroute/RoutableNode.java
```diff
@@ -0,0 +1,570 @@
...
+public class RoutableNode implements Routable{
+	/** A unique index of a rnode */
+	private int index;
+	/** The associated {@link Node} instance */
+	private Node node;
+	/** The type of a rnode*/
+	private RoutableType type;
+	/** The tileXCoordinate and tileYCoordinate of the INT tile that the associated node stops at */
+	private short endTileXCoordinate;
+	private short endTileYCoordinate;
```
eddieh-xlnx, line 55: Is considering just the coordinates of the end tile sufficient in all cases? It seems that this is used for a lot of things, including for finding the geometric center, bounding box, etc.

One extension I've been looking to add is to constrain the routing within a `PBlock`, and this requires knowing the {min,max}{X,Y} of the node.

src/com/xilinx/rapidwright/rwroute/RoutableNode.java
```diff
@@ -0,0 +1,570 @@
...
+		Wire[] wires = this.node.getAllWiresInNode();
+		List<Tile> intTiles = new ArrayList<>();
+		for(Wire w : wires) {
+			if(w.getTile().getTileTypeEnum() == TileTypeEnum.INT) {
+				intTiles.add(w.getTile());
+			}
+		}
+		Tile endTile = null;
+		if(intTiles.size() > 1) {
+			endTile = intTiles.get(1);
```
eddieh-xlnx, line 219: Is it guaranteed that the end tile is the second one in this list?

src/com/xilinx/rapidwright/rwroute/Routable.java
```diff
@@ -0,0 +1,313 @@
...
+	 */
+	public float getHistoricalCongesCost();
+	/**
+	 * Gets the historical congestion cost of a Routable Object.
+	 * @param historicalCongesCost The historical congestion cost to be set.
+	 */
+	public void setHistoricalCongesCost(float historicalCongesCost);
+	/**
+	 * Checks if coordinates of a Routable Object is within the conncetion's bounding box.
+	 * @param connetion The connection that is being routed.
```
eddieh-xlnx, line 139:
```suggestion
	 * @param connection The connection that is being routed.
```

eddieh-xlnx, line 139:
```suggestion
	 * Checks if coordinates of a Routable Object is within the connection's bounding box.
	 * @param connection The connection that is being routed.
```

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -0,0 +1,1843 @@
...
+		this.saveRouting(connection);	
+		connection.getSinkRnode().setTarget(false);		
+		this.resetExpansion();		
+		this.updateUsersAndPresentCongesCost(connection);
+	}
+	
+	/**
+	 * Traces back for a connection from its sink rnode to its source, in order to build the routing path.
+	 * Storing the path of the connection.
+	 * @param con: The connection that is being routed.
```
eddieh-xlnx, line 1444:
```suggestion
	 * @param connection: The connection that is being routed.
```

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -0,0 +1,1843 @@
...
+			
+		return true;
+	}
+	
+	/**
+	 * Unroutes preserved nets to release routing resource to resolve congestion that blocks the routablity of a connection.
+	 * @param connection The connection in question.
+	 * @return The number of unrouted nets.
+	 */
+	private int unrouteReservedNetsToReleaseResources(Connection connection) {
```
eddieh-xlnx, line 1344: Is my understanding correct that this just unroutes reserved nets that are occupying resources immediately downhill of the source, and immediately uphill of the sink? If the congestion is outside of these immediate vicinities, then may not help?

Also, it looks like you're unrouting the whole reserved net. Have you considered unrouting just the associated routing branch of the reserved net (assuming it has more than one sink) in order to preserve as much as possible?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -0,0 +1,1843 @@
...
+			for(SitePinInst sink : this.staticNetAndRoutingTargets.get(n)) {
+				this.preservedNodes.remove(sink.getConnectedNode());
+			}
+		}
+		
+		RouterHelper.invertPossibleGndPinsToVccPins(this.design, this.design.getGndNet());
+		
+		unavailableNodes.addAll(this.preservedNodes.keySet());
+		// if connections are routed first, used resources should be preserved
+		unavailableNodes.addAll(getAllUsedNodesOfRoutedConnections());
```
eddieh-xlnx, line 462: Is this redundant considering you initialised `unavailableNodes` a few lines above with the result of this function already?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -0,0 +1,1843 @@
...
+			rnode.updatePresentCongesCost(this.presentCongesFac);
+		}
+	}
+	
+	/**
+	 * Updates the users and present congestion cost of rnodes used by a routed connection.
+	 * @param connection The routed connection.
+	 */
+	private void updateUsersAndPresentCongesCost(Connection connection){
+		for(int i = connection.getRnodes().size()-1; i >= 0; i--){
```
eddieh-xlnx, line 1189: Is it important to iterate backwards here?

clavin-xlnx:
@YunxZhou - One thing to standardize on all variable, class and method names is regarding the term "connection".  I would recommend that you use either the full word or an abbreviated "conn" in places where it appears to be too long with a focus on consistency.  There are some places where it is "con" which has a negative connotation and, in my opinion, "conn" is a more popular way to abbreviate the term.  Although, some also use "cxn", that might be less obvious to some.  I would recommend "connection" first, and then "conn" where it is part of a phrase that requires some shortening.

src/com/xilinx/rapidwright/rwroute/NetWrapper.java
```diff
@@ -0,0 +1,165 @@
...
+	private float xCenter;
+	private float yCenter;
+	/** The half-perimeter wirelength */
+	private short doubleHpwl;
+	/** A flag to indicate if the source has been swapped */
+	private boolean sourceChanged;
+	/** Stores the old source SitePinInst after output pin swapping */
+	private SitePinInst oldSource;
+	
+	public NetWrapper(int id, short bbRange, Net net){
```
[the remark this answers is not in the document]
YunxZhou: It is no longer used. Removed.

src/com/xilinx/rapidwright/rwroute/RoutableNode.java
YunxZhou: Knowing the end tile of nodes should be enough. For example, for a horizontal node going from left to right, its min X (the starting point) is the end tile X coordinate of its uphill nodes.

src/com/xilinx/rapidwright/rwroute/RoutableNode.java
YunxZhou: Right, the second one in the list is the end tile. One thing should be noted is that the end X and Y coordinates could be inaccurate for some nodes that have two exit points.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
YunxZhou: Redundancy removed.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
YunxZhou: Not really. Updated in the new commit.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
YunxZhou: That is a correct understanding. If the congestion or blockage if outside of these immediate vicinities, enlarging bounding boxes is there to help by allowing enough routing detours.

Unrouting preserved nets is intended to be used for the soft preserve mode. The partial unrouting is not explored yet. (PS: there could be multiple associated branches of a preserved net to a node).

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -0,0 +1,1841 @@
...
+	}
+	
+	/**
+	 * Rips up a connection.
+	 * @param connection The connection to be ripped up.
+	 */
+	private void ripUp(Connection connection){
+		for(Routable rnode : connection.getRnodes()) {
+			rnode.reduceConnectionCountOfUser(connection.getSource());
+			rnode.reduceConnectionCountOfUser(connection.getNetWrapper().getOldSource());
```
eddieh-xlnx, line 1179: Can you explain why this is necessary?
It looks like there could be a mismatch between the number of times `Routable.addUser()` is called and the number of times `Routable.reduceConnectionCountOfUser()` is called.

src/com/xilinx/rapidwright/rwroute/RoutableNode.java
```diff
@@ -0,0 +1,570 @@
...
+	@Override
+	public int numUniqueUsers() {
+		if(this.usersConnectionCounts == null) {
+			return 0;
+		}
+		return this.usersConnectionCounts.size();
+	}
+	
+	@Override
+	public void reduceConnectionCountOfUser(SitePinInst source) {
```
eddieh-xlnx, line 405: Since it looks like this method is the opposite of `addUser`, could their names be made consistent?

src/com/xilinx/rapidwright/rwroute/RoutableNode.java
eddieh-xlnx: I've found one case where the logic in this function seems to fail. My sink node is `INT_X6Y84/BOUNCE_E_15_FT0` which has the following wires: `INT_X6Y84/BOUNCE_E_15_FT0, INT_X6Y85/BOUNCE_E_BLN_15_FT1, CLEL_R_X6Y84/CLE_CLE_L_SITE_0_H_I]`
The logic uses the second of the two `INT` as the end tile, when actually the sink is at `CLEL_R_X6Y84`.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
eddieh-xlnx: Enlarging the bounding box may not be sufficient in all cases. Let's say net 1 is trying to route. If there is at least one preserved net using all of the resources that net 1 needs to successfully route, and those preserved nets do not appear in the net 1's source and sink vicinities, then they will never be unrouted (in soft preserve mode).

My intuition right now is that the root of this stems from the fact that one can *never* use reserved nodes as they simply don't appear in the routing graph. It may be beneficial to allow nets that really need to use a reserved resource to use it (heavily discouraged by setting a very high cost) so that we can examine exactly which reserved nets need to be unrouted.

It also appears that all reserved nets in the vicinity of the source and sink are fully unrouted, even if they are not needed. The approach above might be able to help this. I'm experimenting with this router for constrained scenarios (restricting routing inside a PBlock like `CONTAIN_ROUTING` does) as well as trying to encourage the router to rip up as little as possible.

Regarding unrouting a branch of the net, yes, a branch may lead to multiple sinks depending on if it's a thick or thin branch, but I am suggesting that a subset of all sinks be unrouted where possible (i.e. when the conflicting node does not lie on the trunk) instead of the entire net -- which matches my objective of ripping up as little as possible.

I'm not suggesting you fix this now as it could be a big change on top of what is already a big PR, but perhaps update the comment to indicate this limitation?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -0,0 +1,1841 @@
...
+				for(Node uphill : toBuild.getAllUphillNodes()) {
+					// Without this routethru check, there will be Invalid Programming for Site error shown in Vivado.
+					// Do not use those nodes, because we do not know if the routethru is available or not
+					if(routethruHelper.isRouteThru(uphill, toBuild)) continue;
+					RoutableNode parent = (RoutableNode) this.rnodesCreated.get(uphill);
+					if(parent != null && !parent.childrenNotSet()) {
+						if(!parent.getChildren().contains(rnode)) parent.getChildren().add(rnode);
+					}
+				}
+			}
```
eddieh-xlnx, line 1392: Can this loop be merged with the one above?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -0,0 +1,1841 @@
...
+			
+			NetWrapper netnew = this.createsNetWrapperAndConnections(n, this.config.getBoundingBoxExtensionX(), this.config.getBoundingBoxExtensionY(), this.multiSLRDevice);
+			
+			for(int i = 0; i < reservedNetNodes.size(); i ++) {
+				Node toBuild = reservedNetNodes.get(i);
+				Routable rnode = this.rnodesCreated.get(toBuild);
+				if(rnode == null) {
+					rnode = new RoutableNode(this.rnodeId, toBuild, RoutableType.WIRE);
+					this.rnodeId++;
+					this.rnodesCreated.put(toBuild, rnode);
```
eddieh-xlnx, line 1379: Is it possible to reuse `createAddRoutableNode` here?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -0,0 +1,1841 @@
...
+	/**
+	 * Creates a {@link RoutableNode} Object based on a {@link Node} instance and avoids duplicates,
+	 * used for creating the source and sink rnodes of {@link Connection} instances.
+	 * @param rnodeGlobalIndex The design-wise index of created {@link RoutableNode} instances.
+	 * @param sitePinInst The source or sink {@link SitePinInst} instance.
+	 * @param node The node associated to the {@link SitePinInst} instance.
+	 * @param type The {@link RoutableType} of the {@link RoutableNode} Object.
+	 * @return The created {@link RoutableNode} instance.
+	 */
+	private Routable createAddRoutableNode(int rnodeGlobalIndex, SitePinInst sitePinInst, Node node, RoutableType type){
```
eddieh-xlnx, line 632: Does `rnodeGlobalIndex` need to be passed in since it's always `this.rnodeId`?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -0,0 +1,1841 @@
...
+	
+	protected void addStaticNetRoutingTargets(Net staticNet, List<SitePinInst> sinks) {
+		this.staticNetAndRoutingTargets.put(staticNet, sinks);
+	}
+	
+	/**
+	 * Routes static nets with preserved resources list supplied to avoid conflicting nodes.
+	 * @param preservedConnectedNodesToPins The list of preserved nodes for other nets.
+	 */
+	private void routeStaticNets(List<Node> preservedConnectedNodesToPins){
```
eddieh-xlnx, line 448: Looks like this is only ever called with null. Is this an oversight or can it be removed?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -0,0 +1,1841 @@
...
+		GlobalSignalRouting.setDesignRoutethruHelper(this.design, this.routethruHelper);
+		
+		for(Net net : this.staticNetAndRoutingTargets.keySet()){
+			for(SitePinInst sink : this.staticNetAndRoutingTargets.get(net)) {
+				this.preservedNodes.remove(sink.getConnectedNode());
+			}
+		}
+		
+		RouterHelper.invertPossibleGndPinsToVccPins(this.design, this.design.getGndNet());
```
eddieh-xlnx, line 457: Why inverting from GND to VCC desirable here?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -0,0 +1,1841 @@
...
+		Set<Node> unavailableNodes = getAllUsedNodesOfRoutedConnections();
+		unavailableNodes.addAll(this.preservedNodes.keySet());
+		// When the connections of other nets are not routed yet, 
+		// the nodes connected to pins of other nets should be supplied and preserved.
+		if(preservedConnectedNodesToPins != null) unavailableNodes.addAll(preservedConnectedNodesToPins);
+		unavailableNodes.addAll(this.rnodesCreated.keySet());
+		
+		for(Net net : this.staticNetAndRoutingTargets.keySet()){
+			System.out.println("INFO: Route " + net.getSinkPins().size() + " pins of " + net);
+			Map<SitePinInst, List<Node>> spiRoutedNodes = GlobalSignalRouting.routeStaticNet(net, unavailableNodes);
```
eddieh-xlnx, line 469: Looks like the map value of `staticNetAndRoutingTargets` is not used here?

src/com/xilinx/rapidwright/router/UltraScaleClockRouting.java
```diff
@@ -125,6 +126,117 @@ public static RouteNode routeToCentroid(Net clk, RouteNode clkRoutingLine, Clock
...
 	}
+	
+	/**
+	 * Routes a clock from a routing track to a given transition point called the centroid
+	 * @param clk The clock net to be routed
+	 * @param startingRouteNode The starting routing track
+	 * @param centroid The given centroid node
+	 * @return
+	 */
+	public static RouteNode routeToGivenCentroidNode(Net clk, RouteNode startingRouteNode, Node centroid) {
```
clavin-xlnx, line 137: Perhaps refactor this to `routeToCentroidNode()`

src/com/xilinx/rapidwright/router/UltraScaleClockRouting.java
```diff
@@ -125,6 +126,117 @@ public static RouteNode routeToCentroid(Net clk, RouteNode clkRoutingLine, Clock
...
+	/**
+	 * Routes a clock from a routing track to a transition point where the clock 
+	 * fans out and transitions from clock routing tracks to clock distribution.
+	 * @param clk The current clock net to contribute routing
+	 * @param startingRouteNode The intermediate start point of the clock route
+	 * @param clockRegion The center clock region or the clock region that is one row above or below the center
+	 * @param findCentroidHroute The flag to indicate the returned RouteNode should be HROUTE in the center or VROUTE going up or down
+	 */
+	public static RouteNode centroidHRouteNode = null;
+	public static RouteNode routeToCentroidHRouteOrVRouteAboveBelowCentroid(Net clk, RouteNode startingRouteNode, ClockRegion clockRegion, boolean findCentroidHroute) {
```
clavin-xlnx, line 187: This method name is a bit confusing.  Can you explain what it does exactly?  It seems like it is creating the centroid in the clock network.

src/com/xilinx/rapidwright/router/UltraScaleClockRouting.java
```diff
@@ -134,13 +246,29 @@ public static RouteNode routeToCentroid(Net clk, RouteNode clkRoutingLine, Clock
...
 		centroidRouteLine.setParent(null);
 		if(centroidRouteLine.getIntentCode() == IntentCode.NODE_GLOBAL_VDISTR){
 			return centroidRouteLine;
 		}
-		q.add(centroidRouteLine);
 		ClockRegion currCR = centroidRouteLine.getTile().getClockRegion();
+		return transitionCentroidToDistributoinLine(clk, centroidRouteLine, currCR);
+	}
+	
+	public static RouteNode transitionCentroidToUpDownDistributionLine(Net clk, RouteNode centroidRouteLine, boolean down){
```
clavin-xlnx, line 257: Instead of `UpDown` in the method name, could we replace it with `Vertical`?

src/com/xilinx/rapidwright/rwroute/Connection.java
```diff
@@ -0,0 +1,464 @@
...
+    	
+		if(tempCriticality > this.criticality)
+			this.setCriticality(tempCriticality);
+	}
+	
+	/**
+	 * Checks if a connection has any overused rnodes to indicate the congestion status.
+	 * @return
+	 */
+	public boolean congested() {
```
clavin-xlnx, line 210: Do you mean `isCongested()`?

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -0,0 +1,854 @@
...
+	static boolean clkDebug = false;
+	static boolean debugPrintClkPIPs = false;
+	
+	static Map<String, List<String>> crRoutes;
+	static Map<Pair<String, String>, List<Short>> bufceRowTaps;
+	static Device device;
+	
+	static Map<String, List<String>> dstINTtileRoutes;
+	static Map<String, Short> dstINTtileRoutey;
+	static ClkRouteTiming ceRouteTiming;
```
clavin-xlnx, line 74: Can we make these `private`?

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -0,0 +1,854 @@
...
+import com.xilinx.rapidwright.router.UltraScaleClockRouting;
+import com.xilinx.rapidwright.timing.ClkRouteTiming;
+import com.xilinx.rapidwright.util.Pair;
+
+/**
+ * A collection of methods for routing global signals, i.e. GLOBAL_CLOCK, VCC and GND.
+ * Adapted from RapidWright APIs.
+ */
+public class GlobalSignalRouting {
+	private static Design design;
```
clavin-xlnx, line 62: If this class is full of static methods, why does it have a reference to `Design`?  This would prevent anyone from being able to run two or more instances of `RWRoute` in the same process.

clavin-xlnx: I see a lot of signs that this class should be instantiated and not rely on static methods.  Perhaps it would be better to make it non-static and instantiate one per router?

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -0,0 +1,854 @@
...
+	
+	/**
+	 * Routes from a GLOBAL_VERTICAL_ROUTE to horizontal distribution lines.
+	 * @param clk The clock net to be routed.
+	 * @param vroute The node to start the route.
+	 * @param clockRegions Target clock regions.
+	 * @param down To indicate if is is routing to the group of top clock regions.
+	 * @return A list of RouteNodes indicating the reached horizontal distribution lines.
+	 */
+	private static List<RouteNode> routeFromDirectionalVRouteToHorizontalDistributionLines(Net clk, RouteNode vroute, List<ClockRegion> clockRegions, boolean down) {
```
clavin-xlnx, line 536: The name of this method seems like it should be in `UltraScaleClockRouting` and should probably be shortened a bit.  How does this method differ from those that route through a centroid as that seems to be what it does?

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -0,0 +1,854 @@
...
+	public static void setPrintCLKPIPs() {
+		debugPrintClkPIPs = true;
+	}
+	
+	/**
+	 * Routes a clk enable net with input data.
+	 * @param clockEnable The net to be routed.
+	 * @param device The design device.
+	 */
+	public static void clkEnableRoute(Net clockEnable, Device device) {
```
clavin-xlnx, line 114: This method should not take as an argument a`Device`.

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -0,0 +1,854 @@
...
+	
+	static boolean clkDebug = false;
+	static boolean debugPrintClkPIPs = false;
+	
+	static Map<String, List<String>> crRoutes;
+	static Map<Pair<String, String>, List<Short>> bufceRowTaps;
+	static Device device;
+	
+	static Map<String, List<String>> dstINTtileRoutes;
+	static Map<String, Short> dstINTtileRoutey;
```
clavin-xlnx, line 73: In camelCase, you'll want to capitalize each word.  So that would make these variables `dstINTTileRoutes` or `dstIntTileRoutes`, for example.  I also don't understand the naming `dstINTtileRoutey`.

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -0,0 +1,854 @@
...
+	 * @param clockEnable The net to be routed.
+	 * @param device The design device.
+	 */
+	public static void clkEnableRoute(Net clockEnable, Device device) {
+		Map<String, List<Node>> dstINTtilePaths = getListOfNodesFromRoutes(device, dstINTtileRoutes);
+		// Not import path after HDSTR
+		Set<PIP> ceNetPIPs = new HashSet<>();
+		Map<String, RouteNode> horDistributionLines = new HashMap<>();
+		
+		for(String intTile : dstINTtilePaths.keySet()) {
```
clavin-xlnx, line 120: When iterating over a map, it is preferable to use the `EntrySet` (see https://www.geeksforgeeks.org/difference-between-keyset-and-entryset-method-in-java-map/ for an example).  I see several instances of this throughout the code.  If you are iterating over a map and accessing both keys and values, please use the entry set.

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -0,0 +1,854 @@
...
+			return false;
+		}
+		return false;
+	}
+	
+	/**
+	 * A lightweight useful class for different simple routing-related scenarios, 
+	 * each {@link RoutingNode} Object is associated to a {@link Node} Object.
+	 */
+	static class RoutingNode{
```
clavin-xlnx, line 803: Let's refactor this outside into its own file.

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -0,0 +1,854 @@
...
+public class GlobalSignalRouting {
+	private static Design design;
+	private static RouteThruHelper routeThruHelper;
+	
+	static boolean clkDebug = false;
+	static boolean debugPrintClkPIPs = false;
+	
+	static Map<String, List<String>> crRoutes;
+	static Map<Pair<String, String>, List<Short>> bufceRowTaps;
+	static Device device;
```
clavin-xlnx, line 70: There is a reference to a `Device` here, but there are many methods that take as a parameter a `Device`.  This is confusing.  Again, we probably need to discuss this class -- if it should be static or not.  There are ways of getting a `Device` object (from a `Tile` for example) without having to require the user to provide it.  We can also add helper methods to avoid cumbersome APIs.

src/com/xilinx/rapidwright/rwroute/PartialRouter.java
```diff
@@ -0,0 +1,93 @@
...
+			this.createsNetWrapperAndConnections(net, this.config.getBoundingBoxExtensionX(), this.config.getBoundingBoxExtensionY(),multiSLR);
+		}else{
+			// In partial routing mode, a net with PIPs is preserved.
+			// This means the routed net is supposed to be fully routed without conflicts.
+			// TODO detect partially routed nets.
+			this.preserveNet(net);
+			this.increaseNumPreservedWireNets();
+		}
+	}	
+}
```
clavin-xlnx, line 93: Add a new line at the end?

src/com/xilinx/rapidwright/rwroute/ReportTimingAndWirelength.java
```diff
@@ -0,0 +1,196 @@
...
+		System.out.println("\n");
+		System.out.println("Total nodes: " + this.usedNodes);
+		System.out.println("Total wirelength: " + this.wirelength);	
+		RWRoute.printNodeTypeUsageAndWirelength(true, this.nodeTypeUsage, this.nodeTypeLength);
+	}
+	
+	/**
+	 * Computes the wirelength and delay for each net.
+	 */
+	private void computeWLDlyForEachNet() {
```
clavin-xlnx, line 95: `computeWLDlyForEachNet()` -> `computeNetsWirelengthAndDelay()` perhaps?

src/com/xilinx/rapidwright/rwroute/RouteFixer.java
```diff
@@ -0,0 +1,212 @@
...
+		public int hashCode(){
+			return this.node.hashCode();
+		}
+		
+		@Override
+		public String toString(){
+			return this.id + ", " + this.node.toString() + ", delay = " + this.delay + ", sink? " + this.isSink;
+		}
+	}
+}
```
clavin-xlnx, line 212: Add a new line at end of file?

src/com/xilinx/rapidwright/timing/CLKSkewRouteDelay.java
```diff
@@ -0,0 +1,175 @@
...
+
+import com.xilinx.rapidwright.util.Pair;
+
+/**
+ * Creating a CLKSkewRouteDelay instance parses a supplied text file and stores clock skew, route, and delay data into maps to use.
+ * To obtain the text file, please refer to find_clock_tree_template.tcl under $RAPIDWRIGHT_PATH/tcl/rwroute.
+ * When the file is ready, please use "--clkSkew>" option (see {@link Configuration})
+ * to enable RWRoute to use the file for clock skew-aware timing-driven clock routing.
+ */
+public class CLKSkewRouteDelay {
```
clavin-xlnx, line 43: Since Clk is not an acronym, let's change it to `ClkSkewRouteDelay`, although, perhaps there is a better name?

src/com/xilinx/rapidwright/timing/DelayModelBuilder.java
```diff
@@ -108,24 +108,66 @@ private static void newDelayModel(String mode, String source, String fileName) {
     /**
      * For unit testing.
      */
-    private static int testLogicDelay(DelayModel a, String belName, List<String> config, 
-            String[] src, String[] dst) {
-
+//    private static int testLogicDelay(DelayModel a, String belName, List<String> config,
```
clavin-xlnx, line 111: Let's remove this method or uncomment it.  It won't get maintained as a comment.

src/com/xilinx/rapidwright/timing/TimingGraph.java
```diff
@@ -438,13 +856,148 @@ public void generateGraphvizDotVisualization(String dotFileName) {
...
+    	bramPinsToSuperSink.add("ENARDEN");
+    	bramPinsToSuperSink.add("ENBWREN");
+    	bramPinsToSuperSink.add("WEA");
+    	bramPinsToSuperSink.add("WEBWE");
+    	// that CASDIN goes to FF or not depends on the MUX
+    	// CASOREGIMUX == 1 (vcc) { CASDOMUX == 0 (gnd), do not connect, otherwise connect to superSink}
+    	// CASOREGIMUX && CASDOMUX from signals (not vcc/gnd), connect to superSink
+    }
+    
+    private boolean shouldBRAMInputConnectToSuperSink(Cell cell, String cellPinName) {
```
clavin-xlnx, line 885: What is a super sink?

src/com/xilinx/rapidwright/timing/TimingManager.java
```diff
@@ -53,32 +72,343 @@ public TimingManager(Design design) {
...
+    	
+    	System.out.printf("%-30s\n", "With timing closure guarantee:");
+    	short adjusted = (short) (pessimismA * (arr - criticalEdges.get(0).getDelay() - clkskew) + pessimismB);
+    	System.out.printf("%-30s %10d\n", "Critical path delay (ps):", (short)adjusted);
+    	System.out.printf("%-30s %10d\n\n", "Slack (ps):", (short) (Treq - adjusted));
+    	
+    	this.printPathDelayBreakDown(arr, criticalEdges, timingEdgeConnctionMap, useRoutable, rnodesCreated);
+    }
+    
+    private String getSrcClockRegion() {
```
clavin-xlnx, line 240: This code looks hard coded to a specific use case.  Do we still need it?

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -76,63 +79,63 @@
 
     // some default values, these can be overwritten later by reading in a delay_terms.dat file
     // the code is using hard coded enumerated types, and this will be switched over to use these variables
-    int START_TILE_ROW =		1;
-    int START_TILE_COL =		52;
+    int START_TILE_ROW = 1;
+    int START_TILE_COL = 52;
```
clavin-xlnx, line 83: Should these be `public static`?

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -76,63 +79,63 @@
...
+    float CFRM_LONG = 33.f;
+    float URAM_SINGLE_AND_DOUBLE = 34.f;
+    float URAM_QUAD = 34.f;
+    float URAM_LONG = 34.f;
+    float PCIE_SINGLE_AND_DOUBLE = 62.f;
+    float PCIE_QUAD = 62.f;
+    float PCIE_LONG = 62.f;
+    float IO_SINGLE_AND_DOUBLE = 86.f;
+    float IO_QUAD = 68.f;
+    float IO_LONG = 186.f;
```
clavin-xlnx, line 138: Should these be `public static`?

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -536,6 +578,13 @@ protected boolean readDelayTerms(String filename) {
                 else if (split[0].equalsIgnoreCase("SITEPIN_A6_DELAY"))        SITEPIN_A6_DELAY = value;
                 else if (split[0].equalsIgnoreCase("SITEPIN_A_I_DELAY"))        SITEPIN_A_I_DELAY = value;
                 else if (split[0].equalsIgnoreCase("SITEPIN_AX_DELAY"))        SITEPIN_AX_DELAY = value;
+                else if (split[0].equalsIgnoreCase("SITEPIN_CKEN1_DELAY"))     SITEPIN_CKEN1_DELAY = value;
+                else if (split[0].equalsIgnoreCase("SITEPIN_CKEN2_DELAY"))     SITEPIN_CKEN2_DELAY = value;
+                else if (split[0].equalsIgnoreCase("SITEPIN_CKEN3_DELAY"))     SITEPIN_CKEN3_DELAY = value;
+                else if (split[0].equalsIgnoreCase("SITEPIN_CKEN4_DELAY"))     SITEPIN_CKEN4_DELAY = value;
+                else if (split[0].equalsIgnoreCase("SITEPIN_SRST1_DELAY"))     SITEPIN_SRST1_DELAY = value;
+                else if (split[0].equalsIgnoreCase("SITEPIN_SRST2_DELAY"))     SITEPIN_SRST2_DELAY = value;
+                else if (split[0].equalsIgnoreCase("SITEPIN_WCKEN_DELAY"))     SITEPIN_WCKEN_DELAY = value;
```
clavin-xlnx, line 587: Please reimplement with a Map.

src/com/xilinx/rapidwright/timing/delayestimator/DelayEstimatorBase.java
```diff
@@ -0,0 +1,461 @@
...
+	    	return inputSitePinDelay.getOrDefault(exitNode.getWireName(), (short) 0);
+	    }
+	    
+	    return calcTimingGroupDelay(node.ng, node.begin(), node.end(), 0d);
+    }
+
+    /**
+     * Represent a routing resource in the delay graph
+     */
+    protected class RoutingNode {
```
clavin-xlnx, line 130: There is another inner class called `RoutingNode`. Do we need two?  Can we unify them?  I'd like to avoid inner classes.

src/com/xilinx/rapidwright/timing/delayestimator/DelayEstimatorBase.java
```diff
@@ -0,0 +1,461 @@
...
+        }
+
+        public short end() {
+            short delta = ng.length();
+            short start = ng.orientation() == T.Orientation.HORIZONTAL ? x : y;
+            return  (short) (start + (direction == InterconnectInfo.Direction.U ? delta : -delta));
+        }
+    }
+    
+    private short getLengthFromIntentCode(IntentCode ic) {
```
clavin-xlnx, line 179: This could probably be moved to a more generic location.

src/com/xilinx/rapidwright/timing/delayestimator/DelayEstimatorBase.java
```diff
@@ -0,0 +1,461 @@
...
+    private RoutingNode getTermInfo(Node node) {
+        IntentCode ic = node.getIntentCode();
+        RoutingNode rNode = new RoutingNode();
+
+        rNode.x = (short) node.getTile().getTileXCoordinate();
+        rNode.y = (short) node.getTile().getTileYCoordinate();
+
+
+        String nodeType = node.getWireName();
+        if (nodeType.contains("WW1_E")) {
```
clavin-xlnx, line 206: This seems quite specific to a use case.  Please add a comment explaining its need or remove.

src/com/xilinx/rapidwright/timing/delayestimator/DelayEstimatorBase.java
```diff
@@ -0,0 +1,461 @@
...
+
+
+
+
+
+
+
+
+
+
```
clavin-xlnx, line 461: Let's remove the extra whitespace, but leave a newline at the end.

src/com/xilinx/rapidwright/timing/delayestimator/InterconnectInfo.java
```diff
@@ -0,0 +1,293 @@
...
+/**
+ * Encapsulate interconnect information for a device family.
+ * The info must be just accurate enough for estimating the min delay.
+ * This is for Ultrascale+.
+ */
+// TODO: Consider moving this to a file.  Leaving them in code is easier to be extended for future generations.
+// TODO: The key issue is it is not certain what features are enough to capture future generations.
+public class InterconnectInfo implements java.io.Serializable {
+
+    // TODO: Can we get this association from device model?
```
clavin-xlnx, line 41: The answer to this TODO is yes - `Node.getSitePin()`.

src/com/xilinx/rapidwright/timing/delayestimator/InterconnectInfo.java
```diff
@@ -0,0 +1,293 @@
...
+    private Map<String,String> nodeToSitePin;
+
+
+    /**
+     * Define direction of routing resource in reference to INT tile coordinate.
+     */
+    // up is in increasing INT tile coordinate direction
+    public enum Direction {
```
clavin-xlnx, line 49: Can we standardize this enum somewhere more general?  Are there some enums that could be merged to serve the same purpose?

src/com/xilinx/rapidwright/timing/delayestimator/InterconnectInfo.java
clavin-xlnx: This entire file should be removed.  All the enums should be merged with others or if still needed, moved to their own file.  The Map should not exist and code can be written to get the SitePin directly (or build a map in a more relevant class).

src/com/xilinx/rapidwright/util/Timer.java
```diff
@@ -0,0 +1,197 @@
+/*
```
clavin-xlnx, line 1: This class appears to measure execution runtime.  Perhaps we can find a more descriptive class name.

src/com/xilinx/rapidwright/util/rwroute/GetDelayFromSourceToSinkINT.java
```diff
@@ -0,0 +1,101 @@
...
+import com.xilinx.rapidwright.util.Pair;
+
+/**
+ * A helper class to write delay from the source to the sink of a routed net
+ * (Usage: input.dcp --net net_name outputFilePath --allSinkDelay).
+ * When the "--allSinkDelay" option specified, it writes the delay values from 
+ * the source to the sinks (until the interconnect tile node) of the net under the name.
+ * Otherwise, it is specifically for the CLK_IN sink of the net under the name.
+ */
+public class GetDelayFromSourceToSinkINT {
```
clavin-xlnx, line 46: Perhaps we could find a more descriptive class name for what this does.

clavin-xlnx (changes requested):
@YunxZhou - I've made a first pass through the code.  Overall, we want to keep moving forward and continue to improve RWRoute and refine it.  So, let's keep working on the outstanding comments and suggestions.  Once we can resolve most or all of those, we'll go through it again for a second pass.

Also, one other high level comment is that we will need to add some regression tests that can be run on RWRoute during CI (continuous integration).  These are tests that would be run with each change to RapidWright in the repo.  They would be very small and depend on some existing DCPs that can be uploaded and made available.  @eddieh-xlnx has graciously offered his support to help you with the details of getting the tests up and running.  You'll notice a `test` directory which is a direct sibling to `src` in RapidWright.  This is where the test code would go and would leverage the JUnit5 framework for testing.  Ideally, you won't have to do much, just create some small tests (small DCPs) that would run RWRoute in all the various modes (timing-driven, non-timing driving, full design vs incremental routing, etc).

src/com/xilinx/rapidwright/rwroute/RoutableNode.java
YunxZhou: We noticed the case and it is true that the end tile found by the code can be incorrect for nodes that have two exit points. But it would not hurt the bounding box computing or the routing. It is because [every output / input pin has been mapped to its connecting INT Tile](https://github.com/YunxZhou/RapidWright/blob/a5cdce2c2d651f57c378e051e8534596e93a0417/src/com/xilinx/rapidwright/rwroute/RWRoute.java#L527), considering that each pin connects to a unique node. The mapping provides a regular coordinate system of routing resources of the architecture (using coordinates of INT tiles only). This means the CLEL_R_X6Y84 tile coordinates would not matter when computing the connection bounding box.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
YunxZhou: I understand the point and appreciate the discussion. Updating the method javadoc to indicate the limitation has been done. 

If you would like to experiment with the partial routing feature and need the router to rip up as little as possible, the current preserving mechanism would need to be changed. For instance, it makes all used routing resources of the preserved nets invisible during the routing process, which impedes the negotiation among routing-needing nets and preserved nets.

src/com/xilinx/rapidwright/rwroute/RoutableNode.java
YunxZhou: True. addUser changed to increaseConnectionCountOfUser.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
YunxZhou: Merged.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
YunxZhou: Not really. Removed.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
YunxZhou: It is because of the hardware. There are invertible BELs related to those pins. Without this inverting, Vivado would complain about the routing of GND and VCC.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
YunxZhou: They are not used here. They are used in printing design info and checking the number of static net pins that should be routed. It helps identifying problems if any.

src/com/xilinx/rapidwright/router/UltraScaleClockRouting.java
YunxZhou: Refactored.

YunxZhou:
@eddieh-xlnx and @clavin-xlnx, thanks a lot for the review and suggestions!

I believe the new commits cover all the requested changes according to your helpful comments, except the TimingModel, IntereConnectInfo and DelayEstimatorBase class files. They were not written by me and I am not quite sure about them, so I did not touch those files. 

@eddieh-xlnx Thanks for helping with setting up tests. I would send you example DCPs later.

PR #231, opened 2021-09-18
MetadataParser to close BufferedReader

eddieh-xlnx:


src/com/xilinx/rapidwright/design/MetadataParser.java
```diff
@@ -119,7 +119,7 @@ public String getOriginalInstName(String metadataFileName){
 		return null;
 	}
 	
-	public void parse(String metadataFileName){
+	public void parse(String metadataFileName) throws Exception {
```
clavin-xlnx, line 122: Thanks @eddieh-xlnx for the PR.  One thing to note is that this class (it was written some time ago) will eventually be removed.  I believe we have enough information in a DCP that we can remove it as soon as someone has the bandwidth to do so.  

Having said that, if you want a method to throw an exception, its best to specify which one you could potentially throw (`IOException` in this case).  In most of RapidWright, however, there aren't many recoverable situations once an error is thrown, so typically, the pattern would be to try/catch and print an error and stack trace right there.

eddieh-xlnx: Sorry for being lazy! I'm now catching the IOException and throwing a RuntimeException instead, which matches other situations in this file. Would this be preferable?

PR #235, opened 2021-09-23
Add macro unisim expansion test

eddieh-xlnx:


eddieh-xlnx:
Pulls in and hence depends on #235

PR #236, opened 2021-09-23
Introduce relocatable rectangle of tiles

jakobwenzel:
The class TileRectangle I introduced in #166 was not relocatable. This PR adds the class `RelocatableTileRectangle` which stores references to the actual tiles and can therefore be relocated across the fabric.

This new class is slightly slower than the old one. If you don't need relocatability, the old one should still be used. Therefore, I kept it under the name `SimpleTileRectangle`.


clavin-xlnx:
Thanks!

PR #240, opened 2021-09-24
Adds CFGLUT5 and BUFG to macros.

clavin-xlnx:


clavin-xlnx:
@eddieh-xlnx I think by checking out this branch, you should get the new data file automatically when you run (you can compare the md5sum after running to be sure it should match the diff).

eddieh-xlnx (approved):
Works great. Unblocks #235 and workaround in eddieh-xlnx/RapidWright#1 too!

PR #245, opened 2021-09-28
Add rapidwright.design.tools.RelocationTools

eddieh-xlnx:
RelocationTools is able to relocate in-place the placement and routing of a subset of a design  by a given set of tile rows and columns, without requiring that this subset belong to a `ModuleInst`.

It provides three `relocate()` overloads, accepting:
1. A logical cell instance hierarchical name
2. A `PBlock`
3. Collection of `SiteInst`s
and will duly attempt to relocate all `SiteInst`s within, error-ing out if relocation is not possible (e.g. if corresponding site types do not exist at the destination)
*(Case 1 will also error out if the leaf cells of this logical cell instance do not exclusively occupy its associated `SiteInst`s)*
Any `Net` branches to `SiteInstPin`s not present on an included `SiteInst` will be unrouted, as will clock nets.

Tests are also provided. The input to these tests are DCP files which are provided as part of a new RapidWrightDCP submodule.


eddieh-xlnx:
@clavin-xlnx Please add "needs new release" label

src/com/xilinx/rapidwright/design/tools/RelocationTools.java
```diff
@@ -0,0 +1,272 @@
+package com.xilinx.rapidwright.design.tools;
+
+import java.util.*;
```
clavin-xlnx, line 3: Let's replace `import *`-style with explicit types.

src/com/xilinx/rapidwright/design/tools/RelocationTools.java
```diff
@@ -0,0 +1,272 @@
+package com.xilinx.rapidwright.design.tools;
```
clavin-xlnx, line 1: With new source files, we should probably add the license header https://github.com/Xilinx/RapidWright/blob/master/doc/SOURCE_HEADER.TXT

clavin-xlnx: A setting in the IDE can probably do this automatically.

src/com/xilinx/rapidwright/design/tools/RelocationTools.java
```diff
@@ -0,0 +1,272 @@
...
+        Map<SiteInst, Site> oldSite = new HashMap<>();
+        for (SiteInst si : siteInsts) {
+            assert(si.isPlaced());
+            oldSite.put(si, si.getSite());
+            si.unPlace();
+        }
+
+        boolean revertPlacement = false;
+        for (Map.Entry<SiteInst, Site> e : oldSite.entrySet()) {
+            Site ss = e.getValue();
```
clavin-xlnx, line 160: The variable names `ss, st, dt, ds, ssi, dsi`, I think they mean `srcSite, srcTile, dstTile, dstSite, srcSiteInst, dstSiteInst`, is that correct?  If so, probably no change needed.

clavin-xlnx (changes requested):
Looks good!  Mostly just fix up license headers and imports.  Thanks!

src/com/xilinx/rapidwright/design/tools/RelocationTools.java
eddieh-xlnx: Ooh yes, sorry about that -- you've told me off before! I've changed my IDE to stop doing this now.

clavin-xlnx (approved):

Issue #248, opened 2021-10-01
Calling EDIFNetlist.getSinksFromNet() results in NPE when internal cells ports are unconnected


clavin-xlnx:
While debugging `DesignTools.copyImplementation()`, a specific cell had an input port that was unconnected internally.  When traversing the netlist to identify sinks, `EDIFNetlist.getSinksFromNet()` throws an NPE because the inner net was unconnected (likely removed from synthesis or opt_design in Vivado).

PR #249, opened 2021-10-01
Guard against null internal nets. Fix for #248

clavin-xlnx:


clavin-xlnx:
This is a fix for issue #248.

PR #256, opened 2021-10-11
performance evaluation: find true maximum frequency of designs

jakobwenzel:
This PR adds a tool that takes a unrouted or partially routed design and tries to find the true maximum frequency. The design is routed a number of times in Vivado with varying timing constraints.


clavin-xlnx:
Thanks, looks good!

Issue #226, opened 2021-08-25
LUT route-thru equation gets altered if a net passes it twice (LUT5 and LUT6)


mkurc-ant:
The issue surfaces for the `Murax` SoC design routed using `nextpnr-interchange` for which the routed physical netlist was converted by RapidWright to `dcp` and then read into Vivado. Possibly it shows up in other more complex designs too.

There are at least 3 nets there that are routed according to the following scheme: driver -> routing -> LUT6 -> routing -> LUT5 -> sink. Both LUT6 and LUT5 are route-thru and are at the same LUT BEL of a SLICE. The picture below shows an example of the situation for the ALUT BEL of a SLICE:

![image](https://user-images.githubusercontent.com/47315577/130790306-d40d0c82-5c32-4a4f-a9d1-eb0d5ba2a085.png)

![image](https://user-images.githubusercontent.com/47315577/130789589-b76dcccc-cb85-444d-8d4e-90400b9ccb4d.png)

As it is visible in the picture RapidWright / Vivado did not preserve the A4 to O6 and A5 to O5 connections thinking that if they belong to the same net then the LUT equation can be altered so that the net enters only via A5 and forks to O5 and O6 there. However, that would work only if both A4 and A5 pins were driven by the same driver but in this case they are not and altering the LUT equation breaks the route. Ultimately the design does not work in hardware correctly.

Apart from the net entering and exiting the SLICE twice being suboptimal, neither RapidWright nor Vivado should change the LUT equation. They should preserve what's present in a physical netlist.

eddieh-xlnx:
Test for this scenario now in #262.

PR #262, opened 2021-10-15
Add design.DRC, with single check: NetRoutesThruLutAtMostOnce

eddieh-xlnx:
Warns if situation in #226 is encountered, error if `--strict`.

Proposed infrastructure for an extensible DRC based on a stream of independent checks each implemented as its own class. Potential to make this `stream()` into a `parallelStream()` in the future.


src/com/xilinx/rapidwright/design/DRC.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import com.xilinx.rapidwright.design.drc.NetRoutesThruLutAtMostOnce;
+
+/**
+ * Parent DRC that executes a list of child DRCs, returning the sum of all failed checks.
+ */
+public class DRC {
+    interface DrcStyleFunction {
```
clavin-xlnx, line 34: Perhaps this interface should be declared in its own file, `design/drc/DrcStyleFunction.java` and explicitly declare its scope?

src/com/xilinx/rapidwright/design/drc/NetRoutesThruLutAtMostOnce.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Check that each LUT contains at most one routethru of each net.
+ * Identifies occurrences of issue #226.
+ * Failed checks will print a warning and are not counted unless the strict parameter is true.
+ */
+public class NetRoutesThruLutAtMostOnce {
```
clavin-xlnx, line 46: Should `NetRoutesThruLutAtMostOnce` implement the `DrcStyleFunction` interface?

src/com/xilinx/rapidwright/design/drc/NetRoutesThruLutAtMostOnce.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private static String lutName(Node node) {
```
clavin-xlnx, line 48: Perhaps a bit more description in the name, such as `makeUniqueLutName()`?

src/com/xilinx/rapidwright/design/DRC.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.xilinx.rapidwright.design.drc.NetRoutesThruLutAtMostOnce;
+
+/**
+ * Parent DRC that executes a list of child DRCs, returning the sum of all failed checks.
+ */
+public class DRC {
```
clavin-xlnx, line 33: Should DRC implement `DrcStyleFunction`?

src/com/xilinx/rapidwright/design/drc/NetRoutesThruLutAtMostOnce.java
```diff
@@ -0,0 +1,108 @@
...
+                        lutRoutethrus.merge(lutName(p.getStartNode()), 1, Integer::sum);
+                    }
+
+                    for (SitePinInst spi : n.getSinkPins()) {
+                        for (BELPin bp : spi.getSiteWireBELPins()) {
+                            Cell c = spi.getSiteInst().getCell(bp.getBEL());
+                            // Filter out all non routethru cells
+                            if (c == null || !c.isRoutethru()) continue;
+
+                            assert(DesignTools.isBELALut(bp.getBELName()));
```
clavin-xlnx, line 70: Since asserts are not turned on by default, will this catch the same issues if the assertion is not checked?

src/com/xilinx/rapidwright/design/drc/NetRoutesThruLutAtMostOnce.java
```diff
@@ -0,0 +1,108 @@
...
+
+                            // Check that is pin is actually used by this BEL
```
clavin-xlnx, line 72:
```suggestion
                            // Check that the pin is actually used by this BEL
```

src/com/xilinx/rapidwright/design/DRC.java
```diff
@@ -0,0 +1,83 @@
...
+                    } catch (Exception e) {
+                        throw new RuntimeException(e);
+                    }
+                })
+                .reduce(0, Integer::sum);
+    }
+
+    private static void printUsageAndExit() {
+        System.out.println("USAGE: <input.dcp> [--strict]");
+        System.exit(1);
```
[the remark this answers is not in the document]
clavin-xlnx: The `Security Manager` is being [deprecated in Java 17](https://openjdk.java.net/jeps/411) and will be eventually removed.  There is a possibility of [a new API that could catch `System.exit()` calls](https://bugs.openjdk.java.net/browse/JDK-8199704), but no activity on it for a while.  If we do go down this path, it will dispense warning messages and will get progressively harder to use in later versions of Java.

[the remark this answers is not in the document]
clavin-xlnx: This is an example of the warning message users would see each time they `import rapidwright`:
```
WARNING: A terminally deprecated method in java.lang.System has been called
WARNING: System::setSecurityManager has been called by com.xilinx.rapidwright.util.PythonSecurityManager (file:/C:/home/workspace/RapidWright/bin/)
WARNING: Please consider reporting this to the maintainers of com.xilinx.rapidwright.util.PythonSecurityManager
WARNING: System::setSecurityManager will be removed in a future release
```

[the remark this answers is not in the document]
eddieh-xlnx: Moved conversation to #283

src/com/xilinx/rapidwright/design/DRC.java
eddieh-xlnx: This follows the same pattern as https://github.com/Xilinx/RapidWright/blob/3a6e8fc965926b1dd9bf30b936a052f32b46e9d6/src/com/xilinx/rapidwright/MainEntrypoint.java#L75-L77 in that it's a "functional interface" (rather than a "class interface") which allows it to represent static methods.

My earlier approach used an abstract class interface that forced subclasses to implement the `run` function, but this approach required (a) a new instance of the subclass to be created in order to run the DRC, and (b) reflection in order to get a handle to its `run` method to be invoked.

I believe this approach alleviates both of these problems.

src/com/xilinx/rapidwright/design/drc/NetRoutesThruLutAtMostOnce.java
eddieh-xlnx: My thinking here was to (a) subtly indicate to the reader that I am only expecting/assuming a LUT here and (b) enforce that with a string-based check only if assertions were enabled, but not pay for that if assertions were disabled.

Would it be an extra-exceptional (extra'ceptional?) situation if this assumption was to break down?

clavin-xlnx: What would happen if the routethru cell was not a LUT?  I think this case is common.  Perhaps I add a `BEL.isLUT()` method as part of #277?

eddieh-xlnx: I actually discovered that route-thru PIPs exist on IOLOGICs as well, which were being incorrectly considered before and is now fixed.

I think a fast BEL.isLUT() would be very helpful!

clavin-xlnx (approved):

PR #263, opened 2021-10-15
Part{,NameTools} to return int for appropriate properties 

### LINKED ISSUE #73: Request: Get number of available (as reported in documentation) device resources
First off, I apologize if this feature already exists in RapidWright, but I was unable to find it.

For some parts, there seems to be a difference between the number of a given resource you can use and the number that actually exist on the part. To explain the problem better I'll use the XAZU2EG part as an example. 

The [documentation ](https://www.xilinx.com/support/documentation/user_guides/ug573-ultrascale-memory-resources.pdf) reports that there are 240 DSP slices the XAZU2EG can use. This is also what Vivado reports if `report_utilization` is used. However, if I use Vivado to find all the DSP slices in the device, it will return a list of 360 DSP slices. Similarly, RapidWright reports that 360 DSP slices exist, which is correct. However, Vivado will actually only let me place 240 DSP slices on the device. This same behavior can be found with BRAMs, etc.

From RapidWright, I would like to be able to get the number of resources that Vivado will ultimately let me use (240) rather than the number that exist on the device (360). Is this already possible? If not, it would be a useful feature. Thanks for your time.

eddieh-xlnx:
Fixes #73


clavin-xlnx:
Technically, this needs a new `parts.db` data file release, which can be async to a release.

PR #267, opened 2021-10-21
[python] Do not use standalone jar if RAPIDWRIGHT_PATH set

eddieh-xlnx:
Helpful for dev while not affecting out of the box experience.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -6,7 +6,7 @@ pip install rapidwright
 ```
 
 # How it Works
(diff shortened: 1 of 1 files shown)


clavin-xlnx:
> > However, how does JPype get the `CLASSPATH` associated with the developer's RapidWright repo with this change?
> 
> I believe the JVM uses `CLASSPATH` from the environment, which for me contains all the necessary bits to launch RW from the console prompt, and so when I tested this change locally (with and without it set) it appeared to work for me.

Ok, that makes sense.  We should probably document this behavior.  Would this page be the right place?   
https://github.com/Xilinx/RapidWright/blob/master/python/README.md

PR #279, opened 2021-11-01
gradle: add group and description to update_jars task

jakobwenzel:
This makes the update_jars task visible in Gradle's task list (can be viewed by running `./gradlew tasks`).


clavin-xlnx:
Thanks!

PR #280, opened 2021-11-02
Site routing resolver when combining in-context module designs.

clavin-xlnx:

### DIFF
--- a/test/com/xilinx/rapidwright/design/TestSiteRoutingResolver.java
+++ b/test/com/xilinx/rapidwright/design/TestSiteRoutingResolver.java
@@ -1,0 +1,55 @@
+package com.xilinx.rapidwright.design;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import com.xilinx.rapidwright.device.BELPin;
+import com.xilinx.rapidwright.support.CheckOpenFiles;
+import com.xilinx.rapidwright.support.RapidWrightDCP;
+import com.xilinx.rapidwright.tests.CodePerfTracker;
+import com.xilinx.rapidwright.util.Pair;
+
+public class TestSiteRoutingResolver {
+
+    private Pair<String,String> inputSiteWire1 = new Pair<>("SLICE_X16Y238","A2"); 
+    
+    private Pair<String,String> inputSiteWire2 = new Pair<>("SLICE_X13Y237","F5");
+    
+    @SuppressWarnings("unchecked")
+    private Map<Pair<String,String>,String> mimicInContextInputPortNetSiteRouting(Design design) {
+        Map<Pair<String,String>,String> initialState = new HashMap<>();
+        
+        for(Pair<String,String> siteWire : new Pair[] {inputSiteWire1, inputSiteWire2}) {
+            SiteInst i = design.getSiteInstFromSiteName(siteWire.getFirst());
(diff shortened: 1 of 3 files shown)


test/com/xilinx/rapidwright/design/TestSiteRoutingResolver.java
```diff
@@ -0,0 +1,55 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import com.xilinx.rapidwright.device.BELPin;
+import com.xilinx.rapidwright.support.CheckOpenFiles;
+import com.xilinx.rapidwright.support.RapidWrightDCP;
+import com.xilinx.rapidwright.tests.CodePerfTracker;
+import com.xilinx.rapidwright.util.Pair;
+
+public class TestSiteRoutingResolver {
```
eddieh-xlnx, line 16: Should we stick this inside `TestDesignTools.java` (it might be a good idea to follow the same structure as `:/src`) which could eventually end up as a big monolithic file, but does not impact the ability to run specific tests, or should we have a `DesignTools` subdir where this can sit?

test/com/xilinx/rapidwright/design/TestSiteRoutingResolver.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private Pair<String,String> inputSiteWire1 = new Pair<>("SLICE_X16Y238","A2"); 
+    
+    private Pair<String,String> inputSiteWire2 = new Pair<>("SLICE_X13Y237","F5");
+    
+    @SuppressWarnings("unchecked")
```
eddieh-xlnx, line 22: What warning does this suppress?

eddieh-xlnx (commented):
Is it worth having a DRC (pending #262) that checks for these conflicts and prompts the user to run `DesignTools.resolveSiteRoutingFromInContextPorts()`?

test/com/xilinx/rapidwright/design/TestSiteRoutingResolver.java
clavin-xlnx: I thought about that, but in the past it was mentioned that we should break up the tests to facilitate parallelization.

test/com/xilinx/rapidwright/design/TestSiteRoutingResolver.java
clavin-xlnx: Creating an array of `Pair<String,String>`.  Not sure why its a problem.

clavin-xlnx:
> Is it worth having a DRC (pending #262) that checks for these conflicts and prompts the user to run `DesignTools.resolveSiteRoutingFromInContextPorts()`?

Not a bad idea.

test/com/xilinx/rapidwright/design/TestSiteRoutingResolver.java
eddieh-xlnx: I think there's still opportunity for parallelization at the class granularity given I expect that (eventually) there will be many classes.

test/com/xilinx/rapidwright/design/TestSiteRoutingResolver.java
eddieh-xlnx: Would it be this line `new Pair[] {inputSiteWire1, inputSiteWire2}` perchance? Would `new Pair<String,String>` help? Otherwise I've done `Arrays.asList(...)` in the past too.

eddieh-xlnx (approved):

PR #284, opened 2021-11-05
Catch System.exit() calls and throw an error instead for Python REPL

clavin-xlnx:


clavin-xlnx:
This change will catch `System.exit()` calls for Python REPLs when using Java 16 and below.  By default, it will not be turned on for Java 17 since it prints an ugly WARNING message.  Those Java 17 users who truly need this functionality can run: 
```
from com.xilinx.rapidwright.util import BlockExitSecurityManager
BlockExitSecurityManager.blockSystemExitCalls()
```

Hopefully in the future, an API will be made available that can work around this issue, however, Java 17, at least is doomed to always have a WARNING come out in order to block `System.exit()` calls.

src/com/xilinx/rapidwright/util/BlockExitSecurityManager.java
```diff
@@ -0,0 +1,42 @@
...
+ * @see <a href="JDK-8199704">https://bugs.openjdk.java.net/browse/JDK-8199704</a>
+ */
+@SuppressWarnings("removal")
+public class BlockExitSecurityManager extends SecurityManager {
+    
+    private static BlockExitSecurityManager singleton = null; 
+    
+    @Override 
+    public void checkExit(int status) {
+        throw new SecurityException("Process Exited with Return Value: " + status);
```
eddieh-xlnx, line 19:
```suggestion
        throw new SecurityException("Java VM terminated with status: " + status);
```

eddieh-xlnx (approved):

PR #289, opened 2021-11-10
Gradle to init submodules on test

eddieh-xlnx:
Added two new gradle tasks:
* `updateSubmodules` -- running `git submodule update --init --recursive`
* `initSubmodules` -- which depends on `updateSubmodules` only if `test/RapidWrightDCP/.git` does not exist, which is expected to be the case for a fresh `git clone` (non recursive); thus it is expected to only have any effect on this first occasion.

`test` and `testPython` targets now depend on `initSubmodules` which should reduce initial friction for developers.
`updateSubmodules` remains a task that developers can run manually to sync their submodules too, as long as there are no conflicts within.


eddieh-xlnx:
I believe this change would -- on `test` -- forcibly checkout all submodules to the checked-in reference, even if a developer intentionally moved the reference to something else. To avoid this action, said developer would need to commit their submodule reference.

I don't think it seems straightforward to tell the difference between a developer intentionally moving the ref, versus a `git pull` or other action moving it...

eddieh-xlnx:
Also, this should work on Windows too, as long as `git` is on the `PATH`?

eddieh-xlnx:
> Also, this should work on Windows too, as long as `git` is on the `PATH`?

Confirmed to by merging in #291.

clavin-xlnx (approved):

build.gradle
```diff
@@ -68,6 +68,16 @@ abstract class UpdateJarsTask extends DefaultTask {
 // Create a task using the task type
 tasks.register('update_jars', UpdateJarsTask)
 
+task updateSubmodules(type:Exec) {
```
clavin-xlnx, line 71: Not sure if I understand Gradle well enough to interpret, but It seems like this task will always be called anytime a test is run.  Can we have it run only if `test/RapidWrightDCP` is empty?  That would be the clear situation when we know it will fail and that calling this task would allow the tests to at least have a chance of passing (such as a clean checkout).  If there are files of any type/revision/kind already in the `test/RapidWrightDCP` directory, this task should probably be skipped.  Does that make sense?

eddieh-xlnx: Yes, that would be my understanding too. The worry I have is that the `test/RapidWrightDCP` submodule directory would be empty just once, prior to any initialization. Once the developer happens to do a `git pull` to move to update the version of RW, then there's a danger that `RapidWrightDCP` may have moved on without which their tests would also fail. In this case, the burden falls onto the developer to do a `git submodule update` manually (dropping only the burden of `--init` necessary on the first run). I felt that the always-doing-it was the lesser of the two evils. If you still feel strongly about this, then I can research more about better ways to detect this situation.

clavin-xlnx: Good point, perhaps we'll need to think harder on this one.  Let me pop up to my top level concern:

_Users run `./gradlew test` or equivalent without knowing about the `RapidWrightDCP` submodule dependency.  This would lead to many test failures and a confused user.  I would like to avoid this scenario._  

Possible solutions to avoid this scenario:
1) Educate the user before they run `gradle`.  This could be augmenting the documentation, which already describes this in detail.  However, I don't think this is sufficient to cover all cases as not everyone reads documentation thoroughly.
2) Detect if `test/RapidWrightDCP` is empty when `./gradlew test` or equivalent is run and print a message telling the user what they need to do to successfully run tests in RapidWright and how they will need to keep `test/RapidWrightDCP` current.  If the user attempts to run `./gradlew test` with an empty `test/RapidWrightDCP`, they either don't know about the dependency or forgot.  In both cases, the message immediately informs them of a highly probable reason a bunch of tests just failed.
3) Some `gradle` automation that updates `test/RapidWrightDCP` correctly based on its current state.  I'm less sure I know of a good answer for this one.  If I am writing a test that needs a modified DCP that I haven't committed yet, how would `gradle` handle the update process?  Would it overwrite my modified DCP?  Ideally, `gradle` could be informed by `git` to only update files that are not being modified, but again, my understanding is fuzzy, so maybe this is not an easy thing to automate.

If we can pull off 3, that would be great, but I think 2 might be the easiest to implement.

eddieh-xlnx: Ah! I see where you're coming from. Let me think about 2. In the meantime:

> Some gradle automation that updates test/RapidWrightDCP correctly based on its current state. I'm less sure I know of a good answer for this one. If I am writing a test that needs a modified DCP that I haven't committed yet, how would gradle handle the update process? Would it overwrite my modified DCP? 

My understanding and some casual testing indicates that if the submodule is already at the current reference, new and modified files (DCPs) are both unaffected as it has nowhere else to go ("I'm already at the reference you're telling me to go to!"). 

However, if the submodule is at a different reference, then yes, `git submodule --update` will (a) move it back to the checked-in reference, and (b) leave new unversioned files untouched, as well as (c) carry over any modifications to versioned files post-move to the checked-in reference. By my far-from-complete understanding, the combination of (a) and (c) should mean that no work ever gets lost -- (a) moves from one committed point to another committed point, whereas (c) carries over any uncommitted changes (and I'm making the (untested) assumption that if it can't carry over such changes, it will error out just like how you can't `git checkout` if there's any conflicting changes).

Maybe we can do 2, and make 3 optional, maybe even display a reminder for 3 when tests fail?

clavin-xlnx: Yes, I think just an informative message providing "common solutions to failed tests" if any tests failed would probably be the safest.  Although we'd like to avoid users getting failed tests on the first go, it seems risky to try to automate-away that possibility.

eddieh-xlnx: Okay, I've changed the `git submodule update` to `git submodule init` which appears to be a NOP if the submodule is already initialized (even if it is pointing at something at the wrong commit). I'm going to kick the if-tests-fail-remind-about-submodules message to #290 since that cleans up the gradle file.

eddieh-xlnx: Oops. Obviously, `git submodule init` wasn't enough! Added some more magic to do a `git submodule update --init` if `test/RapidWrightDCP/.git` doesn't exist now.

clavin-xlnx (approved):

PR #290, opened 2021-11-10
Improve gradle test* tasks

eddieh-xlnx:
Previously: `test` task represents the built-in Java tests, and depends on `testPython`.
Now:
* `test` split into `testJava` (along with pre-existing `testPython`) and depends on them both; allows just Java tests to be run
* `test` task no longer supporting `--tests` argument for filtering -- error message now emitted asking users to apply filter onto `testJava` or `testPython` specifically
* `testPython` now supports `--tests` filtering, but only at the module granularity, e.g. `./gradlew testPython --tests com.xilinx.rapidwright.design` is about as fine as you can go for now. Filters with no hits will cause `pytest` to error out, similar to Java tests behaviour.
(I would expect it is possible to make it finer granularity, leaving it for the future)
* On test failure, the following reminder is emitted as a warning message:
`Failed tests detected. Consider checking submodule status and updating with './gradlew updateSubmodules' where necessary.`


build.gradle
```diff
@@ -44,40 +39,82 @@ dependencies {
...
+    proc.consumeProcessOutput(stdout,stderr)
+    proc.waitForOrKill(1100000)
+    println "$stdout\n$stderr"
+    f.delete()
+  }
+}
+
+task update_jars {
+  group = "build setup"
+  description = "Alias for upateJars."
```
eddieh-xlnx, line 63:
```suggestion
  description = "Alias for 'updateJars'."
```

build.gradle
```diff
@@ -82,20 +122,18 @@ task initSubmodules {
...
-task testPython(type:Exec) {
-  testPython() {
-      group = "verification"
-      description = "Runs the Python unit tests."
+task remindSubmodules {
+  onlyIf {
+    testJava.state.failure != null || testPython.state.failure != null
+  }
+  doLast {
+    logger.warn('Failed tests detected. Consider checking submodule status and updating with \'./gradlew updateSubmodules\' where necessary.')
```
clavin-xlnx, line 131:
```suggestion
    logger.warn('Failed tests detected. Some tests depend on DCPs from the git submodule in test/RapidWrightDCP, consider checking status and updating with 'gradlew updateSubmodules'')
```

clavin-xlnx: The `./` doesn't work in Windows, so perhaps we remove it?  I also have been using it, but it appears that after a `git clone`, the `gradlew` script has the executable attribute in Linux.  So, perhaps its not needed in Linux either.

build.gradle
```diff
@@ -44,29 +42,71 @@ dependencies {
...
-// Create a task using the task type
-tasks.register('update_jars', UpdateJarsTask)
+test {
+  dependsOn testJava
+  dependsOn testPython
+}
+
+gradle.taskGraph.whenReady {
+  if (!project.test.filter.commandLineIncludePatterns.isEmpty()) {
+    throw new InvalidUserDataException("'test' task cannot contain filters; apply filters directly to 'testJava'/'testPython' tasks instead.")
```
clavin-xlnx, line 107:
```suggestion
    throw new InvalidUserDataException("'test' task does not support filters (i.e. '--tests' option); please apply filters directly to 'testJava'/'testPython' tasks instead.")
```

clavin-xlnx (approved):

PR #294, opened 2021-11-18
RelocationTools to relocate more Site types

eddieh-xlnx:
Result of this discussion https://github.com/Xilinx/RapidWright/discussions/292#discussioncomment-1661671 as discovered by @bn999.

### DIFF
--- a/src/com/xilinx/rapidwright/design/tools/RelocationTools.java
+++ b/src/com/xilinx/rapidwright/design/tools/RelocationTools.java
@@ -40,6 +40,7 @@ import com.xilinx.rapidwright.design.SitePinInst;
 import com.xilinx.rapidwright.design.blocks.PBlock;
 import com.xilinx.rapidwright.device.PIP;
 import com.xilinx.rapidwright.device.Site;
+import com.xilinx.rapidwright.device.SiteTypeEnum;
 import com.xilinx.rapidwright.device.Tile;
 import com.xilinx.rapidwright.edif.EDIFHierCellInst;
 import com.xilinx.rapidwright.edif.EDIFNetlist;
@@ -53,6 +54,12 @@ import com.xilinx.rapidwright.util.Utils;
  *
  */
 public class RelocationTools {
+
+    public static final Set<SiteTypeEnum> supportedTypes;
+    static {
+        supportedTypes = Utils.sliceDspBramUramTypes;
+        supportedTypes.add(SiteTypeEnum.BUFGCE);
+    }
 
     /**
      * Relocate all SiteInsts (and all associated PIPs) belonging to the logical Cell at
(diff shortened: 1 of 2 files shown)


eddieh-xlnx:
@clavin-xlnx Even though I think this is sufficient for @bn999, should we add even more types?

clavin-xlnx:
> @clavin-xlnx Even though I think this is sufficient for @bn999, should we add even more types?

I think those changes make sense.  However, from the issue raised in #292, I wonder if we should have a wholesale mode.  That is, an option that relocates all types, which is essentially what @bn999 is trying to do.

eddieh-xlnx:
> I think those changes make sense. However, from the issue raised in #292, I wonder if we should have a wholesale mode. That is, an option that relocates all types, which is essentially what @bn999 is trying to do.

I was hesitant to enable blanket relocation of all types, as some things don't make sense, like I/O and perhaps Laguna sites, and even things like BUFG when we are not making SLR magnitude moves. Failed relocation means that nothing is relocated, whereas right now it just emits a warning.

eddieh-xlnx:
Looks like moving BUFGCEs now causes this error for `com.xilinx.rapidwright.design.TestRelocationTools.testMicroBlazeAndILA`:
```
    ERROR: Failed to move SiteInst 'BUFGCE_X1Y50' from Tile 'XIPHY_L_X32Y120' to Tile 'XIPHY_L_X32Y125'
    ERROR: Failed to move SiteInst 'BUFGCE_X1Y49' from Tile 'XIPHY_L_X32Y120' to Tile 'XIPHY_L_X32Y125'
    ERROR: Failed to move SiteInst 'BUFGCE_X0Y47' from Tile 'XIPHY_L_X0Y60' to Tile 'XIPHY_L_X0Y65'
    ERROR: Failed to move SiteInst 'BUFGCE_X0Y24' from Tile 'XIPHY_L_X0Y60' to Tile 'XIPHY_L_X0Y65'
```

eddieh-xlnx:
TODO: Add a `Set<SiteTypeEnum>` overload to `relocate` functions.

eddieh-xlnx:
Okay, I've added a `siteTypes` argument to the following method:
```
public static boolean relocate(Design design,
                                   String instanceName,
                                   int tileColOffset,
                                   int tileRowOffset,
                                   Set<SiteTypeEnum> siteTypes)
```
allowing you specify your own siteTypes, with a default value of what we have today.

Alternatively for @bn999's original problem Instead of using the `instanceName` overload, you can also use the `PBlock` overload:
```
public static boolean relocate(Design design,
                                   PBlock pblock,
                                   int tileColOffset,
                                   int tileRowOffset)
```
which will attempt to move all site instances inside that PBlock (which can be an entire SLR, e.g. `SLR0:SLR0`) irrespective of type.

src/com/xilinx/rapidwright/design/tools/RelocationTools.java
```diff
@@ -66,12 +79,16 @@
      *                     (empty for top cell)
      * @param tileColOffset Relocate this number of tile columns (X axis)
      * @param tileRowOffset Relocate this number of tile rows (Y axis)
+     * @param siteTypes     Set of SiteTypeEnum-s to relocate
+     *                      (overload exists where this is emitted thus
```
clavin-xlnx, line 83:
```suggestion
     *                      (overload exists where this is omitted thus
```

clavin-xlnx (approved):

PR #296, opened 2021-11-23
Corrects some Javadoc and logging info

YunxZhou:


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -289,18 +289,17 @@ private void determineRoutingTargets(){
 				System.err.println("ERROR: Unknown net " + net.toString());
 			}
 		}
-		if(this.config.isPrintConnectionSpan()) this.printConnectionSpanStatistics();
+		this.printConnectionSpanStatistics();
```
clavin-xlnx, line 292: Does this mean that `printConnectionSpanStatistics()` will now print with every run of RWRoute?  Is that what we want?

YunxZhou: Not really. I added the verbose control to printConnectionSpanStatistics() in a new commit. Now the verbose controls the print and config.isPrintConnectionSpan() determines if the detailed span count would be printed or not. Does it makes sense to you?

clavin-xlnx (approved):

PR #317, opened 2021-12-17
Test routethru PIPs do not appear as sink pins

eddieh-xlnx:
It appears that on Series 7 (but not UltraScale+) Vivado chooses to emit a route that has COUT as the driver, and then uses a COUT -> DMUX route thru PIP in order to exit onto the general fabric.

In this case, RapidWright incorrectly recognizes the DMUX as a sink pin. Test for this here, fix expected to appear in an upcoming release.

### DIFF
--- a/test/com/xilinx/rapidwright/design/TestDesign.java
+++ b/test/com/xilinx/rapidwright/design/TestDesign.java
@@ -9,6 +9,7 @@ import com.xilinx.rapidwright.device.BEL;
 import com.xilinx.rapidwright.device.Device;
 import com.xilinx.rapidwright.device.Site;
 import com.xilinx.rapidwright.edif.EDIFNetlist;
+import com.xilinx.rapidwright.support.RapidWrightDCP;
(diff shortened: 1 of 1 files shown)


test/com/xilinx/rapidwright/design/TestDesign.java
```diff
@@ -87,4 +88,16 @@ public void checkDcpRoundtripModuleInstAnchor(@TempDir Path tempDir) throws IOEx
         mi = design.getModuleInst("inst");
         Assertions.assertEquals(oldAnchor, mi.getAnchor().toString());
     }
+
+    @Test
+    @CheckOpenFiles
+    public void testRouteThruPIPsSitePinInsts() {
```
eddieh-xlnx, line 94:
```suggestion
    public void testDcpRouteThruPIPsNotSourcePins() {
```

eddieh-xlnx:
Merged as part of #338.

PR #318, opened 2021-12-20
Test secondary source pins created correctly

eddieh-xlnx:
On Series 7, secondary site types (e.g. RAMB18) can have differently named pins than primary site types (e.g. FIFO18) and thus `Design.readCheckpoint` should be mindful of that. RapidWright currently handles this correctly for sink pins, but not for source pins.

This issue (at least for RAM/FIFO sites) does not appear to be an issue on UltraScale+.

Test for this here, fix expected to appear in an upcoming release.


eddieh-xlnx:
Merged as part of #338.

PR #319, opened 2021-12-22
Fixes missing VCC pin on dual-use routethru LUT in DesignTools.copyIm…

clavin-xlnx:
…plementation()

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -2049,7 +2049,7 @@ public class DesignTools {
 									break;
 								}
 							}
-							if(rtCopy.getBELName().endsWith("6LUT")) {
+							if(rtCopy.getBELName().endsWith("6LUT") && isUltraScale(rtCopy)) {
 							    // Check A6 if it has VCC assignment
 							    BELPin a6 = rtCopy.getBEL().getPin("A6");
 							    Net isVcc = origSiteInst.getNetFromSiteWire(a6.getSiteWireName());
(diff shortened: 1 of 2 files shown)


clavin-xlnx:
A LUT6 and LUT5 pair can be used simultaneously, however, in certain conditions, the A6 pin must be tied to VCC if both are used.  For example, when the 6LUT is used as a route-thru and 5LUT is used as a static source, the A6 pin must be tied to VCC.  In `DesignTools.copyImplementation()` this VCC pin was not getting copied correctly and was causing a design in Vivado to need two calls to `route_design` before the design was fully routed.  This PR fixes this issue.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2048,7 +2048,16 @@ private static void copySiteRouting(Cell copy, Cell orig, Map<String,String> src
 									curr = tmp;
 									break;
 								}
-							}	
+							}
+							if(rtCopy.getBELName().endsWith("6LUT")) {
```
eddieh-xlnx, line 2052: Maybe looking too far ahead, but would this be necessary for Versal? AM005 states:
> Enables dual LUT functions of up to six inputs (five in prior architectures)

eddieh-xlnx (approved):

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx: Yes, this probably only applies to UltraScale{,+}, I've added a check.

PR #320, opened 2021-12-23
Specifies complete set of dependencies on Maven Central in build.gradle

clavin-xlnx:
As an alternative to downloading a set of dependency jars from GitHub, the `build.gradle` file has been augmented to specify the set of dependencies for RapidWright to download directly from Maven Central.  This will make RapidWright better fit into the Java package system and will ease install/setup for developers.  

A task called `copyJars` also copies the set of jar dependencies of the project into the conventional `RapidWright/jars` directory.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,10 +43,6 @@ jobs:
         run: |
           pip install pytest python/
 
-      - name: Download RapidWright Jars
-        if: steps.cache-rapidwright.outputs.cache-hit != 'true'
-        run: ./gradlew update_jars
-
       - name: Compile
         run: ./gradlew assemble
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,6 @@ jobs:
         with:
           path: |
               data
-              jars
           key: rapidwright-clouddata-v2-${{ env.RAPIDWRIGHT_VERSION }}
 
       - name: Install Python Dependencies
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
           path: |
               data
               jars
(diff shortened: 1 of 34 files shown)


build.gradle
```diff
@@ -85,7 +208,7 @@ task testPython(type:Test) {
     exec {
       environment 'CLASSPATH', sourceSets.main.runtimeClasspath.getAsPath()
       executable = 'python3'
-      args = ['-m', 'pytest', '--junitxml', "$buildDir/test-results/testPython/testPython.xml", '--rootdir', 'python/test']
+      args = ['-m', 'pytest', '-v', '--junitxml', "$buildDir/test-results/testPython/testPython.xml", '--rootdir', 'python/test']
```
clavin-xlnx, line 211: @eddieh-xlnx - I left in the `-v` by accident, but perhaps it would be useful to keep it in?

.github/workflows/build.yml
```diff
@@ -37,16 +37,12 @@ jobs:
           path: |
               data
               jars
```
eddieh-xlnx, line 39: Perhaps we should remove `jars` from being cached?

As I understand it, Gradle will run the `copyJars` task as part of `compileJava`.
And gradle is also cached by the `actions/setup-java` step.

clavin-xlnx: We could remove it, technically, Gradle will be looking elsewhere for the jars anyways.   For me, it puts the dependencies in a local home directory called `~/.gradle/caches`.  The `jars` directory will be for users outside of Gradle and for legacy purposes.

eddieh-xlnx: Sorry I wasn't clear. I meant to remove it from being cached redundantly for CI, since it should be re-populated by Gradle (from its cache or not) anyway. If we want to speed it up even more we could stop `copyJars` from being compulsory, for users (like CI) who run everything only through Gradle, but that can be a topic for another day.

eddieh-xlnx (approved):

PR #321, opened 2022-01-05
Customized RWRoute for RapidStream

YunxZhou:
This PR contains tested RWRoute with parameterized customization for the RapidStream flow.
It also includes changes that are related to removing static members of classes for the purpose of running RWRoute instances in parallel.


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -289,14 +299,79 @@ private void determineRoutingTargets(){
...
-		if(this.config.isVerbose()) this.printConnectionSpanStatistics();
+	}
+	
+	/**
+	 * Deals with the conflicted nets.
+	 * Note: this is customized for the RapidStream use case. Other users will need to adapt this method.
+	 * @param design The design to route.
+	 * @param conflictNets A set of conflicted nets.
+	 */
+	private void handleConflictNets(Design design, Set<Net> conflictNets) {
```
clavin-xlnx, line 310: I am confused by the arguments of this method.  In the constructor, the `design` to be routed is provided.  There is also a private field for `conflictNets`.  Why does the caller need to provide these again?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -289,14 +299,79 @@ private void determineRoutingTargets(){
...
+	}
+	
+	/**
+	 * Checks if a net is an anchor net that is from / to a CLB tile.
+	 * Note: this is customized for the RapidStream use case.
+	 * @param net The net in question.
+	 * @param design The design to route.
+	 * @return true if the net is an anchor net from / to a CLB tile.
+	 */
+	private boolean isRegularAnchorNet(Net net, Design design) {
```
clavin-xlnx, line 346: Again, is the `design` parameter necessary here?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -289,14 +299,79 @@ private void determineRoutingTargets(){
...
+			if(net.getSinkPins().size() > 1) {
+				// In the RapidStream flow, the target nets to route are 2-terminal FF-to-FF nets.
+				toPreserveNets.add(net);
+				continue;
+			}
+			
+			if(!this.isRegularAnchorNet(net, design)) {
+				toPreserveNets.add(net);
+				continue;
+			}
```
clavin-xlnx, line 328:
```suggestion
			if(net.getType() != NetType.WIRE || net.getSinkPins().size() > 1 || !this.isRegularAnchorNet(net, design)) {
				// Skip successfully routed CLK, VCC, and GND nets
				// In the RapidStream flow, the target nets to route are 2-terminal FF-to-FF nets.
				toPreserveNets.add(net);
				continue;
			}
```

The bodies of the three `if` statements seem redundant.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
clavin-xlnx: Also, rather than simply using the term "Anchor" net, perhaps we could define the kind of situation that it is detecting so that others could find it useful.  If we state that it is only used for RapidStream, it will not be as easily leveraged.  Can you define in the Javadoc what kind of net it is detecting?

eddieh-xlnx (commented):
Taking @clavin-xlnx's comments a little further, I think it would be more scalable if we separated the RapidStream-specific enhancements into a derived class. This would keep the RWRoute pure and easily maintainable/understandable, and serve as an example for future application-specific modifications. Thanks!

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -187,11 +188,7 @@ public RWRoute(Design design, RWRouteConfig config){
 		
 		RoutableNode.setMaskNodesCrossRCLK(this.config.isMaskNodesCrossRCLK());
 
-		if(this.config.isTimingDriven()) {
-			DSPTimingData.setDSPTimingFolder(config.getDspTimingDataFolder());
-			ClkRouteTiming clkTiming = createClkTimingData(config);
-			this.routesToSinkINTTiles = clkTiming == null? null : clkTiming.getRoutesToSinkINTTiles();
-			this.timingManager = new TimingManager(this.design, true, this.routerTimer, this.config, clkTiming);		
+		if(this.config.isTimingDriven()) {
```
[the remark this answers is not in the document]
clavin-xlnx: I think this issue got overlooked.  Please take a look at the link, I see lots of uses of `this` that are unnecessary.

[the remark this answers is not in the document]
YunxZhou: Thanks for the reminder, Chris. It is because using this allows intellisense to pop up in Eclipse that I used this a lot. I was unware that they are supposed to be heavy. 
Those uses of this have been removed in the updated files.

YunxZhou:
> Taking @clavin-xlnx's comments a little further, I think it would be more scalable if we separated the RapidStream-specific enhancements into a derived class. This would keep the RWRoute pure and easily maintainable/understandable, and serve as an example for future application-specific modifications. Thanks!

Thanks for the comments, Eddie. It is good to keep RWRoute pure. The RapidStream-specific methods (handleConflictNets and isTargetConflicNetToRoute) have been refined and moved to PartialRouter. Javadoc is also optimized. I suppose we are fine with that?

clavin-xlnx (approved):
Looks good to me.  @eddieh-xlnx - did you want to take a look before we merge?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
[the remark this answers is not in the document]
clavin-xlnx: Thanks @YunxZhou for cleaning up the `this` references.  In Eclipse, you can always bring up the Intellisense menu by using the keyboard shortcut `<Ctrl>+<Space>`.  I use it all the time and avoids having to type `this` all the time.

src/com/xilinx/rapidwright/rwroute/PartialRouter.java
```diff
@@ -38,18 +41,84 @@ public PartialRouter(Design design, RWRouteConfig config){
...
+	}
+	
+	/**
+	 * Checks if a net is the target conflict net to be routed. 
+	 * Note: this method provides an example of customizing the partial router for application-specific tool flows.
+	 * It is customized for the RapidStream use case, where the targets are nets connecting anchor FFs of CLB tiles.
+	 * @param net The net in question.
+	 * @return true if the net is a target net.
+	 */
+	private boolean isTargetConflicNetToRoute(Net net) {
```
eddieh-xlnx, line 82: Missing a `t` from Conflict.

eddieh-xlnx:
> The RapidStream-specific methods (handleConflictNets and isTargetConflicNetToRoute) have been refined and moved to PartialRouter.

I'm not sure I agree with moving it into PartialRouter either? In my mind, the partial router should also stay pure, and I was thinking of a RapidStream class, inheriting from PartialRouter, which contains all the application-specific code along with a clean interface such that a different application could override this behaviour in a different manner.

eddieh-xlnx (changes requested):
This is what I meant: https://github.com/eddieh-xlnx/RapidWright/commit/716e796bb799d6dd3f327e3f382b0e2a3c6e50f3

Good software engineering practice is to avoid polluting the base class as much as possible, otherwise once we support more specialist applications every new application will need to add their own code. I think what I suggest accomplishes that. Feel free to cherry-pick if you agree.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -1757,20 +1748,30 @@ public static Design routeDesignWithUserDefinedArguments(Design design, String[]
 	 * @param config A {@link RWRouteConfig} instance consisting of customizable parameters to use.
 	 */
 	private static Design routeDesign(Design design, RWRouteConfig config) {
+		return routeDesign(design, config, () -> {
+			if(config.isPartialRouting()) {
+				return config.isResolveConflictNets()? new RapidStreamRoute(design, config) : new PartialRouter(design, config);
```
eddieh-xlnx, line 1753: This is still wrong, should be: https://github.com/eddieh-xlnx/RapidWright/blob/716e796bb799d6dd3f327e3f382b0e2a3c6e50f3/src/com/xilinx/rapidwright/rwroute/RWRoute.java#L1753

As a suggestion for the future collaboration, please cherry-pick contributions from others as that can help prevent mistakes and preserves attribution.

YunxZhou: I made that change on purpose, as the suggested seems to have a problem in instantiating the correct router object for the specific use case.

YunxZhou: Thanks for the further comment making it clear. This is updated in the new commit.

eddieh-xlnx (approved):

PR #323, opened 2022-01-11
Smoke test for loading DCPs in parallel

clavin-xlnx:

### DIFF
--- a/test/com/xilinx/rapidwright/design/TestParallelDCPLoad.java
+++ b/test/com/xilinx/rapidwright/design/TestParallelDCPLoad.java
@@ -1,0 +1,34 @@
+package com.xilinx.rapidwright.design;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import com.xilinx.rapidwright.support.CheckOpenFiles;
+import com.xilinx.rapidwright.support.RapidWrightDCP;
+
+public class TestParallelDCPLoad {
+
+    @Test
+    @CheckOpenFiles
+    public void testParallelDCPLoad(@TempDir Path tempDir) throws IOException {
+        List<String> dcpList = new ArrayList<>();
+        dcpList.add(RapidWrightDCP.getString("picoblaze_ooc_X10Y235.dcp"));
+        dcpList.add(RapidWrightDCP.getString("picoblaze4_ooc_X6Y60_X6Y65_X10Y60_X10Y65.dcp"));
+        dcpList.add(RapidWrightDCP.getString("optical-flow.dcp"));
+        dcpList.add(RapidWrightDCP.getString("bnn.dcp"));
+        
+        List<Design> designs = dcpList.stream().parallel() // Comment out the .parallel() part to run serially
+            .map(p -> (Design.readCheckpoint(p))).collect(Collectors.toList());
+        
+        // Write out serially for now as a sanity check
+        for(Design dcp : designs) {
+            dcp.writeCheckpoint(tempDir.resolve(dcp.getName() + ".dcp"));
+        }
+    }
+}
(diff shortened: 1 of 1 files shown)


test/com/xilinx/rapidwright/design/TestParallelDCPLoad.java
```diff
@@ -0,0 +1,34 @@
...
+        List<String> dcpList = new ArrayList<>();
+        dcpList.add(RapidWrightDCP.getString("picoblaze_ooc_X10Y235.dcp"));
+        dcpList.add(RapidWrightDCP.getString("picoblaze4_ooc_X6Y60_X6Y65_X10Y60_X10Y65.dcp"));
+        dcpList.add(RapidWrightDCP.getString("optical-flow.dcp"));
+        dcpList.add(RapidWrightDCP.getString("bnn.dcp"));
+        
+        List<Design> designs = dcpList.stream().parallel() // Comment out the .parallel() part to run serially
+            .map(p -> (Design.readCheckpoint(p))).collect(Collectors.toList());
+        
+        // Write out serially for now as a sanity check
```
eddieh-xlnx, line 29: May I suggest removing the write? It would save I/O and runtime, and it's not like we're reading it back in.

An alternative test would be to load the designs serially and in parallel, and then comparing some of their properties (e.g. that the SiteInst and Nets are identical?)

test/com/xilinx/rapidwright/design/TestParallelDCPLoad.java
```diff
@@ -0,0 +1,34 @@
...
+    @Test
+    @CheckOpenFiles
+    public void testParallelDCPLoad(@TempDir Path tempDir) throws IOException {
+        List<String> dcpList = new ArrayList<>();
+        dcpList.add(RapidWrightDCP.getString("picoblaze_ooc_X10Y235.dcp"));
+        dcpList.add(RapidWrightDCP.getString("picoblaze4_ooc_X6Y60_X6Y65_X10Y60_X10Y65.dcp"));
+        dcpList.add(RapidWrightDCP.getString("optical-flow.dcp"));
+        dcpList.add(RapidWrightDCP.getString("bnn.dcp"));
+        
+        List<Design> designs = dcpList.stream().parallel() // Comment out the .parallel() part to run serially
```
eddieh-xlnx, line 26: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources currently states that a Windows/Linux runner has 2 CPU cores.

I believe that `parallel()` streams uses the `ForkJoinPool.commonPool()` which has NumCores - 1 threads, thus both cores (the main thread and that one common pool thread) should be active. To really stress all four DCPs being read in parallel (even inefficiently), perhaps you may want to create your own four thread pool and let the OS interleave them?

eddieh-xlnx (commented):
Since there hasn't been a new release yet, should we expect this test to fail in its current state?

clavin-xlnx:
> Since there hasn't been a new release yet, should we expect this test to fail in its current state?

It might but its not guaranteed to fail, it obviously didn't fail in this first run.

test/com/xilinx/rapidwright/design/TestParallelDCPLoad.java
clavin-xlnx: Perhaps we should try both?  Keep the existing test and add the 4-thread pool as a separate test?

clavin-xlnx: Any pointers on how to setup a 4-thread pool?

eddieh-xlnx: > Perhaps we should try both? Keep the existing test and add the 4-thread pool as a separate test?

I'm not sure I see the value of keeping it back but won't object if you feel more strongly.

> Any pointers on how to setup a 4-thread pool?

I would suggest:
```
ExecutorService exec = Executors.newFixedThreadPool(dcpList.size());
exec.invokeAll(dcpList.stream()
            .map(p -> (Design.readCheckpoint(p))).collect(Collectors.toList()));
exec.shutdown();
```

eddieh-xlnx:
> It might but its not guaranteed to fail, it obviously didn't fail in this first run.

I'd like to see it fail at least once to make sure we're really exercising the right thing.

clavin-xlnx:
Added another test, but it seems like getting a parallel bug to fail reliably will be difficult to control.  Closing for now.

PR #330, opened 2022-01-25
Refactor gradle to facilitate other gradle parents

eddieh-xlnx:
Mainly, moving the dependency and testing stuff into `common.gradle` which is included back.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,20 +17,6 @@ def get_rapidwright_version_line = file('.github/workflows/build.yml').filterLin
 def rapidwright_version = sw.toString().replace("RAPIDWRIGHT_VERSION: v", "").replace("-beta", "").trim()
 
 dependencies {
-    api 'com.esotericsoftware:kryo:5.2.1'
-    api 'org.jgrapht:jgrapht-core:1.3.0'
-    api 'org.capnproto:runtime:0.1.13'
-    api 'net.sf.jopt-simple:jopt-simple:5.0.4'
-    api 'org.python:jython-standalone:2.7.2'
-    api 'com.google.protobuf:protobuf-java:3.11.4'
-    api 'org.jetbrains:annotations:20.1.0'
-    api 'org.zeromq:jeromq:0.5.2'
-    api 'commons-cli:commons-cli:1.2'
-    api 'org.json:json:20160810'
-    api 'com.jcraft:jzlib:1.1.3'
-    api 'commons-io:commons-io:2.11.0'
-    api 'com.xilinx.rapidwright:qtjambi-'+os+':4.5.2_01'
-    api 'com.xilinx.rapidwright:jupyter-kernel-jsr223:1.0.0'
     api 'com.xilinx.rapidwright:rapidwright-api-lib:' + rapidwright_version
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -1,28 +1,12 @@
-import java.util.stream.Collectors
-
 plugins {
     id 'java'
     id 'maven-publish'
-    id 'application'
-    id 'java-library'
     id 'signing'
 }
 
+apply from: 'common.gradle'
 
 mainClassName = "com.xilinx.rapidwright.MainEntrypoint"
-
-def os = System.getProperty("os.name").toLowerCase().contains("windows") ?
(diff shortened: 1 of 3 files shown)


common.gradle
```diff
@@ -0,0 +1,79 @@
+apply plugin: 'application'
+apply plugin: 'java-library'
+
+sourceCompatibility = 8
+targetCompatibility = 8
+
+repositories {
+    mavenCentral()
+}
```
clavin-xlnx, line 9:
```suggestion
repositories {
    if (project.hasProperty('useMavenLocal') && useMavenLocal == "True") {
       project.logger.lifecycle('Using Maven Local Repository')
       mavenLocal()
    }
    mavenCentral()
}
```
This allows the property `useMavenLocal` to be set to `True` if a user wants to run from a local Maven repository.  This can be set in the `gradle.properties` file (generally located in `~/.gradle` in Linux).

build.gradle
```diff
@@ -1,28 +1,28 @@
...
 plugins {
     id 'java'
     id 'maven-publish'
-    id 'application'
-    id 'java-library'
     id 'signing'
 }
 
+if (System.getenv('RAPIDWRIGHT_PATH') == null) {
+  throw new GradleException('RAPIDWRIGHT_PATH environment variable not set')
```
clavin-xlnx, line 8: Although in the past we had recommended users set `RAPIDWRIGHT_PATH`, I've actually tried to remove the dependence of it.  We've had more trouble with users setting it and running into issues than it has solved.  Is there a way we can mimic what https://github.com/Xilinx/RapidWright/blob/567aeaf926381506e055aa9d37e154bed17a5304/src/com/xilinx/rapidwright/util/FileTools.java#L866 does when the variable is not set?

common.gradle
```diff
@@ -0,0 +1,107 @@
+apply plugin: 'application'
+apply plugin: 'java-library'
+
+sourceCompatibility = 8
+targetCompatibility = 8
+
+def rapidwrightDir = System.getenv("RAPIDWRIGHT_PATH")
```
clavin-xlnx, line 7:
```suggestion
def rapidwrightDir = System.getenv('RAPIDWRIGHT_PATH') == null ? buildscript.sourceFile.getParent() : System.getenv("RAPIDWRIGHT_PATH")
```
Perhaps something like this might work?

eddieh-xlnx: Good idea. I've now removed the requirement for `RAPIDWRIGHT_PATH` here, but it is required for other parent gradles that source `common.gradle`.

common.gradle
```diff
@@ -0,0 +1,107 @@
...
+    }
+    mavenCentral()
+}
+
+def os = System.getProperty("os.name").toLowerCase().contains("windows") ?
+                                       "win64-msvc2005x64" : "linux64-gcc"
+def sw = new StringWriter()
+def get_rapidwright_version_line = new File(rapidwrightDir + '/.github/workflows/build.yml').filterLine(sw) { 
+    line -> line.contains('RAPIDWRIGHT_VERSION:') 
+}
```
clavin-xlnx, line 22: Just curious, this code is duplicated from `build.gradle` however, in the other file, the file constructor changed from `new File()` to `file()`, but that same change was not propagated here.  Not sure the difference between the two.

eddieh-xlnx: I cleaned up this duplication just now. My understanding is that `file()` is a Gradle provided function which is relative to the project directory, whereas `File` is I think relative to the current working directory. The former has useful when doing included builds.

common.gradle
```diff
@@ -0,0 +1,106 @@
+apply plugin: 'application'
+apply plugin: 'java-library'
+
+sourceCompatibility = 8
+targetCompatibility = 8
+
+def rapidwrightDir = System.getenv("RAPIDWRIGHT_PATH") ? System.getenv("RAPIDWRIGHT_PATH") : file('.').toString()
```
clavin-xlnx, line 7:
```suggestion
def rapidwrightDir = System.getenv("RAPIDWRIGHT_PATH") ?: file('.').toString()
```
Just learned that Groovy has a more liberal definition of `True` than Java.  This led me to discover the ["Elvis operator"](https://groovy-lang.org/operators.html#_elvis_operator).

PR #333, opened 2022-01-28
Test for checking that modules are legally relocated wrt SLR boundaries

clavin-xlnx:


eddieh-xlnx:
Merged as part of #338.

PR #335, opened 2022-01-28
To copy static net

pongstorn-xilinx:
Please pay special attention to some enums removed from Utils.java


src/com/xilinx/rapidwright/util/Utils.java
```diff
@@ -315,9 +315,7 @@ public static boolean isURAM(SiteInst s){
 		
 		lockedSiteTypes = EnumSet.of(
 			SiteTypeEnum.CONFIG_SITE,
-			SiteTypeEnum.BUFG,
-			SiteTypeEnum.BUFGCE,
-			SiteTypeEnum.BUFGCTRL
+			SiteTypeEnum.BUFG
```
clavin-xlnx, line 318: I think this should be ok.

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -58,6 +61,15 @@
...
 	/** A list of all nets internal to this module instance */
 	private ArrayList<Net> nets;
+
+	/** A list of all SiteInsts that are the sources of static nets */
+	private List<SiteInst> staticSources;
+	
+	private boolean hasBeenPlaced = false;
+
+	/** A map from a static net to the list of its PIPs */
+	private Map<NetType, List<Pair<PIP, Tile>>> staticPIPs;
```
clavin-xlnx, line 71: This data structure seems like it could be more efficient, but I'm not coming up with anything obvious.  The `List` could be replaced with an array, but that might only provide minimal savings and complicate things a bit.  @eddieh-xlnx, thoughts?

clavin-xlnx (approved):

eddieh-xlnx:
Would it be possible to get a standalone test for this new functionality, which I can also use to see how it is used?

src/com/xilinx/rapidwright/design/ModuleInst.java
eddieh-xlnx: One option would be to merge static nets in with the rest of the `nets` array?

I wonder if we can get away with storing `List<PIP>` as the key instead of the pair (or merge with `Net` entirely). I think you're using the `Tile` to determine the original template tile from which you compute the corresponding tile on placement. Could it better to move towards the same system that non-static nets use, which is to get the template *static* net on the `ModuleInst` which will have the template tile, from which you can then compute the corresponding one?

eddieh-xlnx: If you take the above approach, then you wouldn't need your own `addStaticPIPs` at all since it could be pulled from the template Module.

`addStaticSource` may need some reworking though if you really wanted to get rid of it in the same manner.

Perhaps `hasBeenPlaced` can be inferred from whether `instances` has been placed or not?
That brings me to something I think you've missed: the `unplace()` method does not reset this boolean nor does it undo the addition of static PIPs to the global nets?

clavin-xlnx:
Closing since this has been superseded by #357.

PR #337, opened 2022-01-31
Supporting changes for parallel Design.writeCheckpoint()

eddieh-xlnx:


src/com/xilinx/rapidwright/util/ParallelismTools.java
```diff
@@ -0,0 +1,353 @@
...
+
+/**
+ * Utilities to aid in parallel processing
+ *
+ * A class that abstracts away single-threaded and multi-threaded execution.
+ * Single-threaded mode means that all tasks submitted will be executed
+ * immediately (on the submitting thread).
+ */
+public class ParallelismTools {
+    private static boolean parallel = false;
```
clavin-xlnx, line 52: Can we make parallel DCP writes on by default?

test/com/xilinx/rapidwright/design/TestDesign.java
```diff
@@ -87,4 +89,26 @@ public void checkDcpRoundtripModuleInstAnchor(@TempDir Path tempDir) throws IOEx
...
+            before.writeCheckpoint(filenameWrite);
+        } finally {
+            ParallelismTools.setParallel(false);
+        }
+
+        Design after = Design.readCheckpoint(filenameWrite);
+
+        Assertions.assertEquals(before.getNetlist(), after.getNetlist());
+        Assertions.assertEquals(before.getSiteInstMap(), after.getSiteInstMap());
+        Assertions.assertEquals(before.getNetMap(), after.getNetMap());
```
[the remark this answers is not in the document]
eddieh-xlnx: Let me try this. I seem to recall this was problematic because it would perform an ordered compare of the two collections, whereas comparing their maps would one-liner work.

src/com/xilinx/rapidwright/util/ParallelismTools.java
eddieh-xlnx: I would suggest we turn it on at the next release once it's had some time to bed down and get tested externally.

clavin-xlnx: I think the next release sounds reasonable.  I also wonder if we should provide a separate mechanism (environment variable, etc) that would allow users to change this setting without having to write code/recompile, etc.

eddieh-xlnx:
Merged as part of #338.

PR #339, opened 2022-02-10
Memory optimization for read EDIF

clavin-xlnx:
This PR contains three optimization aimed at reducing the memory footprint of an `EDIFNetlist`, primarily large ones.  The three optimizations are:

1) In `EDIFNet`, `Map<String,EDIFPortInst> portInsts` has been converted to `List<EDIFPortInst> portInsts`.  This is because 60% of `EDIFNetlist` memory consumption is from `HashMap` instances in `EDIFNet` and `EDIFCellInst`, with 67% of those instances coming from `EDIFNet`.  
2) In `EDIFCellInst`, `Map<String,EDIFPortInst> portInsts` has been converted to `List<EDIFPortInst> portInsts`.  This is because 60% of `EDIFNetlist` memory consumption is from `HashMap` instances in `EDIFNet` and `EDIFCellInst`, with 30% of those instances coming from `EDIFCellInst`.  
3) In `EDIFName` the default map type is being changed from `LinkedHashMap` to `HashMap`.  The previous was being used to preserve insertion order, but this is not strictly necessary.

Overall, these three optimizations **reduces memory footprint ~70%**, with the latter one contributing ~10% of the reduction.  Unfortunately, this does come at a cost to a few APIs that will exhibit behavior of `O(n)` instead of `O(1)`.  One way to mitigate this is to sort the lists in `EDIFNet` and `EDIFCellInst` which can reduce the impact somewhat to `O(log n)`.  In the end, a bit more testing and work is needed to identify the true impact of these changes and create alternative use modes to avoid the impacted APIs.  The main impacted APIs are:

- `EDIFNet.getPortInst(String fullName)`
- `EDIFNet.removePortInst(EDIFPortInst portInst)`
- `EDIFNet.removePortInst(String portInstFullName)`
- `EDIFCellInst.getPortInst(String name)`
- `EDIFCellInst.removePortInst(EDIFPortInst epr)`

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFCellInst.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFCellInst.java
@@ -27,11 +27,9 @@ package com.xilinx.rapidwright.edif;
 
 import java.io.IOException;
 import java.io.Writer;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 
 /**
@@ -53,7 +51,7 @@ public class EDIFCellInst extends EDIFPropertyObject implements EDIFEnumerable {
 	public static final String BLACK_BOX_PROP = "IS_IMPORTED";
 	public static final String BLACK_BOX_PROP_VERSAL = "black_box";
 	
-	private List<EDIFPortInst> portInsts;
+	private EDIFPortInstList portInsts;
 
     protected EDIFCellInst(){
         
@@ -92,13 +90,27 @@ public class EDIFCellInst extends EDIFPropertyObject implements EDIFEnumerable {
     }
     
     /**
-     * Helper method to help maintain port ref map.  Adds a new
-     * port ref for this instance.
-     * @param epr The port ref to add
-     * @return Any previous port ref of the same name, null if none already exists.
+     * Creates a new map of all the EDIFPortInst objects stored on this EDIFCellInst.  The new map
+     * contains a copy of EDIFPortInsts available at the time of invocation as returned from 
+     * {@link #getPortInstList()}.      
+     * @return A map of EDIFPortInst names ({@link EDIFPortInst#getName()} to the corresponding objects.
+     */
+    public Map<String, EDIFPortInst> getPortInstMap(){
+        if(portInsts == null) return Collections.emptyMap();
+        HashMap<String, EDIFPortInst> map = new HashMap<>();
+        for(EDIFPortInst e : getPortInstList()) {
+            map.put(e.getName(), e);
+        }
+        return map;
+    }
+    
+    /**
+     * Adds a new EDIFPortInst to this cell instance. The port instances
+     * are stored in a sorted ArrayList, so worst case is O(n).
+     * @param epr The port instance to add
      */
     protected void addPortInst(EDIFPortInst epr) {
-        if(portInsts == null) portInsts = new ArrayList<>();
+        if(portInsts == null) portInsts = new EDIFPortInstList();
         if(!epr.getCellInst().equals(this)) 
             throw new RuntimeException("ERROR: Incorrect EDIFPortInst '"+
                 epr.getFullName()+"' being added to EDIFCellInst " + toString());
@@ -106,35 +118,36 @@ public class EDIFCellInst extends EDIFPropertyObject implements EDIFEnumerable {
     }
     
     /**
-     * Removes the provided port ref, if it exists.
-     * @param epr The port ref to remove.
-     * @return The removed port ref, or null if none exists.
+     * Removes the provided port instance from the cell instance, if it exists.  The port instances
+     * are stored in a sorted ArrayList, so worst case is O(n).
+     * @param epr The port instance object to remove
+     * @return The removed port instance, or null if it was not found.
      */
     protected EDIFPortInst removePortInst(EDIFPortInst epr){
         if(portInsts == null) return null;
-        for(int i=0; i < portInsts.size(); i++) {
-            if(portInsts.get(i).equals(epr)) {
-                return portInsts.remove(i);
-            }
-        }
-        return null;
+        return portInsts.remove(epr);
     }
     
     /**
-     * Gets the port ref on this cell by pin name (not full
-     * port ref name).  
-     * @param name Name of the pin in the port ref to get. 
-     * @return A port ref by pin name.
+     * Removes the named port instance from the cell instance, if it exists. The port instances
+     * are stored in a sorted ArrayList, so worst case is O(n).
+     * @param portName Name of the port ref to remove ({@link EDIFPortInst#getName()})
+     * @return The removed port instance, or null if none found by that name.
+     */
+    protected EDIFPortInst removePortInst(String portName){
+        if(portInsts == null) return null;
+        return portInsts.remove(this, portName);
+    }
+    
+    /**
+     * Gets the port instance on this cell by pin name ({@link EDIFPortInst#getName()}). The port 
+     * instances are stored in a sorted ArrayList, so worst case is O(log n).
+     * @param name Name of the port instance to get. 
+     * @return The named port instance, or null if none found by that name. 
      */
     public EDIFPortInst getPortInst(String name){
         if(portInsts == null) return null;
-        for(int i=0; i < portInsts.size(); i++) {
-            EDIFPortInst pi = portInsts.get(i);
-            if(pi.getName().equals(name)) {
-                return pi;
-            }
-        }
-        return null;
+        return portInsts.get(this, name);
     }
     
     /**
@@ -147,8 +160,20 @@ public class EDIFCellInst extends EDIFPropertyObject implements EDIFEnumerable {
         return getCellType().getPort(name);
     }
     
+    /**
+     * Gets the sorted ArrayList of EDIFPortInsts on this cell instance as a collection.
+     * @return The collection of EDIFPortInsts on this cell.
+     */
     public Collection<EDIFPortInst> getPortInsts(){
         return portInsts == null ? Collections.emptyList() : portInsts;
+    }
+    
+    /**
+     * Gets the sorted ArrayList of EDIFPortInsts on this cell instance.
+     * @return The sorted list of EDIFPortInst objects on this cell.
+     */
+    public EDIFPortInstList getPortInstList() {
+        return portInsts == null ? EDIFPortInstList.EMPTY : portInsts;
     }
     
     /**
--- a/src/com/xilinx/rapidwright/edif/EDIFCellInst.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFCellInst.java
@@ -94,6 +94,7 @@ public class EDIFCellInst extends EDIFPropertyObject implements EDIFEnumerable {
      * contains a copy of EDIFPortInsts available at the time of invocation as returned from 
      * {@link #getPortInstList()}.      
      * @return A map of EDIFPortInst names ({@link EDIFPortInst#getName()} to the corresponding objects.
+     * @deprecated
      */
     public Map<String, EDIFPortInst> getPortInstMap(){
         if(portInsts == null) return Collections.emptyMap();
--- a/src/com/xilinx/rapidwright/edif/EDIFCellInst.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFCellInst.java
@@ -99,7 +99,7 @@ public class EDIFCellInst extends EDIFPropertyObject implements EDIFEnumerable {
     public Map<String, EDIFPortInst> getPortInstMap(){
         if(portInsts == null) return Collections.emptyMap();
         HashMap<String, EDIFPortInst> map = new HashMap<>();
-        for(EDIFPortInst e : getPortInstList()) {
+        for(EDIFPortInst e : getPortInsts()) {
             map.put(e.getName(), e);
         }
         return map;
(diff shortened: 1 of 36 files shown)


eddieh-xlnx:
Amazing!! I think sorted O(log(N)) is the way to go. Do you have a feel for how that might affect read runtime?

clavin-xlnx:
### Performance Improvements for this PR
These are the average of 3 runs on a Windows 10 machine, Intel Core i7-9850H @ 2.6GHz with 32GB RAM, SSD, loading a 1.6GB EDIF file
### Memory Footprint 
Code Version | JDK 8 | JDK 11 | JDK 17|
---------------|--------|--------|--------|
Before PR | 4172MB |  4004MB | 4008MB | 
After PR   |  2183MB |  2147MB | 2155MB| 
**Improvement** |  **1.91x** |  **1.86x** | **1.86x** | 

### Runtime 
Code Version | JDK 8 | JDK 11 | JDK 17|
---------------|--------|--------|--------|
Before PR | 90.2s |  47.2s | 42.1s | 
After PR   |  60.5s |  35.7s | 44.7s| 
**Improvement** |  **1.49x** |  **1.32x** | **0.94x** |

clavin-xlnx:
This PR contains #340.

src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -238,16 +236,13 @@ public static void connectPortBus(EDIFCell topCell, EDIFCellInst src, EDIFCellIn
...
 
-			EDIFPortInst outputPortInst = srcPortMap.get(outputPortName);
-			EDIFPortInst inputPortInst = snkPortMap.get(inputPortName);
+			EDIFPortInst outputPortInst = src.getPortInst(outputPortName);
+			EDIFPortInst inputPortInst = snk.getPortInst(inputPortName);
 			
 			@SuppressWarnings("unused")
 			EDIFPortInst outputPort = new EDIFPortInst(outputPortInst.getPort(),net,outputPortInst.getIndex(),src);
 			@SuppressWarnings("unused")
 			EDIFPortInst inputPort = new EDIFPortInst(inputPortInst.getPort(),net,inputPortInst.getIndex(),snk);
```
eddieh-xlnx, line 245: Why is this necessary?

src/com/xilinx/rapidwright/edif/EDIFNet.java
```diff
@@ -131,12 +128,35 @@ public EDIFPortInst createPortInst(EDIFPort port, int index, EDIFCellInst cellIn
...
+	 * {@link #getPortInstList()}.      
+	 * @return A map of EDIFPortInst names ({@link EDIFPortInst#getName()} to the corresponding objects.
+	 */
+	public Map<String, EDIFPortInst> getPortInstMap(){
+	    if(portInsts == null) return Collections.emptyMap();
+	    HashMap<String, EDIFPortInst> map = new HashMap<>();
+	    for(EDIFPortInst e : getPortInstList()) {
+	        map.put(e.getFullName(), e);
+	    }
+	    return map;
```
eddieh-xlnx, line 143: Is this worth keeping around? GH is telling me `EDIFNet` is the only user, and all it seems to do is build this map, do a single lookup, then throw the map away.

src/com/xilinx/rapidwright/edif/EDIFCellInst.java
```diff
@@ -88,57 +88,66 @@ public EDIFName getViewref() {
...
+    
     /**
-     * This gets a map of all the port refs on the cell instance.  
-     * @return A map of port ref names to port ref objects.
+     * Creates a new map of all the EDIFPortInst objects stored on this EDIFCellInst.  The new map
+     * contains a copy of EDIFPortInsts available at the time of invocation as returned from 
+     * {@link #getPortInstList()}.      
+     * @return A map of EDIFPortInst names ({@link EDIFPortInst#getName()} to the corresponding objects.
      */
     public Map<String, EDIFPortInst> getPortInstMap(){
```
eddieh-xlnx, line 98: Is this method worth keeping around? Looks like you've eliminated all users of it.

src/com/xilinx/rapidwright/edif/EDIFTools.java
clavin-xlnx: This code was part of a debug demo, just trying not to break it.

src/com/xilinx/rapidwright/edif/EDIFCellInst.java
clavin-xlnx: We could deprecate it and then, remove in a future release?

src/com/xilinx/rapidwright/edif/EDIFNet.java
```diff
@@ -131,12 +128,35 @@ public EDIFPortInst createPortInst(EDIFPort port, int index, EDIFCellInst cellIn
...
 	}
 	
-	public Map<String,EDIFPortInst> getPortInstMap(){
-		return portInsts == null ? Collections.emptyMap() : portInsts;
+	/**
+	 * Creates a new map of all the EDIFPortInst objects stored on this net.  The new map
+	 * contains a copy of EDIFPortInsts available at the time of invocation as returned from 
+	 * {@link #getPortInstList()}.      
+	 * @return A map of EDIFPortInst names ({@link EDIFPortInst#getName()} to the corresponding objects.
+	 */
```
clavin-xlnx, line 136:
```suggestion
	 * {@link #getPortInstList()}.      
	 * @return A map of EDIFPortInst names ({@link EDIFPortInst#getName()} to the corresponding objects.
	 * @deprecated
	 */
```

src/com/xilinx/rapidwright/edif/EDIFNet.java
clavin-xlnx: I think I meant to tag it as deprecated

src/com/xilinx/rapidwright/edif/EDIFPortInstList.java
```diff
@@ -0,0 +1,87 @@
...
+            }
+            int compare = left.getName().compareTo(rightInstName);
+            return compare == 0 ? (-1*rightPortInstName.length()) : compare;
+        } else if(rightInstName == null) {
+            int compare = left.getCellInst().getName().compareTo(rightPortInstName);
+            return compare == 0 ? left.getName().length() : compare;
+        }
+        int compare = left.getCellInst().getName().compareTo(rightInstName);
+        return compare == 0 ? left.getName().compareTo(rightPortInstName) : compare; 
+    }
```
clavin-xlnx, line 86: @eddieh-xlnx - take extra care looking at this code.

src/com/xilinx/rapidwright/edif/EDIFCellInst.java
```diff
@@ -151,8 +161,20 @@ public EDIFPort getPort(String name){
...
     public Collection<EDIFPortInst> getPortInsts(){
-        return getPortInstMap().values();
+        return portInsts == null ? Collections.emptyList() : portInsts;
+    }
+    
+    /**
+     * Gets the sorted ArrayList of EDIFPortInsts on this cell instance.
+     * @return The sorted list of EDIFPortInst objects on this cell.
+     */
+    public EDIFPortInstList getPortInstList() {
```
eddieh-xlnx, line 176: Does this new method need to exist, since it appears to be identical to `getPortInsts()` above except it returns a specific `EDIFPortInstList` type rather than a `Collection<EDIFPortInst>`? The only new user I'm seeing is:
https://github.com/Xilinx/RapidWright/blob/cd2afa438ca611ba52e1c62a704fa8b5a26a273e/src/com/xilinx/rapidwright/edif/EDIFCellInst.java#L102
which iterates over the object adding each value into a `HashMap`?

src/com/xilinx/rapidwright/edif/EDIFNet.java
```diff
@@ -131,12 +128,36 @@ public EDIFPortInst createPortInst(EDIFPort port, int index, EDIFCellInst cellIn
...
 	public Collection<EDIFPortInst> getPortInsts(){
-		return portInsts == null ? Collections.emptyList() : portInsts.values();
+		return portInsts == null ? Collections.emptyList() : portInsts;
+	}
+	
+	/**
+	 * Gets the sorted ArrayList of EDIFPortInsts on this net.
+	 * @return The sorted list of EDIFPortInst objects on this net.
+	 */
+	public EDIFPortInstList getPortInstList() {
```
eddieh-xlnx, line 159: Similarly, is this not identical to `getPortInsts()` above?

src/com/xilinx/rapidwright/edif/EDIFPortInstList.java
```diff
@@ -0,0 +1,87 @@
...
+     * @return 0 if the left and corresponding right Strings are equal.  A number less than 0 if
+     * left is lexicographically before right, or a number greater than 0 if left is after right. 
+     */
+    private int compare(EDIFPortInst left, String rightInstName, String rightPortInstName) {
+        if(left.getCellInst() == null) {
+            if(rightInstName == null) {
+                return left.getName().compareTo(rightPortInstName);
+            }
+            int compare = left.getName().compareTo(rightInstName);
+            return compare == 0 ? (-1*rightPortInstName.length()) : compare;
```
eddieh-xlnx, line 79: Does `-(rightPortInstName.length())` not work here?

src/com/xilinx/rapidwright/edif/EDIFPortInstList.java
```diff
@@ -0,0 +1,87 @@
...
+     * considered port instance to compare left against.  
+     * @param rightPortInstName This is the port instance name {@link EDIFPortInst#getName()} of the
+     * considered port instance to compare left against.
+     * @return 0 if the left and corresponding right Strings are equal.  A number less than 0 if
+     * left is lexicographically before right, or a number greater than 0 if left is after right. 
+     */
+    private int compare(EDIFPortInst left, String rightInstName, String rightPortInstName) {
+        if(left.getCellInst() == null) {
+            if(rightInstName == null) {
+                return left.getName().compareTo(rightPortInstName);
```
eddieh-xlnx, line 76:
```suggestion
                // left and right are both a top-level port insts, compare their port insts name only
                return left.getName().compareTo(rightPortInstName);
```

src/com/xilinx/rapidwright/edif/EDIFPortInstList.java
eddieh-xlnx, line 79:
```suggestion
            // left is a top-level port inst, right is not. Compare left's port inst name with the right's inst name
            int compare = left.getName().compareTo(rightInstName);
            // If the two happen to be equal, then left's full name is the prefix of right's full name and so should go before, with the remaining length the number of characters that differ
            return compare == 0 ? (-1*rightPortInstName.length()) : compare;
```

Also, do you need to account for the separator between inst name and port inst name that would otherwise appear when doing `getFullName()`?
After thinking about it some more, I think not (because this situation only occurs when comparing a top-level port inst with a non-top-level inst) unless the return value being off-by-one is critically important to the performance of the sorting algorithm?

src/com/xilinx/rapidwright/edif/EDIFPortInstList.java
```diff
@@ -0,0 +1,87 @@
...
+            }
+            int compare = left.getName().compareTo(rightInstName);
+            return compare == 0 ? (-1*rightPortInstName.length()) : compare;
+        } else if(rightInstName == null) {
+            int compare = left.getCellInst().getName().compareTo(rightPortInstName);
+            return compare == 0 ? left.getName().length() : compare;
```
eddieh-xlnx, line 82:
```suggestion
            // right is a top-level port inst, but left is not. Compare left's inst name with right's port inst name.
            int compare = left.getCellInst().getName().compareTo(rightPortInstName);
            // If the two happen to be equal, then right's full name is a prefix of left's full name and thus left should go after right.
            return compare == 0 ? left.getName().length() : compare;
```

src/com/xilinx/rapidwright/edif/EDIFPortInstList.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+        int compare = left.getCellInst().getName().compareTo(rightInstName);
+        return compare == 0 ? left.getName().compareTo(rightPortInstName) : compare;
```
eddieh-xlnx, line 85:
```suggestion
        // left and right are both non top-level port insts. Compare their inst names first, then compare their port inst names.
        int compare = left.getCellInst().getName().compareTo(rightInstName);
        return compare == 0 ? left.getName().compareTo(rightPortInstName) : compare; 
```

src/com/xilinx/rapidwright/edif/EDIFNet.java
eddieh-xlnx: Looks good.

src/com/xilinx/rapidwright/edif/EDIFCellInst.java
```diff
@@ -88,57 +88,66 @@ public EDIFName getViewref() {
...
     }
-
+    
     /**
-     * This gets a map of all the port refs on the cell instance.  
-     * @return A map of port ref names to port ref objects.
+     * Creates a new map of all the EDIFPortInst objects stored on this EDIFCellInst.  The new map
+     * contains a copy of EDIFPortInsts available at the time of invocation as returned from 
+     * {@link #getPortInstList()}.      
+     * @return A map of EDIFPortInst names ({@link EDIFPortInst#getName()} to the corresponding objects.
```
eddieh-xlnx, line 96:
```suggestion
     * @return A map of EDIFPortInst names ({@link EDIFPortInst#getName()} to the corresponding objects.
     * @deprecated
```

eddieh-xlnx (approved):
Just a few minor changes from me.

src/com/xilinx/rapidwright/edif/EDIFCellInst.java
clavin-xlnx: Is this change tagging it to be `@deprecated`?  It seems like it already is.

src/com/xilinx/rapidwright/edif/EDIFPortInstList.java
clavin-xlnx: No, you are correct.  Since we are handling the compare strictly outside of the `getFullName()` scenario, it never becomes an issue.

src/com/xilinx/rapidwright/edif/EDIFCellInst.java
eddieh-xlnx: My mistake, sorry!

PR #345, opened 2022-02-22
Fix #117 by merging {set,update}CellType into new set overload

### LINKED ISSUE #117: EDIFCellInst.setCellType needed?
Hi!

I am unsure about the use case of  `EDIFCellInst.setCellType` compared to `EDIFCellInst.updateCellType`. `setCellType` leaves the `EDIFPortInst`s pointing to the ports of the former cell, while `updateCellType` updates them.

Is it ever desirable to leave them pointing to the former cell?

This behavior was very confusing for me. It is very intuitive to call `setCellType` to change a CellType, so I used it in my code. When that produced invalid EDIF files, I modified `setCellType` to do more or less the same as `updateCellType`. When I later tidied up my code for submitting #116 I discovered the existence of `updateCellType`.

I would suggest having `setCellType` do the same as the current `updateCellType` (`updateCellType` would then either be deleted or just redirect to `setCellType` for backwards compatibility). If the old behaviour of `setCellType` is needed at all, it could be provided under a different name.

Kind Regards,
Jakob

eddieh-xlnx:
I agree with #117 that having two `EDIFCellInst` methods that are synonyms of each other is not particularly useful. 

This is the first of two proposed fixes, and is the safer option:

`EDIFCellInst.updateCellType(EDIFCell)`
`EDIFCellInst.setCellType(EDIFCell)`

now merged into:

``EDIFCellInst.setCellType(EDIFCell, boolean)`

The boolean determines whether to perform the original `updateCellType` functionality (which updates the port references on the portInsts) or whether to just do the original `setCellType`.

All references to the former two methods have been changed to the new overload, and those former methods marked as being `@deprecated`. This should be a non-functional change.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -835,7 +835,7 @@ public class DesignTools {
 		}
 		inst.getCellType().getLibrary().removeCell(inst.getCellType());
 		netlist.migrateCellAndSubCells(cell.getTopEDIFCell(), true);
-		inst.updateCellType(cell.getTopEDIFCell());
+		inst.setCellType(cell.getTopEDIFCell(), true);
 		netlist.removeUnusedCellsFromAllWorkLibraries();
 		
 		// Add placement information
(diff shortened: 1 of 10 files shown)


eddieh-xlnx:
#346 was preferred over this.

PR #346, opened 2022-02-22
Fix #117 by setCellType doing what updateCellType did

### LINKED ISSUE #117: EDIFCellInst.setCellType needed?

eddieh-xlnx:
I agree with #117 that having two `EDIFCellInst` methods that are synonyms of each other is not particularly useful. 

This is the second of two proposed fixes, and is the riskier option.

`EDIFCellInst.setCellType(EDIFCell)` renamed to `setCellTypeRaw(EDIFCell)` (latter is not used, but kept around just in case it is useful in the future)
`EDIFCellInst.updateCellType(EDIFCell)` renamed to `setCellType(EDIFCell)`

with `updateCellType()` marked as `@deprecated` while still calling the new `setCellType`.

~The reasoning here is that one should never update the cell type without updating the port references of its portInsts. The only exception I'm not terribly confident about is:~
https://github.com/Xilinx/RapidWright/blob/7177e016ad6c1beefb1778d5a80aab7a468f4b3f/src/com/xilinx/rapidwright/design/DesignTools.java#L1820-L1835
~which does some portInst parent cell juggling too, and is currently the only reason for `setCellTypeRaw` existing.~ (see resolved discussion)

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -1833,9 +1833,9 @@ public class DesignTools {
 			destNetlist.migrateCellAndSubCells(cellInst.getCellType());
 			EDIFHierCellInst bbInst = destNetlist.getHierCellInstFromName(e.getValue());
 			bbInst.getInst().setCellTypeRaw(cellInst.getCellType());
-            for(EDIFPortInst portInst : bbInst.getInst().getPortInsts()) {
-            	portInst.getPort().setParentCell(cellInst.getCellType());
-            }
+			for(EDIFPortInst portInst : bbInst.getInst().getPortInsts()) {
+				portInst.getPort().setParentCell(cellInst.getCellType());
+			}
 			instsWithSeparator.add(e.getKey() + EDIFTools.EDIF_HIER_SEP);
 		}
 		destNetlist.resetParentNetMap();
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -1832,10 +1832,7 @@ public class DesignTools {
 			EDIFHierCellInst cellInst = src.getNetlist().getHierCellInstFromName(e.getKey());
 			destNetlist.migrateCellAndSubCells(cellInst.getCellType());
 			EDIFHierCellInst bbInst = destNetlist.getHierCellInstFromName(e.getValue());
-			bbInst.getInst().setCellTypeRaw(cellInst.getCellType());
-			for(EDIFPortInst portInst : bbInst.getInst().getPortInsts()) {
-				portInst.getPort().setParentCell(cellInst.getCellType());
-			}
+			bbInst.getInst().setCellType(cellInst.getCellType());
 			instsWithSeparator.add(e.getKey() + EDIFTools.EDIF_HIER_SEP);
 		}
 		destNetlist.resetParentNetMap();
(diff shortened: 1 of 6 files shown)


clavin-xlnx (approved):
I think this is probably the better option.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1832,10 +1832,10 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
 			destNetlist.migrateCellAndSubCells(cellInst.getCellType());
 			EDIFHierCellInst bbInst = destNetlist.getHierCellInstFromName(e.getValue());
-			bbInst.getInst().setCellType(cellInst.getCellType());
-            for(EDIFPortInst portInst : bbInst.getInst().getPortInsts()) {
-            	portInst.getPort().setParentCell(cellInst.getCellType());
-            }
+			bbInst.getInst().setCellTypeRaw(cellInst.getCellType());
+			for(EDIFPortInst portInst : bbInst.getInst().getPortInsts()) {
+				portInst.getPort().setParentCell(cellInst.getCellType());
+			}
```
eddieh-xlnx, line 1838: @clavin-xlnx Are you familiar with this block of code?
What's the reason for forcibly setting the cell type but not updating its portInsts?

clavin-xlnx: This code does update the portInsts, just independently, in lines 1836-1838.  These lines theoretically could be swapped for a call to the newly refactored `setCellType()`.

eddieh-xlnx: It looks like those lines don't update the portInsts in the same way:
https://github.com/Xilinx/RapidWright/blob/b9d72631f0ee2b9fdccc2c846ccc0a98f0704f06/src/com/xilinx/rapidwright/edif/EDIFCellInst.java#L200-L210
Lines 1836-1838 update the parent cell of the portInsts' ports, but `setCellType` updates the portInsts themselves. Are those two things equivalent?

It doesn't look like we have any test coverage on `DesignTools.copyImplementation()` currently for me to be able to make this change with confidence.

eddieh-xlnx: Oh wait, we do have one (admittedly specific) test that calls `DesignTools.copyImplementation()`: https://github.com/Xilinx/RapidWright/blob/131331d75ae66678901e5eb614fd5145428eb0eb/test/com/xilinx/rapidwright/design/TestDesignTools.java#L59

eddieh-xlnx: Okay, I think changing the cell that the port belongs to is equivalent to setting a new port acquired from that new cell.

PR #348, opened 2022-02-24
Easy command to generate all interchange device files

clavin-xlnx:


clavin-xlnx:
Ideally, someone without RapidWright installed, could generate all Interchange devices with these commands (assuming `git` and `java` are available on the system):
```
git clone https://github.com/Xilinx/RapidWright.git
cd RapidWright
./gradlew run --args="GenerateInterchangeDevices <devices_location_dir>"
```

PR #358, opened 2022-03-04
RWRouteConfig info clarification and an oversight fix

YunxZhou:
This PR contains the clarified RWRouteConfig info and the fix of an oversight. Fixing the oversight has no impact on the overall runtime and QoR.


YunxZhou:
I forgot to merge before creating a new PR. Would redo it.

PR #359, opened 2022-03-05
Binary EDIF Format 

clavin-xlnx:
Aims to provide a binary equivalent to text-based EDIF that is an order of magnitude smaller in file size and 5-10X faster to load.  Should include an optional switch to cache loaded text-based EDIF.

### DIFF
--- a/src/com/xilinx/rapidwright/edif/BinaryEDIFReader.java
+++ b/src/com/xilinx/rapidwright/edif/BinaryEDIFReader.java
@@ -1,0 +1,224 @@
+package com.xilinx.rapidwright.edif;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Map;
+
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+import com.xilinx.rapidwright.util.FileTools;
+
+/**
+ * A Reader for the RapidWright Binary EDIF Format
+ *  
+ * Provides a binary alternative to textual EDIF that is ~10-15X smaller and loads 5-10X faster.
+ * This is intended as a cached version of text-based EDIF as it cannot be read by Vivado.  One
+ * additional tradeoff is that it takes about 2.5-3X longer to write than text-based EDIF.
+ */
+public class BinaryEDIFReader {
+    
+    /**
+     * Reads an EDIFName object from kryo-based input stream.
+     * @param o The object to populate
+     * @param is Kryo-based input stream
+     * @param strings Indexed string lookup
+     * @return True if this object has a non-zero property map, false if none
+     * @see BinaryEDIFWriter#writeEDIFName(EDIFName, Output, Map, boolean)
+     */
+    private static boolean readEDIFName(EDIFName o, Input is, String[] strings) {
+        int nameIdx = is.readInt();
+        if((nameIdx & BinaryEDIFWriter.EDIF_NAME_FLAG) == BinaryEDIFWriter.EDIF_NAME_FLAG) {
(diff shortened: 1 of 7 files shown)


src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -773,7 +765,11 @@ public static EDIFNetlist readEdifFile(Path edifFileName){
 			}
 			edif.setEncryptedCells(new ArrayList<>(Arrays.asList(ednFiles)));
 		}
-		
+		if(ENABLE_EDIF_BINARY_CACHING) {
+		    Path bedif = edifFileName.getParent().resolve(
+                    edifFileName.getFileName().toString().replace(".edf", ".bedf"));
+		    BinaryEDIFWriter.writeBinaryEDIF(bedif, edif);
```
eddieh-xlnx, line 771: Can we fail more gracefully if we don't have write permission? This would make the flag a lot safer to enable all the time.

src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -674,21 +679,9 @@ public static void main(String[] args) {
...
-				if(EDIFTools.EDIF_DEBUG) FileTools.writeObjectToKryoFile(edifFileName + ".dat", edif);
-				t.stop();
-			}
-			
-		}
+		t.start("Read EDIF from ASCII EDIF file");
+		edif = EDIFTools.loadEDIFFile(edifFileName);
+		t.stop(); 
 		t.start("Write EDIF file");
 		EDIFTools.writeEDIFFile(edifFileName.replace(".edf", "_byu.edf"), edif, "partname");
```
eddieh-xlnx, line 686: Is the `_byu` suffix important?

src/com/xilinx/rapidwright/edif/BinaryEDIFReader.java
```diff
@@ -0,0 +1,254 @@
...
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+import com.xilinx.rapidwright.util.FileTools;
+
+/**
+ * A Reader for the RapidWright Binary EDIF Format
+ *  
+ * Provides a binary alternative to textual EDIF that is ~10-15X smaller and loads 5-10X faster.
+ * This is intended as a cached version of text-based EDIF as it cannot be read by Vivado.  One
+ * additional tradeoff is that it takes about 2.5-3X longer to write than text-based EDIF.
```
eddieh-xlnx, line 41: Maybe this is clearer wording?
```suggestion
 * This is intended as a RapidWright-only cache (with the extension *.bedf) of this text-based
 * EDIF and cannot be read by Vivado.  One additional tradeoff is that it takes about 2.5-3X
 * longer to write than text-based EDIF, but it is expected that this binary alternative will
 * be written once and read many times.
```

eddieh-xlnx (approved):
Works for me. Just a few suggestions.

src/com/xilinx/rapidwright/edif/EDIFTools.java
clavin-xlnx, line 686:
```suggestion
		EDIFTools.writeEDIFFile(edifFileName.replace(".edf", "_out.edf"), edif, "partname");
```

clavin-xlnx: No, it could be made more generiic.

src/com/xilinx/rapidwright/edif/EDIFTools.java
clavin-xlnx: Made some changes so that in the face of a failed write/read of the binary EDIF, a message is printed but process continues.  In the case of reading, it will fall back to just reading the EDIF file.

PR #361, opened 2022-03-08
RWRouteConfig info clarification and fix of an oversight

YunxZhou:
This PR contains clarification on the RWRouteConfig info printing to avoid concerns and the fix of an oversight. 
The oversight was in the comparator for sorting critical connections when the number of critical connections is beyond a threshold. The fix has no impact on the overall runtime or QoR when running with the default threshold.


clavin-xlnx:
@YunxZhou - It looks like this pull request is failing because it changed the test/RapidWrightDCP area.  It removed some DCPs that are essential to the tests.  Could you revert those changes and try again?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -849,7 +848,8 @@ private void setRerouteCriticality() {
 			criticalConnections.sort(new Comparator<Connection>() {
 				@Override
 				public int compare(Connection connection1, Connection connection2) {
-					return connection1.getCriticality() < connection1.getCriticality()? 1 : -1;
+					if(connection1.getCriticality() == connection2.getCriticality()) return 0;
+					else return connection1.getCriticality() < connection2.getCriticality()? 1 : -1;
```
clavin-xlnx, line 852:
```suggestion
					return Float.compare(connection1.getCriticality(),connection2.getCriticality());
```
@YunxZhou - would this code be simpler?  Is it important that the return values be `-1` and `1` or does it just need to be negative and positive integers?

clavin-xlnx, line 852:
```suggestion
					return Float.compare(connection1.getCriticality(),connection2.getCriticality());
```
@YunxZhou - would this code be simpler?  Is it important that the return values be `-1` and `1` or does it just need to be negative and positive integers?

clavin-xlnx: See https://docs.oracle.com/javase/7/docs/api/java/lang/Float.html#compare(float,%20float) for API details.

clavin-xlnx: See https://docs.oracle.com/javase/7/docs/api/java/lang/Float.html#compare(float,%20float) for API details.

YunxZhou: @clavin-xlnx, thanks for the comment. Simplifying is good. The compare just needs to be negative and positive integers for the sorting. The order needs to be the other way around. I made a commit based on this

PR #365, opened 2022-03-11
Enabling parallelism (for write DCP) by default

clavin-xlnx:

### DIFF
--- a/src/com/xilinx/rapidwright/util/ParallelismTools.java
+++ b/src/com/xilinx/rapidwright/util/ParallelismTools.java
@@ -50,11 +50,18 @@ import java.util.concurrent.TimeUnit;
  */
 public class ParallelismTools {
     /**
-     * Name of the environment variable to enable parallel processing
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/util/ParallelismTools.java
```diff
@@ -50,11 +50,18 @@
...
      */
     public static final String RW_PARALLEL = "RW_PARALLEL";
 
-    private static boolean parallel = System.getenv(RW_PARALLEL) != null;
+    private static boolean parallel = true;
+    
+    static {
+        String value = System.getenv(RW_PARALLEL);
+        if(value != null) {
+            setParallel((value.equals("0") || value.toLowerCase().equals("false")) ? false : true);
```
eddieh-xlnx, line 62:
```suggestion
            setParallel(!(value.equals("0") || value.toLowerCase().equals("false")));
```

eddieh-xlnx (approved):

PR #370, opened 2022-03-15
Adds Missing Apache 2.0 License Source Header

clavin-xlnx:
Also updates some Makefiles to point the Interchange at the proper runtime jar and also adds a new Makefile target to add source headers to files that are missing them.

In the future, when adding a new file, run `make ensure_headers` and an Apache 2.0 license header will be added automatically to any file that is missing one populated with the current user and year.


eddieh-xlnx (approved):
Blimey: 73 files!! I read every single line you added. Honest.

PR #373, opened 2022-03-16
Fix bug and update printout for ReplaceFrameData

pongstorn-xilinx:
We need to consider moving ReplaceFrameData, RelocateModulesIntoBlackboxes and RelocateBitstreamByRow from Util.   Util directory seems to be for low-level misc. class.


src/com/xilinx/rapidwright/util/ReplaceFrameData.java
```diff
@@ -653,15 +658,15 @@ public static void main(String[] args) {
 		// Run
 		if (op.equals("extract")) {
 
-			CodePerfTracker t = new CodePerfTracker("Elapsed time", false);
+			CodePerfTracker t = new CodePerfTracker("Extract frame data", false);
```
clavin-xlnx, line 661: The `if(op.equals("extract"))` and `else if (op.equals("replace"))` code bodies look very similar.  Can we reduce the duplication down to just the differences?

pongstorn-xilinx: It appears that way due to CodePerfTracker.  Because runtime is not important and it is within one second, I removed the use of CodePerfTracker.   In addition, I move reading bitstream and creating ReplaceFrameData objects out before the if (these are the true common code).

clavin-xlnx (approved):

PR #377, opened 2022-03-17
Fix for RW_PARALLEL != 0

eddieh-xlnx:
Setting the environment variable `RW_PARALLEL` to not zero means to enable parallelism.

Doing so means that ` pool.prestartAllCoreThreads();` gets called, but `pool` has not yet been initialized, so juggle the ordering.

### DIFF
--- a/src/com/xilinx/rapidwright/util/ParallelismTools.java
+++ b/src/com/xilinx/rapidwright/util/ParallelismTools.java
@@ -54,15 +54,6 @@ public class ParallelismTools {
      */
     public static final String RW_PARALLEL = "RW_PARALLEL";
 
-    private static boolean parallel = true;
-    
-    static {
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/util/ParallelismTools.java
```diff
@@ -76,6 +67,13 @@
                 return t;
             });
 
+    private static boolean parallel = true;
+
+    static {
+        String value = System.getenv(RW_PARALLEL);
+        setParallel(!(value.equals("0") || value.equalsIgnoreCase("false")));
```
eddieh-xlnx, line 74: Oops.
```suggestion
        setParallel(value == null || !(value.equals("0") || value.equalsIgnoreCase("false")));
```

clavin-xlnx (approved):

PR #379, opened 2022-03-22
Rewrite of EDIFTools.connectPortInstsThruHier() 

clavin-xlnx:
Improves and makes more robust EDIFTools.connectPortInstsThruHier() by leveraging the EDIFHierCellInst rewrite last year.  

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -1036,7 +1036,8 @@ public class DesignTools {
 	    }
 
 	    HashSet<PIP> toRemove = new HashSet<>();
-
+	    ArrayList<Node> updateFanout = new ArrayList<>();
+        
 	    for(SitePinInst p : pins) {
 	        if(p.getSite() == null) continue;
 	        if(p.getNet() != net) continue;
@@ -1052,9 +1053,13 @@ public class DesignTools {
 	            fanoutCount--;
 	            atReversedBidirectionalPip = true;
 	        }
+	        updateFanout.clear();
 	        while(curr != null && curr.size() == 1 && fanoutCount < 2){
 	            PIP pip = curr.get(0);
+	            
 	            toRemove.add(pip);
+	            Node startNode = pip.getStartNode();
+	            updateFanout.add(startNode);
 	            if (new Node(pip.getTile(), pip.getStartWireIndex()).equals(sink) && !atReversedBidirectionalPip) {
 	                // reached the source and there is another branch starting with a reversed
 	                // bidirectional PIP ... don't traverse it
@@ -1100,6 +1105,13 @@ public class DesignTools {
 	            si.unrouteIntraSiteNet(belPin, belPin);
 	        }
 	        p.setRouted(false);	        
+	        for(Node startNode : updateFanout) {
+	            Integer newFanout = fanout.get(startNode);
+	            if(newFanout != null) {
+	                newFanout--;
+	                fanout.put(startNode, newFanout);
+	            }
+	        }
 	    }
 	    ArrayList<PIP> updatedPIPs = new ArrayList<>();
 	    for(PIP pip : net.getPIPs()){
@@ -1971,6 +1983,7 @@ public class DesignTools {
 			if(pin.isInput()) continue;
 			if(pin.getBEL().getBELClass() == BELClass.RBEL) {
 				SitePIP p = siteInst.getUsedSitePIP(pin.getBELName());
+				if(p == null) continue;
 				for(BELPin pin2 : p.getInputPin().getSiteConns()) {
 					if(pin2.isOutput()) {
 						return pin2;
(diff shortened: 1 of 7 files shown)


src/com/xilinx/rapidwright/edif/EDIFHierCellInst.java
```diff
@@ -138,11 +139,62 @@ public EDIFHierCellInst getParent() {
...
+    
+    /**
+     * Given this instance and the provided instance o, it determines the closest common ancestor
+     * between the two instances.  In some cases, this can default to the top cell design instance.
+     * For example, if this instance = "a/b/c/d" and o = "a/b/e/f", the method would return "a/b".  
+     * @param o The other instance to check for a common ancestor.
+     * @return The closest common ancestor between this instance and the provided instance.
+     */
+    public EDIFHierCellInst getCommonAncestor(EDIFHierCellInst o) {
+        EDIFCellInst[] oCellInsts = o.cellInsts;
```
jakobwenzel, line 185:
```suggestion
        if (!isAbsolute() || !o.isAbsolute()) {
            throw new RuntimeException("Can only get a common ancestor of absolute EDIFHierCellInsts");
        }
        EDIFCellInst[] oCellInsts = o.cellInsts;
```
I don't think it makes sense to call this function for relative EHCIs. Relative ones might not share a common ancestor. In that case, we would raise an exception in https://github.com/Xilinx/RapidWright/blob/2382f67c2501cfb9074e516dcb1e0eecd7a1919e/src/com/xilinx/rapidwright/edif/EDIFHierCellInst.java#L46 . Might be better to just forbid it from the start.

src/com/xilinx/rapidwright/edif/EDIFHierCellInst.java
```diff
@@ -138,11 +139,62 @@ public EDIFHierCellInst getParent() {
     public EDIFHierCellInst getChild(EDIFCellInst relativeChild) {
         return new EDIFHierCellInst(cellInsts, cellInsts.length+1, relativeChild);
     }
+    
+    public List<EDIFHierCellInst> getImmediateChildren(){
+        List<EDIFHierCellInst> children = new ArrayList<>();
+        for(EDIFCellInst child : getCellType().getCellInsts()) {
+            children.add(getChild(child));
+        }
```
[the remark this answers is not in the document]
clavin-xlnx: Thanks, that method seems like the better way to go.  I hadn't noticed it previously.

src/com/xilinx/rapidwright/edif/EDIFHierCellInst.java
```diff
@@ -138,11 +139,62 @@ public EDIFHierCellInst getParent() {
...
+    /**
+     * Checks if the provided instance is an ancestor (hierarchical parent) of this instance. For 
+     * example, if this = "disneyland/tomorrow_land/space_mountain" and 
+     * inst = "disneyland/tomorrow_land", this method would return true.  however, if 
+     * inst = "disneyland/adventure_land", it would return false. 
+     *  
+     * @param inst The hierarchical instance in question to check if it is  
+     * @return True if the provided instance is a hierarchical ancestor of this instance.
+     */
+    public boolean isAncestor(EDIFHierCellInst inst) {
```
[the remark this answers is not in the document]
clavin-xlnx: Refactored

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -993,6 +992,313 @@ public static boolean removeConnectedRouting(Net net, Node node){
...
+	    ArrayList<PIP> updatedPIPs = new ArrayList<>();
+	    for(PIP pip : net.getPIPs()){
+	        if(!toRemove.contains(pip)) updatedPIPs.add(pip);
+	    }
+	    net.setPIPs(updatedPIPs);
+	}
+
+
+	/**
+	 * This method will completely remove a placed cell (both logical and physical) from a design.
```
eddieh-xlnx, line 1125:
```suggestion
	 * This method will completely remove a placed cell (both logical and physical) from a design.
	 * In the case where the removed cell is the last user of a shared control signal (CLK, CE, SR) then that pin will also be removed and unrouted immediately if deferRemovals is null, otherwise it is added to this map.
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -993,6 +992,313 @@ public static boolean removeConnectedRouting(Net net, Node node){
...
+	 * @param design The design where the cell is instantiated
+	 * @param cell The cell to remove
+	 * @param deferRemovals An optional map that, if passed in non-null will be populated with 
+	 * site pins marked for removal.  The map allows a persistent tracking if this method is called
+	 * many times as the process is expensive without batching. 
+	 */
+	public static void fullyRemoveCell(Design design, Cell cell, Map<Net, Set<SitePinInst>> deferRemovals) {
```
eddieh-xlnx, line 1132: Does `deferRemovals` need to be a set? Would you expect a pin to be added more than once since it should only be added when the last cell is removed?

test/com/xilinx/rapidwright/edif/TestEDIFHierCellInst.java
```diff
@@ -0,0 +1,56 @@
...
+        while(!q.isEmpty()) {
+            EDIFHierCellInst curr = q.poll();
+            Assertions.assertTrue(curr.isDescendantOf(topInst));
+            Assertions.assertFalse(topInst.isDescendantOf(curr));
+            curr.addChildren(q);
+        }
+    }
+    
+    @Test
+    public void testgetCommonAncestor() {
```
eddieh-xlnx, line 32:
```suggestion
    public void testGetCommonAncestor() {
```

eddieh-xlnx (approved):
Looks reasonable, and best of it all it works!

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx: > Does `deferRemovals` need to be a set? 
No, only when the caller wants to defer.  

> Would you expect a pin to be added more than once since it should only be added when the last cell is removed?
Pins requiring removal will likely exist for each cell removal.  For example, `SitePinInst` objects will exist for a flop's `Q` output.  In theory we might be able to get away with a List, but in practice, it is probably safer to just use a set as I can't foresee every possible scenario.

eddieh-xlnx: That's fair. If it's not cut and dry then let's stick with the status quo.

PR #381, opened 2022-03-23
add parallel EDIF parser

jakobwenzel:
~2.5x faster that existing parser

Most of the `EDIFParser`'s code has ben moved to `AbstractEDIFParser` as well as `EDIFTokenizer` to facilitate reuse in the parallel parser.


src/com/xilinx/rapidwright/edif/EDIFPortCache.java
```diff
@@ -0,0 +1,52 @@
...
+public class EDIFPortCache {
+    private final Map<String, EDIFPort> cache;
+    private final EDIFCell cell;
+
+    public EDIFPortCache(EDIFCell cell) {
+        this.cell = cell;
+        /*if (cell.getPorts().size()<1000) {
+            cache = null;
+            return;
+        }*/
```
eddieh-xlnx, line 38: Remove?
```suggestion
```

src/com/xilinx/rapidwright/edif/EDIFPortCache.java
```diff
@@ -0,0 +1,52 @@
...
+ * 
+ */
+ 
+package com.xilinx.rapidwright.edif;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+public class EDIFPortCache {
```
eddieh-xlnx, line 29: Can we add one liner for Javadoc please?

src/com/xilinx/rapidwright/edif/EDIFTokenizer.java
```diff
@@ -0,0 +1,486 @@
...
+import java.io.InputStream;
+import java.io.SequenceInputStream;
+import java.io.UncheckedIOException;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+
+import org.apache.commons.io.IOUtils;
+
+public class EDIFTokenizer implements AutoCloseable {
```
eddieh-xlnx, line 36: Javadoc please

src/com/xilinx/rapidwright/edif/NameUniquifier.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+
+/**
+ * Optimize memory footprint of loaded EDIF by deduplicating equal strings
+ */
+public class NameUniquifier {
```
eddieh-xlnx, line 33: We already have a `src.com.xilinx.rapidwright.util.StringPool`. Looks like this is superior though, so consider replacing that one with this one?

src/com/xilinx/rapidwright/edif/ParallelEDIFParser.java
```diff
@@ -0,0 +1,283 @@
...
+import com.xilinx.rapidwright.device.Device;
+import com.xilinx.rapidwright.tests.CodePerfTracker;
+import com.xilinx.rapidwright.util.ParallelismTools;
+import com.xilinx.rapidwright.util.function.InputStreamSupplier;
+
+/**
+ * Fast EDIF Parser using parallelism
+ */
+public class ParallelEDIFParser implements AutoCloseable{
+    private static final long SIZE_PER_THREAD = EDIFTokenizer.DEFAULT_MAX_TOKEN_LENGTH * 8L;
```
eddieh-xlnx, line 47:
```suggestion
    /** TODO */
    private static final long MIN_BYTES_PER_THREAD = EDIFTokenizer.DEFAULT_MAX_TOKEN_LENGTH * 8L;
```

src/com/xilinx/rapidwright/edif/ParallelEDIFParser.java
```diff
@@ -0,0 +1,283 @@
...
+
+    private int numberOfThreads;
+    public EDIFNetlist parseEDIFNetlist(CodePerfTracker t) throws IOException {
+
+        t.start("Initialize workers");
+        initializeWorkers();
+        numberOfThreads = workers.size();
+
+        t.stop().start("Parse First Token");
+        final List<ParallelEDIFParserWorker> failedWorkers = ParallelismTools.maybeToParallel(workers.stream())
```
eddieh-xlnx, line 103: Stream parallelism uses `ForkJoinPool.commonPool()` but I would prefer parallel activity to be funneled through `ParallelismTools` (via `invokeAll`) and use the thread pool it has. This way, we can fully control current and future parallel activity ourselves, and this could possibly lead to a more scalable solution because it would support the case where number of workers/tasks exceeds the number of processors in order to deal with work imbalance arising from EDIF parsing or other parsing activities.

What do you think?

src/com/xilinx/rapidwright/edif/ParallelEDIFParser.java
```diff
@@ -0,0 +1,283 @@
...
+        this(p, Files.size(p));
+    }
+
+    protected ParallelEDIFParserWorker makeWorker(long offset) throws IOException {
+        return new ParallelEDIFParserWorker(fileName, inputStreamSupplier.get(), offset, uniquifier, maxTokenLength);
+    }
+
+    private int calcThreads(long fileSize) {
+        int maxUsefulThreads = Math.max((int) (fileSize / SIZE_PER_THREAD),1);
+        int processors = Runtime.getRuntime().availableProcessors();
```
eddieh-xlnx, line 80: Could you replace this with a call to a new method like `ParallelismTools.maxParallelism()`?

src/com/xilinx/rapidwright/edif/ParallelEDIFParserWorker.java
```diff
@@ -0,0 +1,427 @@
...
+import com.xilinx.rapidwright.util.ParallelismTools;
+
+/**
+ * Worker thread inside the parallel EDIF parser
+ */
+public class ParallelEDIFParserWorker extends AbstractEDIFParser implements AutoCloseable{
+    /**
+     * Number of ports in a cell above which a map is used for port name lookup
+     */
+    private static final int PORT_LOOKUP_MAP_LIMIT = 300;
```
eddieh-xlnx, line 46:
```suggestion
    private static final int PORT_LOOKUP_MAP_THRESHOLD = 300;
```

test/com/xilinx/rapidwright/edif/TestingParallelEDIFParser.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+import com.xilinx.rapidwright.util.function.InputStreamSupplier;
+
+public class TestingParallelEDIFParser extends ParallelEDIFParser{
```
eddieh-xlnx, line 33:
```suggestion
public class ParallelEDIFParserTestSpecificOffsets extends ParallelEDIFParser{
```

eddieh-xlnx (commented):
A few initial comments to start the discussion.

src/com/xilinx/rapidwright/edif/ParallelEDIFParser.java
```diff
@@ -0,0 +1,283 @@
...
+        int threads = calcThreads(fileSize);
+        long offsetPerThread = fileSize / threads;
+        for (int i=0;i<threads;i++) {
+            ParallelEDIFParserWorker worker = makeWorker(i*offsetPerThread);
+            workers.add(worker);
+        }
+    }
+
+    private int numberOfThreads;
+    public EDIFNetlist parseEDIFNetlist(CodePerfTracker t) throws IOException {
```
eddieh-xlnx, line 96: Is it possible to make CodePerfTracker optional for those who don't need to know?

eddieh-xlnx: Maybe just an overload with no args which internally calls this with `CodePerfTracker.SILENT`?

src/com/xilinx/rapidwright/edif/EDIFTokenizer.java
```diff
@@ -0,0 +1,486 @@
...
+     * @param length1 first part length
+     * @param start2 second part start
+     * @param length2 second part length
+     * @return the string assembled from the two locations
+     */
+    public static String byteArrayToStringMulti(byte[] buffer, int start1, int length1, int start2, int length2) {
+        //To support multi-byte characters being split between the parts, we have to take
+        // care to first concatenate, then decode.
+        //TODO benchmark which approach is faster. Arraycopy is probably faster for small strings?
+        if (length1+length2 < 10000) {
```
clavin-xlnx, line 132: Perhaps label the magic number?

clavin-xlnx, line 132: Perhaps label the magic number?

src/com/xilinx/rapidwright/edif/ParallelEDIFParser.java
jakobwenzel: > Maybe just an overload with no args which internally calls this with `CodePerfTracker.SILENT`?

Oohh nice, didn't know that existed!

src/com/xilinx/rapidwright/edif/EDIFTokenizer.java
jakobwenzel: I've now benchmarked it. The streams approach is always slower. So no need for the magic number anymore.

src/com/xilinx/rapidwright/edif/NameUniquifier.java
jakobwenzel: Ok, I will replace the old class with this one.

src/com/xilinx/rapidwright/edif/ParallelEDIFParser.java
eddieh-xlnx: `ParallelismTools` was built to be a *parallel* (hee hee) to `ForkJoinPool`, which also has one thread less than the number of procs. The reason for this design is that, for blocking operations (i.e. an `invoke` which blocks until all jobs are done) the blocking thread will also "steal" tasks from the queue. In non-blocking operations, (i.e. a `submit`) then the submitting thread can continue and do some other async work without causing the number of working threads to exceed the number of processors.

Where it differs from the `ForkJoinPool` is that it has a shared thread queue (rather than one queue per thread) which I've found to have give better thread utilization for coarse-grained tasks, rather than the finer-grained divide-and-conquer tasks generated by `parallelStream()`.

I'm not fully convinced that what I've built is indeed the best thing, so if you do dive into it please critique!

src/com/xilinx/rapidwright/edif/ParallelEDIFParser.java
```diff
@@ -0,0 +1,288 @@
...
+                        });
+        //Name the port insts, order is not important
+        t.stop().start("Name port insts");
+                ParallelismTools.maybeToParallel(workers.stream())
+                        .flatMap(w -> w.linkPortInstData.stream())
+                        .forEach(list -> {
+                            for (ParallelEDIFParserWorker.LinkPortInstData linkPortInstData : list) {
+                                linkPortInstData.name();
+                            }
+                        });
```
eddieh-xlnx, line 245: Spacing

src/com/xilinx/rapidwright/edif/AbstractEDIFParser.java
```diff
@@ -0,0 +1,501 @@
...
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+
+import com.xilinx.rapidwright.util.StringPool;
+
+public abstract class AbstractEDIFParser {
```
eddieh-xlnx, line 36: Maybe this could be called `AbstractEDIFParserWorker` because it doesn't provide the base class for both the seq and parallel EDIFParser-s, just the core worker functionality?

src/com/xilinx/rapidwright/edif/ParallelEDIFParser.java
```diff
@@ -0,0 +1,288 @@
...
+        t.stop();
+    }
+
+    private EDIFNetlist mergeParseResults(CodePerfTracker t) {
+        EDIFNetlist netlist = Objects.requireNonNull(workers.get(0).netlist);
+        netlist.setDesign(getEdifDesign());
+
+        t.stop().start("Assemble Libraries");
+        addCellsAndLibraries(netlist);
+        t.stop().start("process cell inst links");
```
eddieh-xlnx, line 266:
```suggestion
        t.stop().start("Process cell inst links");
```

jakobwenzel:
Ready for second round of reviews.

`ParallelismTools.invokeAll` was hard to use if you have a list of data objects and want to perform a task on all of them. I added a utility method for that. I added those both for consumer-style tasks without a return value and for function-style tasks with a return value. When using lambdas, Java can't figure out whether to infer a `Consumer` or a `Function`, so I used different method names instead of overloading one method.

jakobwenzel:
Rebased to master and added missing signoffs

src/com/xilinx/rapidwright/util/function/ThrowingConsumer.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ * 
+ */
+ 
+package com.xilinx.rapidwright.util.function;
+/**
+ * Like {@link java.util.function.Consumer}, but with checked exceptions
+ */
+@FunctionalInterface
+public interface ThrowingConsumer <T, E extends Throwable> {
```
eddieh-xlnx, line 28: Doesn't look like this interface is being used. Is it present for completeness?

src/com/xilinx/rapidwright/util/function/ThrowingFunction.java
```diff
@@ -0,0 +1,31 @@
...
+ * 
+ */
+ 
+package com.xilinx.rapidwright.util.function;
+
+@FunctionalInterface
+/**
+ * Like {@link java.util.function.Function}, but with checked exceptions
+ */
+public interface ThrowingFunction<T, R, E extends Throwable> {
```
eddieh-xlnx, line 29: Doesn't look like this interface is being used. Is it present for completeness?

eddieh-xlnx (approved):
I've merged in eddieh-xlnx/RapidWrightDCP#7 so please update the submodule.

src/com/xilinx/rapidwright/util/function/ThrowingConsumer.java
jakobwenzel: Correct, I added them for completeness.

However, I am now aware of the interfaces in `org.apache.commons.io.function`. I'm essentially duplicating them, so I might as well switch to those.

jakobwenzel:
Publishing Unit Tests results failed: https://github.com/Xilinx/RapidWright/runs/6154823166?check_suite_focus=true
It's interesting that this build is still green...

Seems that Github Packages are currently down: https://www.githubstatus.com/incidents/b40k7ckrs7sp

eddieh-xlnx:
> Publishing Unit Tests results failed: https://github.com/Xilinx/RapidWright/runs/6154823166?check_suite_focus=true It's interesting that this build is still green...

Strange that there were only 2 checks run on this PR -- which does not include the publishing test results job -- both of which passed and gives it the overall green light. Perhaps GH is broken. I'm going to try and trigger CI again.

jakobwenzel:
If you don't see a rerun button at https://github.com/Xilinx/RapidWright/runs/6154823166?check_suite_focus=true in the top right, maybe @clavin-xlnx is the only one who gets to rerun jobs?

eddieh-xlnx:
> If you don't see a rerun button at https://github.com/Xilinx/RapidWright/runs/6154823166?check_suite_focus=true in the top right, maybe @clavin-xlnx is the only one who gets to rerun jobs?

Got it, thanks! I can click it, but I was puzzled why it didn't appear as a (failing) check to begin with. It does appear now though, as a pass.

clavin-xlnx (approved):

PR #382, opened 2022-04-07
First cut at tracking logical cell changes

clavin-xlnx:
This adds a capability to the logical netlist to keep track of `EDIFCell` objects that are modified by monitoring basic APIs.  It does not consider cells that have their name changed or to switch to a different library as a trigger for tracking.

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFNet.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFNet.java
@@ -71,11 +71,25 @@ public class EDIFNet extends EDIFPropertyObject {
 	 */
 	public void addPortInst(EDIFPortInst portInst){
 		if(portInsts == null) portInsts = new EDIFPortInstList();
-		if(parentCell != null && portInst.getCellInst() == null) {
+		boolean isParentCellNonNull = parentCell != null;
+		EDIFCellInst inst = portInst.getCellInst();
+		if(isParentCellNonNull && inst == null) {
 			parentCell.addInternalPortMapEntry(portInst.getName(), this);
 		}
 		portInst.setParentNet(this);
+		if(isParentCellNonNull) {
+		    trackChanges(EDIFChangeType.PORT_INST_ADD, inst, portInst.getName());
+		}
 		portInsts.add(portInst);
+	}
+	
+	public void trackChanges(EDIFChangeType type, EDIFCellInst inst, String portInstName) {
+        EDIFNetlist netlist = parentCell.getNetlist();
+        if(netlist != null && netlist.isTrackingCellChanges()) {
+            String instName = inst == null ? null : inst.getName();
+            EDIFChangeNet change = new EDIFChangeNet(type, portInstName, getName(), instName);
+            netlist.addTrackingChange(parentCell, change);
+        }
 	}
 	
 	public EDIFPortInst createPortInst(EDIFPort port){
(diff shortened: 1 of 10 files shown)


eddieh-xlnx (approved):

src/com/xilinx/rapidwright/edif/EDIFNet.java
```diff
@@ -267,6 +281,9 @@ public EDIFPortInst removePortInst(String portInstName) {
 	 */
 	public EDIFPortInst removePortInst(EDIFCellInst inst, String portInstName){
         if (portInsts == null) return null;
+        if(parentCell != null) {
+            trackChanges(EDIFChangeType.PORT_INST_ADD, inst, portInstName);
```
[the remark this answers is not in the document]
eddieh-xlnx: (Not that I need this functionality currently)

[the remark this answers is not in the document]
clavin-xlnx: In theory, the portInstName would have the index in brackets, but I suppose that shouldn't be relied on.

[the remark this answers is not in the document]
eddieh-xlnx: Maybe just jot a comment there so that future historians won't suffer?

src/com/xilinx/rapidwright/edif/EDIFNet.java
```diff
@@ -267,6 +281,9 @@ public EDIFPortInst removePortInst(String portInstName) {
 	 */
 	public EDIFPortInst removePortInst(EDIFCellInst inst, String portInstName){
         if (portInsts == null) return null;
+        if(parentCell != null) {
+            trackChanges(EDIFChangeType.PORT_INST_REMOVE, inst, portInstName);
```
clavin-xlnx, line 285:
```suggestion
            // This does not explicitly track the port instance index, in most cases the name should be sufficient.
            trackChanges(EDIFChangeType.PORT_INST_REMOVE, inst, portInstName);
```

src/com/xilinx/rapidwright/edif/EDIFNet.java
```diff
@@ -71,13 +71,27 @@ protected EDIFNet(){
...
 		if(portInsts == null) portInsts = new EDIFPortInstList();
-		if(parentCell != null && portInst.getCellInst() == null) {
+		boolean isParentCellNonNull = parentCell != null;
+		EDIFCellInst inst = portInst.getCellInst();
+		if(isParentCellNonNull && inst == null) {
 			parentCell.addInternalPortMapEntry(portInst.getName(), this);
 		}
 		portInst.setParentNet(this);
+		if(isParentCellNonNull) {
+		    trackChanges(EDIFChangeType.PORT_INST_ADD, inst, portInst.getName());
```
clavin-xlnx, line 81:
```suggestion
		    // This does not explicitly track the port instance index, in most cases the name should be sufficient.
		    trackChanges(EDIFChangeType.PORT_INST_ADD, inst, portInst.getName());
```

src/com/xilinx/rapidwright/edif/EDIFNet.java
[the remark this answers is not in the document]
clavin-xlnx: Comments added.

src/com/xilinx/rapidwright/edif/EDIFNet.java
eddieh-xlnx: Actually, this isn't true for the ADD case. If we've added a port inst, then we can query its index (which is what I have been doing). Unless of course, that port inst is later REMOVE-d.

clavin-xlnx: If the only record used to track a design change is an `EDIFChangeNet` object, it is not tracked explicitly.

eddieh-xlnx: `EDIFChangeNet` tracks net and instance, and then from the inst I can find the port inst via its name, and thus find the index?

PR #385, opened 2022-04-13
Method to create a new netlist using existing EDIFCellInst as top cell

clavin-xlnx:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -65,6 +65,7 @@ import com.xilinx.rapidwright.edif.EDIFCellInst;
 import com.xilinx.rapidwright.edif.EDIFHierCellInst;
 import com.xilinx.rapidwright.edif.EDIFHierNet;
 import com.xilinx.rapidwright.edif.EDIFHierPortInst;
+import com.xilinx.rapidwright.edif.EDIFLibrary;
 import com.xilinx.rapidwright.edif.EDIFNet;
 import com.xilinx.rapidwright.edif.EDIFNetlist;
 import com.xilinx.rapidwright.edif.EDIFPort;
@@ -2152,6 +2153,17 @@ public class DesignTools {
 		HashSet<String> instsWithSeparator = new HashSet<>();
 		for(Entry<String,String> e : srcToDestInstNames.entrySet()) {
 			EDIFHierCellInst cellInst = src.getNetlist().getHierCellInstFromName(e.getKey());
+			if(e.getValue().length() == 0) {
+			    // If its the top cell, remove the top cell from destNetlist
+			    EDIFLibrary destLib = destNetlist.getLibrary(cellInst.getCellType().getLibrary().getName());
+			    if(destLib == null){
+			        destLib = destNetlist.getWorkLibrary();
+			    }
+			    EDIFCell existingCell = destLib.getCell(cellInst.getCellType().getLegalEDIFName());
+			    if(existingCell != null) {
+			        destLib.removeCell(existingCell);
+			    }
+			}
 			destNetlist.migrateCellAndSubCells(cellInst.getCellType());
 			EDIFHierCellInst bbInst = destNetlist.getHierCellInstFromName(e.getValue());
 			bbInst.getInst().setCellType(cellInst.getCellType());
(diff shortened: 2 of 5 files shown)


src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -906,9 +906,23 @@ public static EDIFNetlist createNewNetlist(String topName, boolean addINV){
...
+	 * @param cellInst The new top cell/top cell inst in the netlist.
+	 * @return The newly created netlist from the provided cell inst.
+	 */
+	public static EDIFNetlist createNewNetlist(EDIFCellInst cellInst){
+	    EDIFNetlist n = new EDIFNetlist(cellInst.getName());
+	    n.generateBuildComments();
+	    EDIFDesign eDesign = new EDIFDesign(cellInst.getName());
+	    n.setDesign(eDesign);
+	    eDesign.setTopCell(cellInst.getCellType());
+	    n.migrateCellAndSubCells(cellInst.getCellType());
```
eddieh-xlnx, line 923: If I'm reading the code for this function correctly, this is a destructive operation with respect to the input netlist, right? If so (and I'm okay with it being so) would a more apt description be `moveToNewNetlist` and the doc to be updated accordingly?

.github/workflows/build.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
```
eddieh-xlnx, line 14: Do we plan to remove windows permanently?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2152,6 +2153,17 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
 		HashSet<String> instsWithSeparator = new HashSet<>();
 		for(Entry<String,String> e : srcToDestInstNames.entrySet()) {
 			EDIFHierCellInst cellInst = src.getNetlist().getHierCellInstFromName(e.getKey());
+			if(e.getValue().length() == 0) {
+			    // If its the top cell, remove the top cell from destNetlist
```
eddieh-xlnx, line 2157: Similar to my other comment, is this a destructive operation on the source netlist?

eddieh-xlnx: Also, it looks like this method was changed but I'm not seeing the linkage with the rest of this PR?

.github/workflows/build.yml
clavin-xlnx: This was causing failures, possibly because it is tracking the latest windows.  I don't see much value in testing it as it causes false positives and I don't believe we've seen a windows-specific issue.  Java is pretty robust cross-platform.  The only thing might be testing the GUI aspects of RapidWright.

eddieh-xlnx: I have no issue disabling it.

src/com/xilinx/rapidwright/edif/EDIFTools.java
clavin-xlnx: This is probably easy enough to make it non-destructive, I will see if I can change it so that it leaves the original netlist intact.

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx: `copyImplementation()` is a destructive operation on the source design, although maybe it shouldn't be.  These changes aren't technically related to the ones below--but they were necessary in trying to solve the same issue.  Should I change the PR title or create a new PR?

eddieh-xlnx: I would prefer to put it into a different PR to keep this one clean.

clavin-xlnx: See #395

eddieh-xlnx: Looks like this one includes some of the changes in #395, but #395 also depends on this PR?
Does the whole of `DesignTools.java` need to be reverted in this PR and for this to be merged in first?

clavin-xlnx: I plan to merge in this order, bringing in the latest to each PR after each merge: #396, #385, and #395.

eddieh-xlnx: Roger that. Too many numbers!

src/com/xilinx/rapidwright/edif/EDIFTools.java
clavin-xlnx: I looked again at this code, I don't believe it is destructive to the source netlist.  Do you see anything that is?  

I can see sharing of `EDIFCell` objects between the original and new `Netlist` objects, but I don't see how the original netlist is changed.

eddieh-xlnx: I may be quite wrong, but `migrateCellAndSubCells` gave me the impression that it was a "move" operation and not a "copy", and so when I dove into that method I saw this:
https://github.com/Xilinx/RapidWright/blob/3ebcff1a4816e092e0e38658ebb8945c8f886c01/src/com/xilinx/rapidwright/edif/EDIFNetlist.java#L436-L458
which seems to be adding the same cell reference into both netlists (is this wise?) and the cell's instances are then modified as more migration recursively happens.

clavin-xlnx: Yes, that seems right.  I wasn't thinking hard enough about it.  This method has always been used such that the source design was not needed for anything after, so it hasn't caused an issue.  I'll add some documentation that denotes this and we can create a non-destructive version whenever that is needed.

eddieh-xlnx (approved):

PR #390, opened 2022-04-18
Application code to create hardware contract  (depend on PR #435,436)

pongstorn-xilinx:
This method is called from copyImplementation to copy static net.   Two flags are added to the original copyImplementation. One is to control whether to copy static nets and the other to indicate copying only internal nets.


clavin-xlnx:
@pongstorn-xilinx This looks to be the same set of changes as #385.  Did some of your changes get lost?

pongstorn-xilinx:
Chris,  you are right. I forgot to commit the change. I will update it.

pongstorn-xilinx:
I have updated this PR. It contains PR #395, #385.  This is probably the reason for conflicts.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2261,7 +2309,87 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
+				net.addPIP(p);
+			}
+		}
+	}
+	
+	private static String getNewHierName(String srcName, Map<String,String> srcToDestInstNames, 
+	                                        Map<String,String> prefixes, String prefixMatch) {
+        String newCellPrefix = srcToDestInstNames.get(prefixes.get(prefixMatch));
+        int idx = prefixMatch.length() - (newCellPrefix.length() == 0 ? 0 : 1);
+        return newCellPrefix + srcName.substring(idx);
```
clavin-xlnx, line 2392:
```suggestion
```

clavin-xlnx: Removing due to duplication during merge.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2144,6 +2144,31 @@ public static Map<String,String> getInvertiblePinMap(Series series, Unisim unisi
...
+	 * @param src The source design with or without implementation
+	 * @param cellName The full hierarchy cell name to be extracted as another design
+	 * @return A newly created design copied from the source
+	 */
+	public static Design createDesignFromCellWithStatic(Design src, String cellName) {
+		return createDesignFromCell(src, cellName, true);
+	}
+	public static Design createDesignFromCellWithoutStatic(Design src, String cellName) {
+		return createDesignFromCell(src, cellName, false);
+	}
```
clavin-xlnx, line 2158: These two APIs seem a bit redundant.  If you want to overload `createDesignFromCell()` to include static routing by default without having to pass a parameter, I think that would be a better approach.  If someone doesn't want to include it, then they can use the API where they pass in a param.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2144,6 +2144,31 @@ public static Map<String,String> getInvertiblePinMap(Series series, Unisim unisi
...
+	public static Design createDesignFromCell(Design src, String cellName, boolean copyPIPOfStaticNet) {
```
clavin-xlnx, line 2159: Instead of `copyPIPOfStaticNet`, how about `keepStaticRouting`?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2157,7 +2182,25 @@ public static Map<String,String> getInvertiblePinMap(Series series, Unisim unisi
...
+	 * @param copyStaticNet Flag indicating if static nets should be copied
+	 * @param copyOnlyInternalNets Flag indicating if only nets with every terminal inside the cell should be copied
+	 * @param lockPlacement Flag indicating if the destination implementation copy should have the
+	 * 	placement locked
+	 * @param lockRouting Flag indicating if the destination implementation copy should have the
+	 * 	routing locked
+	 * @param srcToDestInstNames A map of source (key) to destination (value) pairs of cell
+	 * instances from which to copy the implementation
+	 */
+	public static void copyImplementation(Design src, Design dest, boolean copyStaticNet, boolean copyOnlyInternalNets, boolean lockPlacement,
```
clavin-xlnx, line 2203: `copyStaticNets`?  There are two.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2157,7 +2182,25 @@ public static Map<String,String> getInvertiblePinMap(Series series, Unisim unisi
...
+	public static void copyImplementation(Design src, Design dest, boolean lockPlacement,
+										  boolean lockRouting, Map<String,String> srcToDestInstNames) {
+		copyImplementation(src, dest, false, false, lockPlacement, lockRouting, srcToDestInstNames);
+	}
+
+	/**
+	 * Copies the logic and implementation of a set of cells from one design to another with additional flags to control copying nets.
+	 * @param src The source design (with partial or full implementation)
+	 * @param dest The destination design (with matching cell instance interfaces)
+	 * @param copyStaticNet Flag indicating if static nets should be copied
```
clavin-xlnx, line 2194: It only copies the portion of static nets inside the designated part, correct?

.gitignore
```diff
@@ -10,3 +10,11 @@
...
 /designs/
 interchange/schema/capnp/java.capnp
+*.bit
+vivado*
+*.dcp
+*.edf
+*.edif
+.*
+*.log
+*.bak
```
clavin-xlnx, line 20: Could we hold off on these changes for now?  Perhaps in a different PR?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2261,7 +2315,98 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
+		}
+
+		if (copyStaticNet) {
+			copyStaticNets(dest, staticNets, siteInstsOfCells);
+		}
+	}
+
+	private static boolean equalNodes(Node lNode, Node rNode) {
+		return (lNode.getTile().getName().equals(rNode.getTile().getName()) && lNode.getWireName().equals(rNode.getWireName()));
+	}
```
clavin-xlnx, line 2327: Why not use `Node.equals()`?  This seems like it would be slower/redundant.

src/com/xilinx/rapidwright/util/MakeHWCTFromImpl.java
```diff
@@ -0,0 +1,191 @@
+/*
```
clavin-xlnx, line 1: This class seems specific for a particular application.  Could this be distributed outside of RapidWright?

src/com/xilinx/rapidwright/util/RelocateModulesIntoBlackboxes.java
```diff
@@ -61,9 +60,17 @@ public class RelocateModulesIntoBlackboxes {
 	 * @param blackboxes The list of pairs of a black box cell to be filled and its reference INT tile.
 	 *                    The x-coordinate of this INT tile must match that of the cellAnchor.
 	 */
-	public static boolean relocateModuleInsts(Design top, Module mod, String cellAnchor, List<Pair<String, String>> blackboxes) {
+	public static boolean relocateModuleInsts(Design top, Module mod, String cellAnchor, ArrayList<Pair<String, String>> blackboxes) {
```
clavin-xlnx, line 63: Is there an advantage to making this an `ArrayList`?

eddieh-xlnx:
This PR has morphed beyond "Copying static net while copyImplementation" into one that also does HWCT. Can we separate the two again please?

.gitignore
pongstorn-xilinx: I believe .gitignore should never been merged. However, each person will have a different .gitignore, we need to find a way to ignore the .gitignore.    I can revert to the version from master.   I hope there is a way to ignore the file during PR merging.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: Remove both and call createDesignFromCell with flag instead.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: changed.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: yes, it traces only from the static sink within the target cell.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: change to copyStaticNets

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: Node.equals() uses wire != other.wire.  Is that  object address comparison?

At least I can use Tile.equals().

src/com/xilinx/rapidwright/util/RelocateModulesIntoBlackboxes.java
pongstorn-xilinx: Changed to List.
Previously, an operation, was outside this method, on the list uses indexing. The op was moved into the method, thus ArrayList was used to make clear of the requirement. Now, the op is changed to not rely on the indexing.

src/com/xilinx/rapidwright/util/MakeHWCTFromImpl.java
pongstorn-xilinx: This will be used with PYNQ (I want to update OpenDFX to use this as well). I were thinking that if this is a part of RW, downloading rw_api jar and just call it. If this is distributed with PYNQ, does it means PYNQ have to has a full fork of RW? Once we are closing on the option, we need feedback from the PYNQ team.

Eddie and I agreed on that we will splits this PR after we clean up all comments.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: Changed to Node.equals().

src/com/xilinx/rapidwright/util/MakeHardwareContract.java
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        return true;
+    }
+
+    public static void main(String[] args) {
+        String usage = String.join(System.getProperty("line.separator"),
+                "Make a implemented Hardware Contract (HWCT) by copying an implementation into a target design.",
+                "  -in       <name of the routed DCP to make HWCT of>",
+                "  -formCell <name of the cell in the DCP specified by -in to copy the implementation from>",
```
eddieh-xlnx, line 103:
```suggestion
                "  -fromCell <name of the cell in the DCP specified by -in to copy the implementation from>",

```

Issue #389, opened 2022-04-18
copyImplementation does not remove route to an external sink


pongstorn-xilinx:
There are some nets whose their drivers are in the cell to be copied and some sinks outside the cell. The route to outside sinks should be removed, but they are not.

https://github.com/Xilinx/RapidWright/blob/a34c0d71521a758e757056714ff0503de28f0c71/src/com/xilinx/rapidwright/design/DesignTools.java#L2216-L2221

pongstorn-xilinx:
The issue was resolved with PR #395.

PR #395, opened 2022-04-19
Fixes copyImplementation() partial net routing 

clavin-xlnx:
`DesignTools.copyImplementation()` can be a destructive behavior to the source input design.  This change removes the destructive behavior on the source side for net routing and also fixes an issue that was leaving extra routing behind in the copy.  This behavior was manifest in Issue #389.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


clavin-xlnx:
This includes/depends on #385.

eddieh-xlnx (approved):

PR #397, opened 2022-04-20
check open files in all testcases, not just explicitly annotated ones

jakobwenzel:
This applies checking for open files to all testcases, removing the need for the annotation.

When building with Eclipse, it will probably not pick up the two changes I did in Gradle:
* Add `testResources` as Resource Path
* When running tests, set the system property `junit.jupiter.extensions.autodetection.enabled` to `true`.


build.gradle
```diff
@@ -29,6 +29,9 @@ sourceSets {
     java {
       srcDirs = ['test']
     }
+    resources {
+      srcDirs = ['testResources']
```
clavin-xlnx, line 33: Could we make it `test/resources` instead of `testResources`?  Just want to minimize the number of folders at the root.  If we need to (not sure how Gradle handles it), we can change `srcDirs = ['test/src']`.

jakobwenzel: So we would move all test classes down one level into the new src directory?

clavin-xlnx (approved):

PR #403, opened 2022-04-29
DesignTools methods to use Net.createPin()

eddieh-xlnx:
For:
* `DesignTools.createMissingSitePinInsts()`
* `DesignTools.getRoutedSitePinFromPhysicalPin()`

Instead of constructing a new `SitePinInst`, use `Net.createPin()` which is (will be) responsible for adding output pins (main and alternate ones) to the list returned by `Net.getPins()`.

New test added to check that this is the case.


clavin-xlnx (approved):
Looks fine by me!

PR #406, opened 2022-05-03
Fix netlist issue when ordering cells for export

clavin-xlnx:
During what seems like some rare circumstances, the ordering of cells on export can mistakenly find a false dependency if the library is not checked.  This ensures the library is checked as part of the new ordering so that identically named cells aren't mistaken.

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFLibrary.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFLibrary.java
@@ -247,7 +247,7 @@ public class EDIFLibrary extends EDIFName {
 		yetToVisit.remove(cell.getLegalEDIFName());
 		for(EDIFCellInst i : cell.getCellInsts()){
 			EDIFCell childCell = i.getCellType();
-			if(containsCell(childCell) && yetToVisit.containsKey(childCell.getLegalEDIFName())){
+			if(childCell.getLibrary() == this && containsCell(childCell) && yetToVisit.containsKey(childCell.getLegalEDIFName())){
 				visit(childCell,visited,yetToVisit);
 			}
 		}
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/edif/EDIFLibrary.java
```diff
@@ -247,7 +247,7 @@ private void visit(EDIFCell cell, List<EDIFCell> visited, Map<String,EDIFCell> y
 		yetToVisit.remove(cell.getLegalEDIFName());
 		for(EDIFCellInst i : cell.getCellInsts()){
 			EDIFCell childCell = i.getCellType();
-			if(containsCell(childCell) && yetToVisit.containsKey(childCell.getLegalEDIFName())){
+			if(childCell.getLibrary() == this && containsCell(childCell) && yetToVisit.containsKey(childCell.getLegalEDIFName())){
```
eddieh-xlnx, line 250: Is the `containsCell()` condition still necessary?

`containsCell()` eventually ends up here:
https://github.com/Xilinx/RapidWright/blob/96a79eb27b8f98d07ca3d8aa7726f80ca9153aaf/src/com/xilinx/rapidwright/edif/EDIFLibrary.java#L148-L149

thus the new condition `childCell.getLibrary() == this` catches the situation where a cell is attached to a library, but does not belong in that library's `cells` map?

eddieh-xlnx: (I'm probing to see if there's a good reason for that latter case to appear to understand if we're just papering over the symptom and not the root cause)

clavin-xlnx: I think you are right, I don't know that `containsCell()` is still necessary.  The issue is that it was only checking by name if the cell was in the library.  Identically named cells from different libraries were causing an issue.  I will remove the second clause of the `if` statement.

eddieh-xlnx: Ah, got it! The problem isn't that a cell wasn't present in the library it says it belongs to, but with identically named cells in two different libraries. As you were.

eddieh-xlnx (approved):

PR #409, opened 2022-05-06
Replace hardcoded START_TILE_COL/ROW from the file by computing if needed.

pongstorn-xilinx:


eddieh-xlnx:
Can we isolate this PR to the minimum (just [6e239fe](https://github.com/Xilinx/RapidWright/pull/409/commits/6e239fe90dd04ef8e44194797f891b6120919693)?) and create a test please?

The commit description states:
> test with xzcu7, xcvu3p, kv260
Is the timing model relevant for non US+ devices?
In my particular case, I found that `intersite_delay_terms.txt` caused the timing model to break for a vu19p.

pongstorn-xilinx:
There are only two actual file changes for this PR (TimingModel and intersite_delay_term). However, it branched off from the code with build_hwct. I will find a way to separate the two files out.

eddieh-xlnx:
> > test with xzcu7, xcvu3p, kv260
> 
> Is the timing model relevant for non US+ devices? In my particular case, I found that `intersite_delay_terms.txt` caused the timing model to break for a vu19p.

Sorry, I've since come to understand that these are all US+ devices. 

We should also have a test for those 3 devices plus the vu19p.

The failure I'm seeing with the vu19p comes from RWRoute:
```
Exception in thread "main" java.lang.RuntimeException: Invalid values of START_TILE_COL and START_TILE_ROW in timing/ultrascaleplus/intersite_delay_terms.txt. They must specify an interconnect tile with SLICE on both sides.
	at com.xilinx.rapidwright.timing.delayestimator.DelayEstimatorBase.loadInputSitePinDelay(DelayEstimatorBase.java:284)
	at com.xilinx.rapidwright.timing.delayestimator.DelayEstimatorBase.<init>(DelayEstimatorBase.java:88)
<<SNIP>>
```

which should be originating from here: 

https://github.com/Xilinx/RapidWright/blob/70ae9d339ccfeb34c8a294a27b62934f3bc2bbf8/src/com/xilinx/rapidwright/rwroute/RWRoute.java#L188

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -500,17 +491,93 @@ protected List<TimingGroup> determineGroups(List<Node> nodes, List<IntentCode> n
...
+        for (int j = 0; j < d.getRows(); j++) {
+            int span = 0;
+            for (int i = 0; i < d.getColumns(); i++) {
+                if (d.getTile("INT_X" + i + "Y" + j) != null) {
+                    span++;
+                }
+            }
+            rowHelper.insert(j,span);
+        }
+
```
eddieh-xlnx, line 537:
```suggestion

        // Find the tile with the most number of INT tiles on the same row/col
```

What's the significance of most number of INT tiles? Would some upper limit (e.g. equivalent to the longest wire) be sufficient?

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -500,17 +491,93 @@ protected List<TimingGroup> determineGroups(List<Node> nodes, List<IntentCode> n
...
+        for (int INTCol : colHelper.longest()) {
+            for (int INTRow : rowHelper.longest()) {
+                Tile tile = d.getTile("INT_X" + INTCol + "Y" + INTRow);
+                int col = tile.getColumn();
+                int row = tile .getRow();
+                // Want an INT tile that has CLB on both side
+                if (isCLE(d.getTile(row, col-1)) && isCLE(d.getTile(row, col+1))) {
```
eddieh-xlnx, line 544:
```suggestion
                if (Utils.isCLB(d.getTile(row, col-1)) && Utils.isCLB(d.getTile(row, col+1))) {
```

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -500,17 +491,93 @@ protected List<TimingGroup> determineGroups(List<Node> nodes, List<IntentCode> n
...
+        return (
+                t == TileTypeEnum.CLE_M
+                || t == TileTypeEnum.CLE_M_R
+                || t == TileTypeEnum.CLEL_L
+                || t == TileTypeEnum.CLEL_R
+                || t == TileTypeEnum.CLEM
+                || t == TileTypeEnum.CLEM_R
+        );
+    }
+
```
eddieh-xlnx, line 563:
```suggestion
```

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -614,6 +681,8 @@ protected boolean readDelayTerms(String filename) {
             e.printStackTrace();
             result = false;
         }
+
+
```
eddieh-xlnx, line 685:
```suggestion
```

src/com/xilinx/rapidwright/timing/TimingModel.java
pongstorn-xilinx: The model needs to compute "d" which is a function of tile types. Two long wires may span different set of tiles. Thus, need to scan the whole device to build distance array for both vertical and horizontal. Thus, one full row (column) is required.

src/com/xilinx/rapidwright/timing/TimingModel.java
pongstorn-xilinx: done

src/com/xilinx/rapidwright/timing/TimingModel.java
pongstorn-xilinx: done

src/com/xilinx/rapidwright/timing/TimingModel.java
pongstorn-xilinx: remove one

test/src/com/xilinx/rapidwright/util/TestTimingModel.java
```diff
@@ -0,0 +1,46 @@
+package com.xilinx.rapidwright.util;
```
eddieh-xlnx, line 1: Can you place this file into
`test/src/com/xilinx/rapidwright/timing`
to mirror the location of the class you're trying to test please?

test/src/com/xilinx/rapidwright/util/TestTimingModel.java
```diff
@@ -0,0 +1,33 @@
...
+        TimingModel model = new TimingModel(Device.getDevice(deviceName));
+        model.build();
+        Tile tile = model.getRefIntTile();
+        // check tile coor that can be visually checked in Vivado gui
+        Assertions.assertEquals(expectedTileX,tile.getTileXCoordinate());
+        Assertions.assertEquals(expectedTileY,tile.getTileYCoordinate());
+        // check tile consistent with what specify in the file
+        Assertions.assertEquals(expectedCol,tile.getColumn());
+        Assertions.assertEquals(expectedRow,tile.getRow());
+    }
```
eddieh-xlnx, line 32: Can we also add a check that the reference INT tile does indeed have CLBs on both sides?

test/src/com/xilinx/rapidwright/util/TestTimingModel.java
pongstorn-xilinx: My bad, I meant to do that.

test/src/com/xilinx/rapidwright/util/TestTimingModel.java
pongstorn-xilinx: I were thinking about that. But, if we supply the target tiles, we need to check that already.  Thus, I decided to keep the test as simple as possible.

pongstorn-xilinx: Furthermore, the test will use Utils.isCLB which is the same as that at the end of the method (L555). Thus, it does not provide much benefit.

eddieh-xlnx (approved):

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -500,17 +502,82 @@ protected List<TimingGroup> determineGroups(List<Node> nodes, List<IntentCode> n
         return result;
     }
 
+    private Pair<Integer,Integer> findReferenceTileLocation(Device d) {
```
clavin-xlnx, line 505: The instance of this class has a reference to a `Device` object.  Should this method be static or can we remove parameter?

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -500,17 +502,82 @@ protected List<TimingGroup> determineGroups(List<Node> nodes, List<IntentCode> n
         return result;
     }
 
+    private Pair<Integer,Integer> findReferenceTileLocation(Device d) {
+        class Helper {
```
clavin-xlnx, line 506: Perhaps consider a more descriptive name than `Helper`?  

The Guava library has a map that mostly does what `Helper` does:
```java
        SetMultimap<Integer, Integer> map = TreeMultimap.create();
        map.put(1, 2);
        map.put(1, 1);
        map.put(1, 1);
        System.out.println(map.get(1));
        // output is [1,2]
```
RapidWright already includes Guava, so it would be a (mostly) drop in replacement.  Ultimately, it would be nice to not have an inner class.

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -500,17 +502,82 @@ protected List<TimingGroup> determineGroups(List<Node> nodes, List<IntentCode> n
...
+                if (d.getTile("INT_X" + i + "Y" + j) != null) {
+                    span++;
+                }
+            }
+            rowHelper.insert(j,span);
+        }
+
+        for (int INTCol : colHelper.longest()) {
+            for (int INTRow : rowHelper.longest()) {
+                Tile tile = d.getTile("INT_X" + INTCol + "Y" + INTRow);
```
clavin-xlnx, line 551:
```suggestion
        for (int x : colHelper.longest()) {
            for (int y : rowHelper.longest()) {
                Tile tile = d.getTile("INT", x, y);

```
Variables should not start with capital letters and replacing tile getter with one that doesn't need to create String objects.

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -500,17 +502,82 @@ protected List<TimingGroup> determineGroups(List<Node> nodes, List<IntentCode> n
...
+                int col = tile.getColumn();
+                int row = tile .getRow();
+                // Want an INT tile that has CLB on both side
+                if (Utils.isCLB(d.getTile(row, col-1).getTileTypeEnum()) && Utils.isCLB(d.getTile(row, col+1).getTileTypeEnum())) {
+                    return new Pair<>(col,row);
```
clavin-xlnx, line 556: Could this method be changed to return the `Tile` instead of creating a `Pair` object?  In the case a location is not found, it could just return `null`.

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -500,17 +502,82 @@ protected List<TimingGroup> determineGroups(List<Node> nodes, List<IntentCode> n
...
+        // for each column, look for valid row
+        Helper colHelper = new Helper();
+        for (int i = 0; i < d.getColumns(); i++) {
+            int span = 0;
+            for (int j = 0; j < d.getRows(); j++) {
+                if (d.getTile("INT_X" + i + "Y" + j) != null) {
+                    span++;
+                }
+            }
+            colHelper.insert(i,span);
```
clavin-xlnx, line 534:
```suggestion
        for (int x = 0; x < d.getColumns(); x++) {
            int span = 0;
            for (int y = 0; y < d.getRows(); y++) {
                if (d.getTile("INT", x, y) != null) {
                    span++;
                }
            }
            colHelper.insert(x,span);

```
Refactor iterator names to match x,y.  Using more efficient tile getter.

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -500,17 +502,82 @@ protected List<TimingGroup> determineGroups(List<Node> nodes, List<IntentCode> n
...
+        Helper rowHelper = new Helper();
+        for (int j = 0; j < d.getRows(); j++) {
+            int span = 0;
+            for (int i = 0; i < d.getColumns(); i++) {
+                if (d.getTile("INT_X" + i + "Y" + j) != null) {
+                    span++;
+                }
+            }
+            rowHelper.insert(j,span);
+        }
```
clavin-xlnx, line 547:
```suggestion
        for (int y = 0; y < d.getRows(); y++) {
            int span = 0;
            for (int x = 0; x < d.getColumns(); x++) {
                if (d.getTile("INT", x, y) != null) {
                    span++;
                }
            }
            rowHelper.insert(y,span);
        }

```

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -500,17 +502,82 @@ protected List<TimingGroup> determineGroups(List<Node> nodes, List<IntentCode> n
...
+
+        for (int INTCol : colHelper.longest()) {
+            for (int INTRow : rowHelper.longest()) {
+                Tile tile = d.getTile("INT_X" + INTCol + "Y" + INTRow);
+                int col = tile.getColumn();
+                int row = tile .getRow();
```
clavin-xlnx, line 553:
```suggestion
                int row = tile.getRow();

```

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -500,17 +502,82 @@ protected List<TimingGroup> determineGroups(List<Node> nodes, List<IntentCode> n
...
      * @return Boolean indication of completion.
      */
     protected boolean readDelayTerms(String filename) {
-        boolean result = true; 
-        try (BufferedReader br = new BufferedReader(new FileReader(FileTools.getRapidWrightPath() + File.separator + filename))){
+
+        // Compute before reading from file to allow overriding.
+        Pair<Integer,Integer> loc = findReferenceTileLocation(device);
+        START_TILE_COL = loc.getFirst();
+        START_TILE_ROW = loc.getSecond();
```
clavin-xlnx, line 574: These variable names are neither static nor final.  They should probably use a lowerCamelCase naming.

clavin-xlnx (changes requested):

src/com/xilinx/rapidwright/timing/TimingModel.java
pongstorn-xilinx: These are just two of like 20 legacy variable names.  If we change these two, we have to change them all.  Are you OK with that?

pongstorn-xilinx: Actually, there are like 80 of them!

src/com/xilinx/rapidwright/timing/TimingModel.java
pongstorn-xilinx: Will remove the parameter and use from the class member.

src/com/xilinx/rapidwright/timing/TimingModel.java
pongstorn-xilinx: Will do and change the method name to findReferenceTile().

src/com/xilinx/rapidwright/timing/TimingModel.java
clavin-xlnx: Let's defer then to another PR for now.

clavin-xlnx (approved):

PR #412, opened 2022-05-10
Fix thread pool for single processor case

clavin-xlnx:
When running in the `mybinder.org` Docker, only a single processor core is allocated to run the Jupyter Notebook.  The parameters to `ThreadPoolExecutor` were not legal and was throwing an `IllegalArgumentException`:

```
jshell> import com.xilinx.rapidwright.util.ParallelismTools;

jshell> ParallelismTools.get
get(            getParallel()
jshell> ParallelismTools.getParallel()
|  Exception java.lang.ExceptionInInitializerError
|        at (#2:1)
|  Caused by: java.lang.IllegalArgumentException
|        at ThreadPoolExecutor.<init> (ThreadPoolExecutor.java:1293)
|        at ThreadPoolExecutor.<init> (ThreadPoolExecutor.java:1215)
|        at ParallelismTools.<clinit> (ParallelismTools.java:67)
|        ...
```

This fixes the issue.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
+	jvm_opt: [, -Dorg.gradle.jvmargs=-XX:ActiveProcessorCount=1]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
@@ -46,7 +47,7 @@ jobs:
         run: ./gradlew assemble
 
       - name: Test
-        run: ./gradlew test
+        run: ./gradlew test ${{ matrix.jvm_opt }}
 
       - name: Archive Java Test Report
         uses: actions/upload-artifact@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-	jvm_opt: [, -Dorg.gradle.jvmargs=-XX:ActiveProcessorCount=1]
+        single_threaded: [0, 1]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
@@ -47,7 +47,7 @@ jobs:
         run: ./gradlew assemble
 
       - name: Test
-        run: ./gradlew test ${{ matrix.jvm_opt }}
+        run: ./gradlew -PlimitToSingleProc=${{ matrix.single_threaded }} test 
 
       - name: Archive Java Test Report
         uses: actions/upload-artifact@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        single_threaded: [0, 1]
+        single_threaded: [multi-threaded, single-threaded]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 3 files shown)


eddieh-xlnx (commented):
Perhaps we can disable the `parallel` boolean from being set when `maxParallelism() == 1` (which would mean everything gets executed on the submitting thread), which should also mean we can initialize `pool` to `null`?

clavin-xlnx:
> Perhaps we can disable the `parallel` boolean from being set when `maxParallelism() == 1` (which would mean everything gets executed on the submitting thread), which should also mean we can initialize `pool` to `null`?

Is there any value in being able to switch between using parallel and single-threaded execution during the same run?  I'm wondering if we want to have that flexibility, at least for debugging purposes.

clavin-xlnx:
> > Is there any value in being able to switch between using parallel and single-threaded execution during the same run? I'm wondering if we want to have that flexibility, at least for debugging purposes.
> 
> Yes, there's value in that for machines with more than one core. If you're on a single core machine (as determined by `maxParallelism()`) then perhaps `parallel` should be locked to `false` so that the submitting thread executes the job immediately rather than putting it into a queue to execute later (and also no thread pool needs to be created).

If we keep the pool `null`, it seems like it opens the potential for several NPE bugs to crop up.  I am assuming that a running in a single-core environment is somewhat rare these days.  Thoughts on proceeding down this path?

eddieh-xlnx:
> If we keep the pool `null`, it seems like it opens the potential for several NPE bugs to crop up. I am assuming that a running in a single-core environment is somewhat rare these days. Thoughts on proceeding down this path?

I personally prefer a NPE to crop up. It means we/someone is using the thread pool even when there really should be no reason to. But that does assume we test our code sufficiently to catch it before it goes out into the field.

clavin-xlnx:
> > If we keep the pool `null`, it seems like it opens the potential for several NPE bugs to crop up. I am assuming that a running in a single-core environment is somewhat rare these days. Thoughts on proceeding down this path?
> 
> I personally prefer a NPE to crop up. It means we/someone is using the thread pool even when there really should be no reason to. But that does assume we test our code sufficiently to catch it before it goes out into the field.

Would you mind taking a look?  I don't feel confident in my understanding of how this works to code up the vision here.

eddieh-xlnx:
@clavin-xlnx Here's my stab at it. We don't have any testing for the single core case though. Should we add a new build that either forces `ParallelismTools.maxParallelism()` to return `1`, or sets `RW_PARALLEL=0`?

clavin-xlnx:
> @clavin-xlnx Here's my stab at it. We don't have any testing for the single core case though. Should we add a new build that either forces `ParallelismTools.maxParallelism()` to return `1`, or sets `RW_PARALLEL=0`?

> There are cases where the algorithm used is different depending on `ParallelismTools.getParallel()` -- such as in the EDIF parser and in the DCP writer.

Perhaps we should.  The number of processors available can be controlled with the JVM arg `-XX:ActiveProfcessorCount=1`.  

However, I'm concerned that this change is going to require some re-writes in a few places.  My principle concern is around that way the `submit()` methods access the `pool` object unprotected.

eddieh-xlnx:
> However, I'm concerned that this change is going to require some re-writes in a few places. My principle concern is around that way the `submit()` methods access the `pool` object unprotected.

My observation is that accesses to `pool` should be protected by a `getParallel()` call. The invariant I am trying to apply is that `getParallel()` should never return true if `maxParallelism() == 1`.

eddieh-xlnx (approved):

PR #417, opened 2022-05-12
Add EDIFNetlist.copyCellsAndSubCells()

eddieh-xlnx:


src/com/xilinx/rapidwright/edif/EDIFNetlist.java
```diff
@@ -537,6 +537,38 @@ public void migrateCellAndSubCells(EDIFCell cell, boolean uniqueifyCollisions){
...
+	 * @param cell The cell (and all its descendants) to copy into this netlist's libraries
+	 */
+	public void copyCellAndSubCells(EDIFCell cell) {
+		copyCellAndSubCellsWorker(cell);
+	}
+
+	private EDIFCell copyCellAndSubCellsWorker(EDIFCell cell) {
+		EDIFLibrary destLib = getLibrary(cell.getLibrary().getName());
+		if(destLib == null){
+			if(cell.getLibrary().getName().equals(EDIFTools.EDIF_LIBRARY_HDI_PRIMITIVES_NAME)){
```
clavin-xlnx, line 552:
```suggestion
			if(cell.getLibrary().isHDIPrimitivesLibrary()){
```

test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
```diff
@@ -100,4 +100,35 @@ public void testTrackChanges() {
             Assertions.assertTrue(potentiallyModifiedCells.contains(modifiedCell));
         }
     }
+
+    @Test
+    public void testCopyCellsAndSubCells() {
+        String dcpPath = RapidWrightDCP.getString("picoblaze_ooc_X10Y235.dcp");
+        Design design = RapidWrightDCP.loadDCP(dcpPath);
```
clavin-xlnx, line 107:
```suggestion
        Design design = RapidWrightDCP.loadDCP("picoblaze_ooc_X10Y235.dcp");
```

src/com/xilinx/rapidwright/edif/EDIFNetlist.java
```diff
@@ -537,6 +537,38 @@ public void migrateCellAndSubCells(EDIFCell cell, boolean uniqueifyCollisions){
...
+		if(existingCell == null){
+			EDIFCell newCell = new EDIFCell(destLib, cell, cell.getName());
+			for(EDIFCellInst inst : newCell.getCellInsts()){
+				inst.setCellType(copyCellAndSubCellsWorker(inst.getCellType()));
+				//The view might have changed
+				inst.getViewref().setName(inst.getCellType().getView());
+			}
+			return newCell;
+		} else {
+			return existingCell;
```
clavin-xlnx, line 569: Overall, I agree with this solution.  However, if a cell already exists in the destination library that is not the primitives library, what should we do with name collisions of cells?  Sometimes, the cells might indeed be identical (should we perform a check to ensure they are?), however, they might also not be (should we unique-ify?).  Should we have a pre-emptive check before calling `copyCellAndSubCells()`?

eddieh-xlnx: You're right. Current behaviour is based on `migrateCellAndSubCells()` which assumes that if a source cell exists in the destination netlist with the same name and in the same library, then it must be the same cell. That is not of course not always true.

I think the perfect thing to do would be to uniquify, but then you would be losing a 1:1 mapping with the cell names present in the original `Module` and it's not clear what effects this may have with all the other module code. Hier paths should not be affected though, even if the cells are renamed.
The `Design` would also need to keep track of which (possibly renamed) cell that has been copied into its internal netlist is the top cell for each `Module` (the alternative is to disallow top cells from being rename-able).

There does exist a `migrateCellAndSubCells(EDIFCell,boolean)` overload which seems to pessimistically assumes that cell name collisions are always of two different cells, and thus renames before migrating.

eddieh-xlnx: Solution: upon encountering an collision, throw an exception and abort.

eddieh-xlnx: Resolved in 1415591.

clavin-xlnx (approved):

Issue #419, opened 2022-05-13
[Versal] [placeCell] Site type DSP58_PRIMARY not supported for cell type DSP_XXX


nqdtan:
Hi,

I'm encountering placement issue related to the Versal DSP BELs. Please find the code for reproducing the issue as follows. (also the DCP and EDIF files in the attached zip file.)

```java
package com.xilinx.rapidwright.examples;
import java.io.FileNotFoundException;

import com.xilinx.rapidwright.design.*;
import com.xilinx.rapidwright.device.*;

public class TestPlaceDSPVersal {
  public static void main(String[] args) throws FileNotFoundException {
    Design design1 = Design.readCheckpoint("add_placed.dcp");
    Design design2 = new Design("new_design", design1.getDevice().getName());
    for (Cell c : design1.getCells()) {
      design2.copyCell(c, c.getName());
    }
  }
}
```
**RapidWright version:** 2022.1 (beta release; it seems that the code at master somehow crashes when reading my DCP.)

**Vivado version (for the DCP):** 2022.1

**Error log**

> Exception in thread "main" java.lang.RuntimeException: ERROR: Site type DSP58_PRIMARY not supported for cell type DSP_PREADD58
	at com.xilinx.rapidwright.design.i.a(Unknown Source)
	at com.xilinx.rapidwright.design.i.a(Unknown Source)
	at com.xilinx.rapidwright.design.i.a(Unknown Source)
	at com.xilinx.rapidwright.design.Design.placeCell(Unknown Source)
	at com.xilinx.rapidwright.design.Design.copyCell(Unknown Source)
	at com.xilinx.rapidwright.examples.TestPlaceDSPVersal.main(TestPlaceDSPVersal.java:15)

The DCP only contains a single cell that does 32b addition. I think there are also other DSP cells that are currently not supported as well: DSP_ALUADD, DSP_ALUMUX, DSP_ALUREG, DSP_A_B_DATA58, DSP_C_DATA58, DSP_MULTIPLIER58, DSP_M_DATA58, DSP_OUTPUT58, DSP_PATDET, DSP_PREADD58, DSP_PREADD_DATA58, DSP_SRCMX_OPTINV.

Thanks!
[src.zip](https://github.com/Xilinx/RapidWright/files/8683972/src.zip)

clavin-xlnx:
Thanks for the issue and example code/design, always helpful to have those!  

[...]

PR #421, opened 2022-05-13
Test for copyCell() fix from #419

clavin-xlnx:
Test that ensures `Design.copyCell()` identifies proper alternate site type for the destination `SiteInst` if it does not already exist.

### DIFF
--- a/test/src/com/xilinx/rapidwright/design/TestDesign.java
+++ b/test/src/com/xilinx/rapidwright/design/TestDesign.java
@@ -241,4 +241,12 @@ public class TestDesign {
         queue.addJob(job);
         Assertions.assertTrue(queue.runAllToCompletion());
     }
+    
+    @Test
+    public void testCopyCell() {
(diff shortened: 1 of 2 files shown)


eddieh-xlnx (approved):

test/src/com/xilinx/rapidwright/design/TestDesign.java
```diff
@@ -237,4 +237,12 @@ public void testBug349(@TempDir Path tempDir) throws IOException {
         queue.addJob(job);
         Assertions.assertTrue(queue.runAllToCompletion());
     }
+    
+    @Test
+    public void testCopyCell() {
```
eddieh-xlnx, line 242:
```suggestion
    public void testCopyDSP58Cell() {
```

clavin-xlnx: I'm not sure if the fact that I used a DSP_PREADD58 cell is the relevant bit.  More so it is trying to test copying a cell that belongs exclusively to a secondary site type.

PR #426, opened 2022-05-16
Fix #422: Correctly resolve HierObject search with prefix match and '/' in name

### LINKED ISSUE #422: [copyCell] Cells with hierarchical names return null EDIFCellInst
(This is also related to the `copyCell` issue reported in #419. But I think the cause is quite different so I decided to file a new one instead.)

I have a design in which there are some cells with specific hierarchical names that crashes the tool when I tried to do copyCell(). I have created a minimal example that helps reproduce the issue.

**Verilog design**
```Verilog
module lut (
  input i0,                                                                      
  input i1,                                                                      
  input i2,                                                                      
  input i3,                                                                      
  output out0,                                                                   
  output out1                                                                    
);                                                                               
                                                                                 
  LUT2 #(                                                                        
    .INIT(4'h8)                                                                  
  ) i_ (                                                                         
    .I0(i0),                                                                     
    .I1(i1),                                                                     
    .O(out0)                                                                     
  );                                                                             
                                                                                 
  LUT2 #(                                                                        
    .INIT(4'h8)                                                                  
  ) \i_/a  (                                                                                                                                                                           
    .I0(i2),                                                                     
    .I1(i3),                                                                     
    .O(out1)                                                                     
  );                                                                             
endmodule
```
**RapidWright code**

```Java
public class TestCopyCell {                                                      
  public static void main(String[] args) throws FileNotFoundException {          
    Design design1 = Design.readCheckpoint("lut.dcp");               
    Design design2 = new Design("new_design", design1.getDevice().getName());       
    for (Cell c : design1.getCells()) {                                          
      System.out.println("Cell " + c.getName() + " " + c.getBEL() + " " + c.getSiteInst() + " " + c.getEDIFCellInst());
      SiteInst orig = c.getSiteInst();                                           
      design2.createSiteInst(orig.getName(), orig.getSiteTypeEnum(), orig.getSite());
      design2.copyCell(c, c.getName());                                          
    }                                                                            
  }                                                                              
}
```

The tool crashes when I have two LUTs with their instance names as above. The error that I got is

> Cell i_ H6LUT(BEL) SiteInst(name="SLICE_X182Y139", type="SLICEL", site="SLICE_X182Y139") i_
Cell i_/a G6LUT(BEL) SiteInst(name="SLICE_X182Y139", type="SLICEL", site="SLICE_X182Y139") null
Exception in thread "main" java.lang.NullPointerException
	at com.xilinx.rapidwright.design.Design.copyCell(Unknown Source)
	at com.xilinx.rapidwright.examples.TestCopyCell.main(TestCopyCell.java:18)

Since it fails to get an *EDIFCellInst* for cell *i_/a*, I suspect that might cause a crash in `copyCell`.

If I rename the second LUT from *\i_/a* to something like *\i0_/a*, there is no issue. Or if the design only contains one of the two LUTs, there's also no problem too.

Please find the source code in the attachment. I test this with RW 2022.1.1 and Vivado 2022.1.

Thank you!

[src.zip](https://github.com/Xilinx/RapidWright/files/8691857/src.zip)

clavin-xlnx:
This is an attempt to resolve incorrect resolution of a hierarchical name search of a netlist object that happens to use a hierarchy separator (`/`) in its name AND also has another instance name collision prefix in the existing netlist hierarchy.  The issue is illustrated in #422.  

Unfortunately, this implementation will increase runtime on those cases where a cell instance is queries but does not exist.


clavin-xlnx:
> How many slashes do we typically see in a CellInst's name? This idea would waste memory if we have names with many slashes in them without the prefix-cells also existing.

Most designs do not have any cell instances with slashes in their name.  A small percentage of designs do have some cell instances with slashes, and typically, I only see one slash per name, although it is quite possible to have more.  
 
> How often do we query by relative name (made slower by this tree idea) versus by hierarchical name (improved by this)? Or would we spend the memory to have both representations?

I think this is a rather rare circumstance, I don't really want to spend much to support it.  I think it will be slower when designs have this.  I think the idea is interesting and has merit, but, I think a more practical approach of flagging cells that have this as a way to trigger special their handling.

eddieh-xlnx:
Does the test here pass today?

PR #427, opened 2022-05-16
Allows DeviceBrowser to browse devices prior to downloading them.

clavin-xlnx:
Prior to this change, without downloading devices, the `DeviceBrowser` would be not functional.

### DIFF
--- a/src/com/xilinx/rapidwright/device/browser/DeviceBrowser.java
+++ b/src/com/xilinx/rapidwright/device/browser/DeviceBrowser.java
@@ -111,7 +111,7 @@ public class DeviceBrowser extends QMainWindow{
 		initializeSideBar();
 		
 		// Gets the available parts in RapidWright and populates the selection tree
-		Set<String> parts = WidgetMaker.getAvailableDevices();
+		Set<String> parts = WidgetMaker.getSupportedDevices();
 		if(parts.size() < 1){
 		    throw new RuntimeException("Error: No available parts. " +
 					"Please generate part database files.");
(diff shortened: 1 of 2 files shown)


eddieh-xlnx (commented):
As I understand, `WidgetMaker.getAvailableDevices()` leans on `PartNameTools.getParts()` to see all parts _supported_ by RapidWright.

Previously, `Device.getAvailableDevices()` only returns the set of devices for which the associated device file exists.

Is this understanding correct?
Perhaps `WidgetMaker.getAvailableDevices()` would be better named `WidgetMaker.getSupportedDevices()`?

Another question I have is whether attempted to browse a supported (but not available) device causes the device information to be downloaded prior to rendering it in DeviceBrowser?

clavin-xlnx:
> As I understand, `WidgetMaker.getAvailableDevices()` leans on `PartNameTools.getParts()` to see all parts _supported_ by RapidWright.
> 
> Previously, `Device.getAvailableDevices()` only returns the set of devices for which the associated device file exists.
> 
> Is this understanding correct? Perhaps `WidgetMaker.getAvailableDevices()` would be better named `WidgetMaker.getSupportedDevices()`?

Good point, I've refactored the name.

> Another question I have is whether attempted to browse a supported (but not available) device causes the device information to be downloaded prior to rendering it in DeviceBrowser?

Yes, the intent is to not force the user to download everything first, but to download on demand.  The first time they try to access a device, it will take an extra second or two (or five), but after that it should be fast.

eddieh-xlnx (approved):

PR #428, opened 2022-05-17
Save some memory in EDIF

jakobwenzel:
* Lower the initial size of HashMaps to conserve space
* The parallel EDIF parser did not uniquify port inst names


clavin-xlnx (approved):
It's possible that since `EDIFName.getNewMap()` is called in a variety of places, that there are more optimal numbers for certain objects vs. others.  However, I ran this and tested it on a few designs and it doesn't impact runtime significantly and provides 8-14% memory savings on the netlist.

PR #435, opened 2022-05-19
Copying static net while copyImplementation

pongstorn-xilinx:


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2228,6 +2228,26 @@ public static Map<String,String> getInvertiblePinMap(Series series, Unisim unisi
...
+		Design d2 = new Design(srcCellNetlist);
+		d2.setAutoIOBuffers(false);
+		d2.setDesignOutOfContext(true);
+
+		// TODO: Skip this step if the design was not implemented.
+		Map<String, String> cellMap = Collections.singletonMap(cellName, "");
+		DesignTools.copyImplementation(src, d2,  keepStaticRouting, true, true, true, cellMap);
+		return d2;
+	}
+
```
eddieh-xlnx, line 2250: Is this a necessary part of this PR?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2241,14 +2261,32 @@ public static Map<String,String> getInvertiblePinMap(Series series, Unisim unisi
...
+	 * @param copyStaticNets Flag indicating if static nets should be copied
+	 * @param copyOnlyInternalNets Flag indicating if only nets with every terminal inside the cell should be copied
+	 * @param lockPlacement Flag indicating if the destination implementation copy should have the
+	 * 	placement locked
+	 * @param lockRouting Flag indicating if the destination implementation copy should have the
+	 * 	routing locked
+	 * @param srcToDestInstNames A map of source (key) to destination (value) pairs of cell
+	 * instances from which to copy the implementation
+	 */
+	public static void copyImplementation(Design src, Design dest, boolean copyStaticNets, boolean copyOnlyInternalNets, boolean lockPlacement,
```
eddieh-xlnx, line 2282: Two new parameters have been added to this method: `copyOnlyInternalNets` and `copyStaticNets`. Are they both necessary?

Would it be better to pick a behaviour as default and stick with that?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2241,14 +2261,32 @@ public static Map<String,String> getInvertiblePinMap(Series series, Unisim unisi
 	 * instances from which to copy the implementation. If targeting the top instance, use an 
 	 * empty String ("") as the destination instance name.
 	 */
-	public static void copyImplementation(Design src, Design dest, boolean lockPlacement, 
+	public static void copyImplementation(Design src, Design dest, boolean lockPlacement,
+										  boolean lockRouting, Map<String,String> srcToDestInstNames) {
+		copyImplementation(src, dest, false, false, lockPlacement, lockRouting, srcToDestInstNames);
```
eddieh-xlnx, line 2266: Are `copyStaticNets=false` and `copyOnlyInternalNets=false` good defaults?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2345,7 +2395,128 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
+			put("DMUX", Arrays.asList("D5LUT", "D6LUT"));
+			put("E_O",  Arrays.asList("E5LUT", "E6LUT"));
+			put("EMUX", Arrays.asList("E5LUT", "E6LUT"));
+			put("F_O",  Arrays.asList("F5LUT", "F6LUT"));
+			put("FMUX", Arrays.asList("F5LUT", "F6LUT"));
+			put("G_O",  Arrays.asList("G5LUT", "G6LUT"));
+			put("GMUX", Arrays.asList("G5LUT", "G6LUT"));
+			put("H_O",  Arrays.asList("H5LUT", "H6LUT"));
+			put("HMUX", Arrays.asList("H5LUT", "H6LUT"));
+		}};
```
eddieh-xlnx, line 2432: Can this be made into a static map?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2345,7 +2395,128 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
+
+						sitePin = node.getSitePin();
+					}
+					if ((sitePin != null) && !node.getWireName().contains("VCC_WIRE"))  { // GND source
+						String  pinName = sitePin.getPinName();
+						String  siteName = sitePin.getSite().getName();
+						List<String> bels = sitePin2Bels.get(pinName);
+						for (String bel : bels) {
+							prohibitBels.add(siteName + "/" + bel);
+						}
```
eddieh-xlnx, line 2499: These BELs need to be prohibited because they are used to supply static sources, right? Let's drop that as a comment for the future.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2345,7 +2395,128 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
+					}
+					netToPIPs.put(net,allPIPs);
+				}
+			}
+		}
+
+		for (String bel : prohibitBels) {
+			dest.addXDCConstraint(ConstraintGroup.LATE, "set_property PROHIBIT true [get_bels " + bel + "]");
+		}
+
```
eddieh-xlnx, line 2509: I also recall that we did or wanted to look at PROHIBIT-ing these BELs in a different way than using a XDC constraint. Do you remember what the outcome was?

Using XDC doesn't seem to be the best way if the consumer was RapidWright?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2345,7 +2395,128 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
+			put("EMUX", Arrays.asList("E5LUT", "E6LUT"));
+			put("F_O",  Arrays.asList("F5LUT", "F6LUT"));
+			put("FMUX", Arrays.asList("F5LUT", "F6LUT"));
+			put("G_O",  Arrays.asList("G5LUT", "G6LUT"));
+			put("GMUX", Arrays.asList("G5LUT", "G6LUT"));
+			put("H_O",  Arrays.asList("H5LUT", "H6LUT"));
+			put("HMUX", Arrays.asList("H5LUT", "H6LUT"));
+		}};
+
+		// Map from a node to its driver PIP
```
eddieh-xlnx, line 2434: This functionality looks a lot like `DesignTools.getTrimmablePIPsFromPins()`, which takes a `Net` and a set of `SitePinInst`-s, and returns a set of PIPs necessary to feed those. Is it possible to reuse that code?

https://github.com/Xilinx/RapidWright/blob/7aafb67e3460f5373d0911c722cc39764d039efc/src/com/xilinx/rapidwright/design/DesignTools.java#L1107

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: This could be moved to PR 390.  The only reason I see for the move is it call copyImplementation with 3 flags as true, which I don't know if it is generic.

If we don't think, the flags set are not generic, we should move it to PR 390.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: Yes, it it does not break any existing use cases.

Is there any existing code that use copyImplementation?  I presume yes, because copyImplementation added before this PR. The behavior of the existing copyImplementation is maintained with these two flags false.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: I believe it depends on a use case.  see above comment.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: Will move to be private static of DesignTools.   My only reservation is that it will pollute the class because it is not used anywhere else in the class.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: comment added.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: Eddie, I remember that you want to use tag. IIRC, after your investigation, you conclude that it is not possible.

You have a point that this won't work if RapidWright placer is invoked after. But, I don't think we can have a good solution for that in a reasonable timeframe. I will add TODO for that.

eddieh-xlnx: I have a faint recollection I/we looked at this but it's not clear why we wouldn't be able to block off those BELs.

Perhaps @clavin-xlnx can shed some light here?

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: There is at least one difference.  Let's consider a route from one GND source (S) to two sinks (T1 and T2) and the last common node/pip is X.  If only T1 is in a cell to be copied, the tracing inside copyStaticNets will extract all pips from S to T1. However, based on the description of getTrimmablePIPsFromPins, only the pips from X to T1 are returned.

eddieh-xlnx: Ah yes, you're right! Let's jot that down as a comment so I don't forget about it in the future.

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: Yes, I think that's a good idea since we aren't using nor testing it in this PR.

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: I would say it's okay to break older under-featured functionality in favour of more working features. It seems reasonable to expect static nets to be copied, and to not copy them be the exception (which users can explicitly set with the overload).

What's the difference between the two values for `copyOnlyInternalNets`? Why was the old functionality `false` -- meaning it would copy a net's sinks (if it was internal) but not the source since that's external?

pongstorn-xilinx: Based on previous conversation with Chris, it is always expected that Vivado will be invoked after copyImplementation. Thus, I would consider my use case an outliner that need to preserve the static routes. Besides, we can defer this decision until when we need to make it.

pongstorn-xilinx: @clavin-xlnx, do you have comment about eddie question?

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx: I would say we could make copying static nets the default.  However, I can see situations that might be useful where partial routing is copied.  So maybe keeping one of the flags would be useful.  Perhaps `copyOnlyInternalNets` could be inverted and called `copyPartialNetRouting`

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2345,7 +2398,113 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
+					// Backtrack through routing nodes (no SitePin)
+					while ((sitePin == null) || sitePin.isInput()) {
+						PIP pip = nodeToDriverPIP.get(node);
+						allPIPs.add(pip);
+						if (pip.isBidirectional()) {
+							node = pip.getStartNode().equals(node) ? pip.getEndNode() : pip.getStartNode();
+						} else {
+							node = pip.getStartNode();
+						}
+						if (node.getWireName().contains("VCC_WIRE"))
```
clavin-xlnx, line 2474:
```suggestion
						if (node.getWireName().contains(Net.VCC_WIRE_NAME))
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2345,7 +2398,113 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
+							break;
+
+						sitePin = node.getSitePin();
+					}
+					if ((sitePin != null) && !node.getWireName().contains("VCC_WIRE"))  { // GND source
```
clavin-xlnx, line 2479:
```suggestion
					if ((sitePin != null) && !node.getWireName().contains(Net.VCC_WIRE_NAME))  { // GND source
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2345,7 +2398,113 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
+	 * @param siteInstsOfCells The set of SiteInst containing the sinks of the static nets
+	 */
+	private static void copyStaticNets(Design dest, List<Net> staticNets, Set<SiteInst> siteInstsOfCells) {
+		// This method traces a route similar to that in getTrimmablePIPsFromPins. However, there is one subtle difference.
+		// Let's consider a route from one GND source (S) to two sinks (T1 and T2) and the last common node/pip is X.
+		// If only T1 is in a cell to be copied, the tracing code here will extract all pips from S to T1.
+		// However, getTrimmablePIPsFromPins return only the pips from X to T1.
+
+		// Map from a node to its driver PIP
+		// Note: Some PIPs are bidirectional. But, every PIP performs only one direction.
```
clavin-xlnx, line 2421:
```suggestion
		// Note: Some PIPs are bidirectional. But, every PIP allows the signal to flow in only one direction.
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2345,7 +2398,113 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
+		// The direction of a bidirectional PIP is determined from the context, ie., its connecting directional PIPs.
+		// To determine that context, go through directional PIPs first. This process does not support consecutive bidirectional PIPs.
+		Map<Net,Map<Node,PIP>> netToUphillPIPMap = new HashMap<>();
+		// netToPIPs to store PIPs extracted for static nets.
+		Map<Net,Set<PIP>> netToPIPs = new HashMap<>();
+		for (Net net : staticNets) {
+			netToPIPs.put(net, new HashSet<>());
+			Map<Node,PIP> nodeToDriverPIP = new HashMap<>();
+			List<PIP> biPIPs = new ArrayList<>();
+			// TODO: This goes through the whole design. Try to limit it to the target cell, if possible.
```
clavin-xlnx, line 2431: Is this TODO still relevant?

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx: From what we looked at, Vivado's only mode to exclude placement on those locations is with the `PROHIBIT` constraint.  There is obviously the `EXCLUDE_PLACEMENT` on a pblock, but that is less precise.  

I don't exactly like the idea of adding the `PROHIBIT` constraints, but I don't have any better suggestions.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: move to PR 390

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: We should not rush to a not-urgent decision and should leave the default as it is. The default change should be made outside of this PR.

pongstorn-xilinx: > I would say we could make copying static nets the default. However, I can see situations that might be useful where partial routing is copied. So maybe keeping one of the flags would be useful. Perhaps `copyOnlyInternalNets` could be inverted and called `copyPartialNetRouting`
When copyOnlyInternalNets is true, only the net completely in the cell is copy, ie., no partial nets are copied.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: Apart from my reason above, I'm not confident to make copy static nets a default for one technical reason. Copying static nets requires preserving their sources. These sources could be outside of the placement footprint of other cells. I can imagine some users do not want those location blocked, but want to merge the static sources with those of other cells.  

It is true that such users can turn off copying static nets, if that a default. There is no clear winner and thus logistically, we should not hold up this PR and defer the decision until the full impact is vetted.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: moved

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: We could use two modes. Prohibit a bel is to communicate to Vivado. Another is to communicate to downsteam RW (which we don't have and can be added when the need arise.)

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: changed.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: changed

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: Yes. I don't know how to do that and don't even know if it is possible because static nets can go through many hierarchy.  

If there is no way to do it, should it be removed?

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: I was thinking of locked cells, set using `Cell.setLocked()` or `Cell.setEmptyLocked()`? What are the purposes of those cells if not to act as a prohibit?

test/src/com/xilinx/rapidwright/design/TestDesignTools.java
```diff
@@ -114,6 +117,59 @@ public void testCopyImplementationRouteThruVCCPinCheck() {
...
+        DesignTools.copyImplementation(src, d2,  keepStaticRouting, true, true, true, cellMap);
+
+
+        List<Net> staticNets = new ArrayList();
+        for (Net net : d2.getNets()) {
+            if (net.isStaticNet()) {
+                staticNets.add(net);
+                continue;
+            }
+        }
```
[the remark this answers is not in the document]
pongstorn-xilinx: changed.

src/com/xilinx/rapidwright/design/DesignTools.java
pongstorn-xilinx: @clavin-xlnx What is your thought on this?     Are they transfered to dcp?  If not, they are not our options.

test/src/com/xilinx/rapidwright/design/TestDesignTools.java
```diff
@@ -115,6 +118,52 @@ public void testCopyImplementationRouteThruVCCPinCheck() {
         }
     }
 
+
```
eddieh-xlnx, line 121:
```suggestion
```

test/src/com/xilinx/rapidwright/design/TestDesignTools.java
```diff
@@ -115,6 +118,52 @@ public void testCopyImplementationRouteThruVCCPinCheck() {
...
+    public void testCopyImplementation() {
+        boolean keepStaticRouting = false;
+        HashMap<String, Integer> numPIPs = new HashMap<String, Integer>()
+        {{
+            put("GLOBAL_LOGIC0",  0);
+            put("GLOBAL_LOGIC1",  0);
+        }};
+
+        testCopyImplementationHelper(keepStaticRouting, numPIPs);
+    }
```
eddieh-xlnx, line 166:
```suggestion
    }

```

test/src/com/xilinx/rapidwright/design/TestDesignTools.java
```diff
@@ -115,6 +118,52 @@ public void testCopyImplementationRouteThruVCCPinCheck() {
...
+        boolean keepStaticRouting = true;
+        HashMap<String, Integer> numPIPs = new HashMap<String, Integer>()
+        {{
+            put("GLOBAL_LOGIC0",  201);
+            put("GLOBAL_LOGIC1",  601);
+        }};
+
+        testCopyImplementationHelper(keepStaticRouting, numPIPs);
+    }
+    @Test
```
eddieh-xlnx, line 156:
```suggestion

    @Test
```

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: Take out TODO. Mention limitation in javadoc.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2390,6 +2441,112 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
+                    // Backtrack through routing nodes (no SitePin)
+                    while ((sitePin == null) || sitePin.isInput()) {
+                        PIP pip = nodeToDriverPIP.get(node);
+                        allPIPs.add(pip);
+                        if (pip.isBidirectional()) {
+                            node = pip.getStartNode().equals(node) ? pip.getEndNode() : pip.getStartNode();
+                        } else {
+                            node = pip.getStartNode();
+                        }
+                        if (node.getWireName().contains("VCC_WIRE"))
```
eddieh-xlnx, line 2516:
```suggestion
                        if (node.getWireName().contains(Net.VCC_WIRE_NAME))
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2390,6 +2441,112 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
+                            break;
+
+                        sitePin = node.getSitePin();
+                    }
+                    if ((sitePin != null) && !node.getWireName().contains("VCC_WIRE"))  { // GND source
```
eddieh-xlnx, line 2521:
```suggestion
                    if ((sitePin != null) && !node.getWireName().contains(Net.VCC_WIRE_NAME))  { // GND source
```

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: Conclusion: `Cell.set*` methods don't achieve the same as `PROHIBIT` constraint.

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: Leave as is.

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: Leave as is.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2390,6 +2441,112 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
...
+                            prohibitBels.add(siteName + "/" + bel);
+                        }
+                    }
+                    netToPIPs.put(net,allPIPs);
+                }
+            }
+        }
+
+        // When we copy the static nets, we must preserve their sources so that nothing should be placed on it.
+        // TODO: need to annotate so that RapidWright placer avoids placing on these bels.
```
eddieh-xlnx, line 2535:
```suggestion
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2308,19 +2348,21 @@ public static void copyImplementation(Design src, Design dest, boolean lockPlace
         for (String srcPrefix : srcToDestInstNames.keySet()) {
             if (srcPrefix.length()==0) {
                 prefixes.put(srcPrefix, srcPrefix);
-            } else {
+            }else {
```
eddieh-xlnx, line 2351:
```suggestion
            } else {
```

eddieh-xlnx (approved):

test/src/com/xilinx/rapidwright/design/TestDesignTools.java
```diff
@@ -115,6 +118,53 @@ public void testCopyImplementationRouteThruVCCPinCheck() {
...
+        testCopyImplementationHelper(keepStaticRouting, numPIPs);
+    }
+
+    @Test
+    public void testCopyImplementation() {
+        boolean keepStaticRouting = false;
+        HashMap<String, Integer> numPIPs = new HashMap<String, Integer>()
+        {{
+            put("GLOBAL_LOGIC0",  0);
+            put("GLOBAL_LOGIC1",  0);
```
clavin-xlnx, line 162:
```suggestion
            put(Net.GND_NET,  0);
            put(Net.VCC_NET,  0);
```

test/src/com/xilinx/rapidwright/design/TestDesignTools.java
```diff
@@ -115,6 +118,53 @@ public void testCopyImplementationRouteThruVCCPinCheck() {
...
+        Assertions.assertEquals(numPIPs.get(gndNet.getName()), gndNet.getPIPs().size());
+    }
+
+    @Test
+    public void testCopyImplementationWithCopyStaticNets() {
+        boolean keepStaticRouting = true;
+        HashMap<String, Integer> numPIPs = new HashMap<String, Integer>()
+        {{
+            put("GLOBAL_LOGIC0",  201);
+            put("GLOBAL_LOGIC1",  601);
```
clavin-xlnx, line 150:
```suggestion
            put(Net.GND_NET,  201);
            put(Net.VCC_NET,  601);
```

clavin-xlnx (approved):

PR #440, opened 2022-05-20
Give OS time to catch up before checking open files

clavin-xlnx:
Ideally, we'd call `System.gc()` right before `getOpenFiles()` to ensure that everything is cleaned up properly before and after the test, but that doubles the test runtime.  But perhaps some spurious failures are just due to a race condition between tests starting and finishing.  This inserts 10ms before the start and end of each test's call to `getOpenFiles()`, this might help remove the false positives in testing, but we'll have to deploy it to be sure.


clavin-xlnx:
Closed in favor of #443

PR #457, opened 2022-06-10
Changes to correspond to Cell updates

clavin-xlnx:
The `Cell` object will soon store `EDIFHierCellInst` instead of `EDIFCellInst`.  This change corresponds to that update.


src/com/xilinx/rapidwright/interchange/PhysNetlistReader.java
```diff
@@ -186,7 +186,7 @@ private static void readPlacement(PhysNetlist.Reader physNetlist, Design design,
                 // c Alternative Blocked Site Type // TODO
             } else if(physCells.get(cellName) == PhysCellType.PORT) {
                 siteInst.getBEL(belName);
```
eddieh-xlnx, line 188:
```suggestion
```

eddieh-xlnx (approved):

PR #462, opened 2022-06-28
Automatically Check License Header

jakobwenzel:
This adds a make target to verify that all Java files have a license header. This target is also run in CI. An example failed run can be found here: https://github.com/jakobwenzel/RapidWright/runs/7092413363?check_suite_focus=true

The make target is intended to be run in a *pre-commit hook*. It uses `git grep --cached` to only look at the state of files as they will be committed. 

Pre-Commit hooks are run before the user enters a commit message to verify that a commit is sensible. Hooks are not set in the repository itself, but every user has to install the hook in their local checked out copy.

To make sure that we can easily update the hook, I'd suggest we standardize on this hook script:

* Create the file `.git/hooks/pre-commit` with this content:
  ```
  #!/bin/bash
  make pre_commit
  ```
* Mark the file as executable

If we want to add more pre-commit targets later, we only have to change the Makefile and not update every user's hook script.

Should we add info about the hooks to the Readme?


clavin-xlnx:
@jakobwenzel - Circling back to this issue, I think we decided to add a `make` target that would handle creating the pre-commit script if it doesn't already exist, right?

jakobwenzel:
Oh right, I forgot about that part. I added it now.

jakobwenzel:
Ready to merge, fixed another missing header on master.

PR #463, opened 2022-06-28
Optimize EDIF Memory usage

jakobwenzel:
This trades EDIF load (+15%) and save (+8%) runtime for memory (-18%) by no longer storing the EDIF renames. This also includes a number of other changes that happened along the way:
* Optional switch on writing EDIF that makes the written EDIF deterministic. Two files with this switch enabled should be diff-able without any noise.
* Move the owner field from EDIFPropertyObject into EDIFPropertyValue. This makes our in-memory representation follow EDIF files more closely while also saving memory.
* Use an OutputStream instead of a Writer when outputting EDIF. This avoids repeatedly encoding the same Strings (hardcoded Syntax strings and Renames) into `byte[]`.


jakobwenzel:
Finally ready for review. Added a summary of the PR in the opening comment

.gitignore
```diff
@@ -10,3 +10,4 @@
 *.iml
 /designs/
 interchange/schema/capnp/java.capnp
+*.orig
```
eddieh-xlnx, line 13: Is this necessary?

src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -47,313 +48,323 @@
...
 		instances.put(instance.getName(), instance);
 		return instance;
 	}
-	
-	public EDIFCellInst getCellInst(String name){
-		if(instances == null) return null;;
+
+	public EDIFCellInst getCellInst(String name) {
+		if (instances == null) return null;
+		;
```
eddieh-xlnx, line 186:
```suggestion
```

src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -485,123 +500,136 @@ public boolean hasCompatibleInterface(EDIFCell cell) {
...
 
-    @Override
-    public String getUniqueKey() {
-        return getLibrary().getName() + "_" + getName();
-    }
-        
 	/**
 	 * Recursively finds all leaf cell descendants of this cell
-	 *
+	 * <p>
```
eddieh-xlnx, line 571:
```suggestion
	 *
```

src/com/xilinx/rapidwright/edif/EDIFLibrary.java
```diff
@@ -118,16 +157,17 @@ public void setNetlist(EDIFNetlist netlist) {
...
 	/**
-	 * Removes the cell by 'legal EDIF name' (rename construct)
-	 * @param legalEdifName The legal EDIF name (rename) of the cell to remove
+	 * Removes the cell by name
+	 * @param name The name of the cell to remove
 	 * @return The removed cell, or null if it did not exist in the library.
 	 */
-	public EDIFCell removeCell(String legalEdifName){
-		return cells == null ? null : cells.remove(legalEdifName);
+	//TODO this method changed from legal name to regular name. Do we want to do anything special?
```
eddieh-xlnx, line 168: Does this need resolving?

src/com/xilinx/rapidwright/edif/EDIFLibrary.java
```diff
@@ -137,16 +177,17 @@ public EDIFCell removeCell(String legalEdifName){
...
 	/**
-	 * Checks if the library contains a cell with the legal EDIF Name provided.
-	 * @param legalEDIFName The legal EDIF name of the cell to query in the library.
+	 * Checks if the library contains a cell with the name provided.
+	 * @param name The name of the cell to query in the library.
 	 * @return True if a cell by such name was found in the library, False otherwise.
 	 */
-	public boolean containsCell(String legalEDIFName){
-		return cells == null ? false : cells.containsKey(legalEDIFName);
+	//TODO this method changed from legal name to regular name. Do we want to do anything special?
```
eddieh-xlnx, line 188: Does this need resolving?

src/com/xilinx/rapidwright/edif/EDIFLibrary.java
```diff
@@ -214,12 +252,17 @@ public void uniqueifyCellsWithPrefix(String name){
 	 * the list.  This is a requirement when exporting the EDIF to a file.
 	 * @return The ordered list.
 	 */
```
eddieh-xlnx, line 254: Spacing looks a bit funny. Plus document `@param stable` and why you would want it on/off.

eddieh-xlnx (commented):
Thanks @jakobwenzel! Can you add a high level overview of how the read/write caches are used and their lifetimes?

> Optional switch on writing EDIF that makes the written EDIF deterministic. Two files with this switch enabled should be diff-able without any noise.

This non-determinism is only applicable for parallel writes, right?

src/com/xilinx/rapidwright/edif/EDIFLibrary.java
```diff
@@ -79,16 +79,55 @@ public EDIFCell addCell(EDIFCell cell){
...
+		if (collision == null) {
+			return cell;
+		}
+		//Restore the old mapping
+		cells.put(cell.getName(), collision);
+
+		if (preferredSuffix == null) {
+			preferredSuffix = "collisionRename";
+		}
+		String newName = findUniqueCellName(cell.getName()+"_HDI_"+preferredSuffix);
```
eddieh-xlnx, line 113: Can you explain this behaviour some more please? The `preferredSuffix` comes from the EDIF rename, and then you're appending it after `_HDI_` to generate a new unique cell name?

Can you give an example?

I'm also thinking that maybe we shouldn't use the `_HDI_` moniker either, since that's what Vivado typically uses.

.gitignore
jakobwenzel: Oops, this is an unrelated change: IntelliJ likes to create these temporary files when handling merge conflicts. I can remove it if you oppose this change.

src/com/xilinx/rapidwright/edif/EDIFLibrary.java
jakobwenzel: This handles a weird edge case where two cells have the same name but only differ in their rename. This actually occurs in practice. I added an example here: https://github.com/eddieh-xlnx/RapidWrightDCP/pull/12

I'm open for discussion about migrating away from `_HDI_`.

jakobwenzel:
I forgot to discuss the two TODOs before submitting the PR. The point is that those two methods used to take a legal name, but now take a cell name. Outside the parser, there is no way to access a legal name. So all code that called these methods with a legal name has to change anyway. Therefore, I think it is pretty safe to change those methods.

The only way this could break is if there is some user code out there that call these methods with hardcoded strings.

jakobwenzel:
> Can you add a high level overview of how the read/write caches are used and their lifetimes?

The read cache stores actual renames found in the file and is used to figure out what elements to connect in the linking phase. Its lifetime is limited to parsing.

The write cache stores the renames we generate during writing. It is used to make sure we output the exact same rename when writing an element itself and when writing references to it. As an optimization, it stores generated renames as `byte[]` instead of `String`s. Its lifetime is limited to writing EDIF.
 
> > Optional switch on writing EDIF that makes the written EDIF deterministic. Two files with this switch enabled should be diff-able without any noise.
> 
> This non-determinism is only applicable for parallel writes, right?

This only applies to serial writes. We have non-determinism because of HashMaps. If stable is enabled, all HashMaps are sorted before iterating through them.
Parallel writing cannot easily be made deterministic without making it more or less serial. Here, we have additional non-determinism from a race condition: Different threads race to create a rename.

test/src/com/xilinx/rapidwright/edif/TestEDIFParseCellNames.java
```diff
@@ -0,0 +1,39 @@
+package com.xilinx.rapidwright.edif;
```
clavin-xlnx, line 1: License header?

src/com/xilinx/rapidwright/edif/EDIFLibrary.java
jakobwenzel: I have changed the code to use `_RW_`.

jakobwenzel:
Tests fail because this needs a new release.

eddieh-xlnx:
Looks like d371146e749881d27b61e387b2ff9e8396487230 is needed too.

eddieh-xlnx (approved):
Looks good to go into a staging branch.

Issue #464, opened 2022-06-28
Issues converting a post-placement .dcp file to an IF file


zhilix:
I ran this round-test command for physical netlist conversion:
`java com.xilinx.rapidwright.interchange.PhysicalNetlistExample FPGA01_postplace.dcp FPGA01_postplace_interchange.dcp`

And the errors messages are:
`zhilix@eda15:/home/local/eda15/zhilix/projects/RapidWright$ java com.xilinx.rapidwright.interchange.PhysicalNetlistExample FPGA01_postplace.dcp FPGA01_postplace_interchange.dcp
Exception in thread "main" java.lang.NullPointerException
        at com.xilinx.rapidwright.interchange.Enumerator.getKey(Enumerator.java:41)
        at com.xilinx.rapidwright.interchange.Enumerator.getIndex(Enumerator.java:52)
        at com.xilinx.rapidwright.interchange.PhysNetlistWriter.writeRouteBranch(PhysNetlistWriter.java:432)
        at com.xilinx.rapidwright.interchange.PhysNetlistWriter.writeRouteBranch(PhysNetlistWriter.java:470)
        at com.xilinx.rapidwright.interchange.PhysNetlistWriter.writeRouteBranch(PhysNetlistWriter.java:470)
        at com.xilinx.rapidwright.interchange.PhysNetlistWriter.populateRouting(PhysNetlistWriter.java:413)
        at com.xilinx.rapidwright.interchange.PhysNetlistWriter.writePhysNets(PhysNetlistWriter.java:328)
        at com.xilinx.rapidwright.interchange.PhysNetlistWriter.writePhysNetlist(PhysNetlistWriter.java:505)
        at com.xilinx.rapidwright.interchange.PhysicalNetlistExample.main(PhysicalNetlistExample.java:67)`

Also the .dcp and .edf files dumped by vivado after I ran place_design are here:
[FPGA01_postplace.zip](https://github.com/Xilinx/RapidWright/files/9004305/FPGA01_postplace.zip)

clavin-xlnx:
Thank you @zhilix for the test example and error message.  It turns out this is stumbling on a known issue in the Interchange schema.  I have created a PR (#466) for you which you can use as a work-around until we get the schema updated and RapidWright changed to accommodate.  Please give it a try and see if you are able to convert your design.  As noted in the PR, you'll have to run `update_clock_routing` after loading design back into Vivado if you hope to route it.

zhilix:
[...]

Issue #134, opened 2021-03-11
NullPointerException running Interchange.PhysicalNetlistExample


reillymck:
I am trying to run the PhysicalNetlistExample, as described in [interchange/README.md.](https://github.com/Xilinx/RapidWright/blob/master/interchange/README.md) Everything works great until I try to run the last line in the example.

When I use either of the following commands, I get a java.lang.NullPointerException.

`java com.xilinx.rapidwright.interchange.PhysicalNetlistExample pblock0.dcp pblock0_interchange.dcp`
`java com.xilinx.rapidwright.interchange.PhysicalNetlistExample pblock0.dcp pblock0.edf pblock0_interchange.dcp`

Here is the stack trace for reference:
```
Exception in thread "main" java.lang.NullPointerException
        at com.xilinx.rapidwright.interchange.Enumerator.getKey(Enumerator.java:19)
        at com.xilinx.rapidwright.interchange.Enumerator.getIndex(Enumerator.java:30)
        at com.xilinx.rapidwright.interchange.PhysNetlistWriter.writeRouteBranch(PhysNetlistWriter.java:389)
        at com.xilinx.rapidwright.interchange.PhysNetlistWriter.writeRouteBranch(PhysNetlistWriter.java:427)
        at com.xilinx.rapidwright.interchange.PhysNetlistWriter.populateRouting(PhysNetlistWriter.java:376)
        at com.xilinx.rapidwright.interchange.PhysNetlistWriter.writePhysNets(PhysNetlistWriter.java:285)
        at com.xilinx.rapidwright.interchange.PhysNetlistWriter.writePhysNetlist(PhysNetlistWriter.java:462)
        at com.xilinx.rapidwright.interchange.PhysicalNetlistExample.main(PhysicalNetlistExample.java:45)
```

clavin-xlnx:
This is because the DCP being loaded has routing stubs in the clock network.  I believe this DCP might be out-of-context and thus doesn't have a fully routed clock network.  It is using a half-PIP to denote where it wants the clock to go when the design gets fully routed.

The Interchange format doesn't yet support this notion of a half-PIP and so it errors out.  Here is the half-PIP:

`RCLK_CLEM_L_X5Y209/RCLK_CLEM_L.CLK_BUFCE_ROW_FSR_0_CLK_IN null`

[...]

PR #466, opened 2022-06-29
Work around for placed designs in Interchange Format

clavin-xlnx:
There is currently a deficiency in the Interchange Schema (see #134 and #464) that doesn't allow the partially routed clock of a placed design to be represented correctly.  This PR is an example work-around until the Interchange Schema can be updated with the necessary constructs.  Note, this simply omits the routings stubs in the clock routing and will be lost through translation.  These should be able to be reconstructed by running `update_clock_routing` in Vivado.


clavin-xlnx:
This has been formally fixed as part of #467, so this workaround is no longer needed.

PR #471, opened 2022-07-12
Get all leaf port insts for a given EDIFHierNet

clavin-xlnx:
This contains two changes to support the ECO flow.

1. A new method that gets all leaf port insts on an EDIFHierNet
2. Fixes to EDIFTools.connectPortInstsThruHier()

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFHierNet.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFHierNet.java
@@ -25,7 +25,13 @@
  */
 package com.xilinx.rapidwright.edif;
 
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Objects;
+import java.util.Queue;
 
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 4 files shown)


eddieh-xlnx:
Is it possible to synchronize with `EDIFNetlist.getSinksFromNet()` in terms of sharing code, and/or being in the same package?

src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -544,7 +546,18 @@ public static void connectPortInstsThruHier(EDIFHierPortInst src, EDIFHierPortIn
...
-	        snkNet.removePortInst(finalSnk.getPortInst());
+	        if(createdSrcNet && snkNet.getParentCell() == finalSrc.getNet().getParentCell()) {
+	            // Let's delete the net we created and use the existing snkNet instead
+	            EDIFNet net = finalSrc.getNet();
+	            if(snkNet != net) {
+	                net.getParentCell().removeNet(net);
+	                snkNet.addPortInst(finalSrc.getPortInst());
+	            } else {
+	                return;
+	            }
```
eddieh-xlnx, line 557:
```suggestion
	            }
	            return;
```
Is it safe to return immediately because at this point `finalSrc.getNet() == finalSnk.getNet()` and thus `finalSrc.getNet().addPortInst(finalSnk.getPortInst());` is not necessary?

src/com/xilinx/rapidwright/edif/EDIFHierNet.java
```diff
@@ -141,11 +147,80 @@ public boolean equals(Object obj) {
 	 */
 	@Override
 	public String toString() {
-		return "EDIFHierNet [hierarchicalInstName=" + hierarchicalInst + ", net=" + net + "]";
+		return getHierarchicalNetName();
```
clavin-xlnx, line 150: @eddieh-xlnx - Just FYI - the previous toString() is not consistent with `EDIFHierCellInst` and `EDIFHierPortInst`.  Here, I just want the full hierarchical name.

eddieh-xlnx: Good! I've been wanting to make this change too!

eddieh-xlnx (approved):

PR #472, opened 2022-07-13
Fix DesignTools.getTrimmablePIPsFromPins()

eddieh-xlnx:
`DesignTools.getTrimmablePIPsFromPins()` previously failed for a case like the following:

```
# (Numbers represent nodes)
0 -> 1 -> 2 -> 3 -> A_I
            +> A_X
```

where trying to unroute just the `A_X` pin would result in some PIPs marked for trimming, when the correct answer should be no PIPs.


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1091,10 +1093,14 @@ public static Set<PIP> getTrimmablePIPsFromPins(Net net, Collection<SitePinInst>
 	        if(p.getSite() == null) continue;
 	        if(p.getNet() != net) continue;
 	        Node sink = p.getConnectedNode();
+	        Integer fanoutCount = fanout.getOrDefault(sink, 0);
+	        if (fanoutCount > 1) {
+	        	// This node is used to connect another downstream pin, no more
+	        	// analysis necessary
+	        	continue;
+	        }
```
eddieh-xlnx, line 1101:
```suggestion
	        if (fanoutCount > 1) {
	            // This node is used to connect another downstream pin, no more
	            // analysis necessary
	            continue;
	        }
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1091,10 +1093,14 @@ public static Set<PIP> getTrimmablePIPsFromPins(Net net, Collection<SitePinInst>
 	        if(p.getSite() == null) continue;
 	        if(p.getNet() != net) continue;
 	        Node sink = p.getConnectedNode();
+	        Integer fanoutCount = fanout.getOrDefault(sink, 0);
+	        if (fanoutCount > 1) {
+	        	// This node is used to connect another downstream pin, no more
```
eddieh-xlnx, line 1098:
```suggestion
	        	// This node is also used to connect another downstream pin, no more
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1091,10 +1093,14 @@ public static Set<PIP> getTrimmablePIPsFromPins(Net net, Collection<SitePinInst>
 	        if(p.getSite() == null) continue;
 	        if(p.getNet() != net) continue;
 	        Node sink = p.getConnectedNode();
+	        Integer fanoutCount = fanout.getOrDefault(sink, 0);
+	        if (fanoutCount > 1) {
+	            // This node is used to connect another downstream pin, no more
```
clavin-xlnx, line 1098:
```suggestion
	            // This node is also used to connect another downstream pin, no more
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1076,12 +1076,14 @@ public static Set<PIP> getTrimmablePIPsFromPins(Net net, Collection<SitePinInst>
...
-	            fanout.put(startNode, 1);
-	        }else{
-	            fanout.put(startNode, count+1);
-	        }
+	        SitePin sp = startNode.getSitePin();
+	        SiteInst si = (sp != null && sp.isInput()) ? net.getDesign().getSiteInstFromSite(sp.getSite()) : null;
+	        SitePinInst spi = (si != null) ? si.getSitePinInst(sp.getPinName()) : null;
+
+			// If a site pin was found and it belongs to this net, add an extra fanout to
+			// reflect that it was both used for downstream connection as well as this site pin
```
clavin-xlnx, line 1084:
```suggestion
		// If a site pin was found and it belongs to this net, add an extra fanout to
		// reflect that it was both used for downstream connection as well as this site pin
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1076,12 +1076,14 @@ public static Set<PIP> getTrimmablePIPsFromPins(Net net, Collection<SitePinInst>
...
 	        }
 
-	        Integer count = fanout.get(startNode);
-	        if(count == null){
-	            fanout.put(startNode, 1);
-	        }else{
-	            fanout.put(startNode, count+1);
-	        }
+	        SitePin sp = startNode.getSitePin();
+	        SiteInst si = (sp != null && sp.isInput()) ? net.getDesign().getSiteInstFromSite(sp.getSite()) : null;
```
clavin-xlnx, line 1080: A few comments:

1) It might be best to store a reference to the `Design` object outside of the PIP loop.
2) Should we guard the last two lines with a null check for `sp`?  It would reduce the number of null checks for the common case.
3) Does this not happen for the `endNode` case?

eddieh-xlnx: For 2. which are the last two lines you are referring to? Feel free to use the code suggestion, I often find it easier/clearer to write small snippet of code to show what I mean than describe it using English. Pfffft to that imprecise language!

Re: 3. My thinking was that we only care about incrementing fanout if it is both a node-with-site-pin and is used by a PIP. You can't detect that condition by looking at the `endNode` of the PIP that takes you to the input site pin, but you can detect the condition on the following PIP that takes you away from the input pin. Is that logic sound?
In any case, this loop is about computing the fanout count of the `startNode`.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1076,12 +1077,14 @@ public static Set<PIP> getTrimmablePIPsFromPins(Net net, Collection<SitePinInst>
...
-	        }else{
-	            fanout.put(startNode, count+1);
-	        }
+	        SitePin sp = startNode.getSitePin();
+	        SiteInst si = (sp != null && sp.isInput()) ? design.getSiteInstFromSite(sp.getSite()) : null;
+	        SitePinInst spi = (si != null) ? si.getSitePinInst(sp.getPinName()) : null;
+
+		// If a site pin was found and it belongs to this net, add an extra fanout to
+		// reflect that it was both used for downstream connection as well as this site pin
+	        int fanoutCount = (spi != null && spi.getNet().equals(net)) ? 2 : 1;
```
clavin-xlnx, line 1086:
```suggestion
	        int fanoutCount = 1;
	        SitePin sp = startNode.getSitePin();
	        if(sp != null && sp.isInput()) {
	            SiteInst si = design.getSiteInstFromSite(sp.getSite());
	            SitePinInst spi = (si != null) ? si.getSitePinInst(sp.getPinName()) : null;
    
		    // If a site pin was found and it belongs to this net, add an extra fanout to
		    // reflect that it was both used for downstream connection as well as this site pin
	            if(spi != null && spi.getNet().equals(net)) {
	                fanoutCount = 2;
	            }
	        }
```
Sorry for the bad description.  The idea it to optimize for the common case where a PIP does not connect to a SitePin.

clavin-xlnx (approved):
FYI - I switch the test to run on a smaller device.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1076,12 +1077,19 @@ public static Set<PIP> getTrimmablePIPsFromPins(Net net, Collection<SitePinInst>
 	            rPips.add(pip);
 	        }
 
-	        Integer count = fanout.get(startNode);
-	        if(count == null){
-	            fanout.put(startNode, 1);
-	        }else{
-	            fanout.put(startNode, count+1);
+	        int fanoutCount = 1;
+	        SitePin sp = startNode.getSitePin();
```
clavin-xlnx, line 1081: I was thinking about this some more, and `Node.getSitePin()` can be expensive.  I'm thinking about an alternate solution.  Doesn't have to block this, but I think we might see some slow down with this approach as it is called a fair amount in the regressions.

clavin-xlnx: Since we know the net upfront and we are relying on the `SitePinInst` objects to be present, we can quickly build a map upfront of all the nodes connected to the `SitePinInst` of the net.  Then, we simply perform a hash lookup into the map to see if the node of interest has a pin on it.

clavin-xlnx: Take a look at https://github.com/Xilinx/RapidWright/pull/474

PR #473, opened 2022-07-14
Inconsistent behaviour for Cell.getSitePinFromLogicalPin()

eddieh-xlnx:
Despite intra-site routing from A3 -> AFF.D, `Cell.getSitePinFromLogicalPin()` will return the AX site pin.

### DIFF
--- a/test/src/com/xilinx/rapidwright/design/TestCell.java
+++ b/test/src/com/xilinx/rapidwright/design/TestCell.java
@@ -1,0 +1,34 @@
+package com.xilinx.rapidwright.design;
+
+import com.xilinx.rapidwright.device.BELPin;
+import com.xilinx.rapidwright.device.Device;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+
+public class TestCell {
+    @Test
+    void testGetSitePinFromLogicalPin() {
+        Design design = new Design("top", Device.PYNQ_Z1);
+        SiteInst si = design.createSiteInst("SLICE_X1Y0");
+        Net net = design.createNet("net");
+        SitePinInst A3 = net.createPin("A3", si);
+        SitePinInst AX = net.createPin("AX", si);
+
+        Cell ff = design.createAndPlaceCell("ff", Unisim.FDRE, "SLICE_X1Y0/AFF");
+
+        BELPin ffD = ff.getBEL().getPin("D");
+        for (SitePinInst spi : Arrays.asList(AX, A3)) {
+            Assertions.assertNull(DesignTools.getRoutedSitePin(ff, net, "D"));
+
+            BELPin bp = spi.getBELPin();
+            Assertions.assertTrue(si.routeIntraSiteNet(net, bp, ffD));
+
+            Assertions.assertEquals(bp.getName(), DesignTools.getRoutedSitePin(ff, net, "D"));
+            Assertions.assertEquals(spi, ff.getSitePinFromLogicalPin("D", null));
+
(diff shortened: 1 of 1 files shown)


test/src/com/xilinx/rapidwright/design/TestCell.java
```diff
@@ -110,4 +113,38 @@ public void testFFRoutethruCell() {
...
+
+        BELPin ffD = ff.getBEL().getPin("D");
+        for (SitePinInst spi : Arrays.asList(AX, A3)) {
+            if (spi == null)
+                continue;
+
+            Assertions.assertNull(DesignTools.getRoutedSitePin(ff, net, "D"));
+            if (createAX) {
+                // FIXME: Known broken -- see https://github.com/Xilinx/RapidWright/issues/473
+                Assertions.assertEquals("IN SLICE_X1Y0.AX", ff.getSitePinFromLogicalPin("D", null).toString());
```
eddieh-xlnx, line 136: @clavin-xlnx (see PR main description first). This is where I waive the possibly-incorrect result. The question is: is it really incorrect?

eddieh-xlnx: Thinking about it some more, when we have a net that reaches a site's `A3` and `AX` input pins, and for some reason there is no intra-site routing to follow, then I think it's reasonable `Cell.getSitePinFromLogicalPin()` to default to `AX` for flops.

clavin-xlnx: > The question in the `A3`-only case, it should consider a LUT routethru (where feasible) as a way of reaching that net's sink?

It seems like the behavior of `Cell.getSitePinFromLogicalPin()` is in question.  We should probably update the Javadoc of this method once we settle on a behavior.  I would say that this method should return the first routed `SitePinInst` if one exists.  If a routed `SitePinInst` does not exist, it seems like it should return null.  

In the case where the `SitePinInst` already exists, but is not routed **_and the `SitePinInst` belongs to the net connected to the logical pin on the cell_**, then I am ok with changing its behavior to identify what it should be.  But again, I think we should also update the Javadoc of `DesignTools.getRoutedSitePin()` because it currently claims to return `null` when there is no routed pin.  

I think defaulting to the `*X` input site pins for logical `D` inputs on `SLICE` flip flops makes sense.  For unoccupied LUT inputs site pins with no other options, I am ok with those when there is no other option.

eddieh-xlnx: > I would say that this method should return the first routed `SitePinInst` if one exists. If a routed `SitePinInst` does not exist, it seems like it should return null.

My understanding is that `DesignTools.getRoutedSitePin()` examines intra-site routing to tell you what the current site pin is -- which the test does check returns `null`. It looks like `Cell.getSitePinFromLogicalPin()` is for answering the question of what a future site pin could/should be -- presumably, ahead of performing any intra-site routing.

In this context, it looks like `Cell.getSitePinFromLogicalPin()` is working as expected, except it doesn't seem to consider LUT routethrus when looking to re-use existing site pins.

Just to be clear, calling `Cell.getSitePinFromLogicalPin()` on a site with no intra-site routing with the following net sinks:
- No sinks gives `null` (correct)
- `{A3, AX}` gives `AX` (correct)
- `AX` gives `AX` (correct)
- `A3` gives `null` (in question, but I think I can live with this until it bites me next time).

clavin-xlnx: I think this behavior is fine and if we want to extend the `A3` case to also return `A3`, that is fine too. But it would appear that the Javadoc descriptions are incorrect.

PR #475, opened 2022-07-14
Unroute source pin and preserve rest of net

clavin-xlnx:
This method is a complimentary method to `DesignTools.unroutePins()` in that it can unroute a source pin from a net while also keeping the routing from other source pins on the net intact.  This also sets the set of driven `SitePinInst` sinks from the unrouted routing to unrouted.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -1032,28 +1032,43 @@ public class DesignTools {
 	 * removed.
 	 */
 	public static void unroutePins(Net net, Collection<SitePinInst> pins) {
-	    Set<PIP> toRemove = getTrimmablePIPsFromPins(net, pins);
-	    if(toRemove.size() > 0) {
-	        ArrayList<PIP> updatedPIPs = new ArrayList<>();
-	        for(PIP pip : net.getPIPs()){
-	            if(!toRemove.contains(pip)) updatedPIPs.add(pip);
-	        }
-	        net.setPIPs(updatedPIPs);
-	    }
+	    removePIPsFromNet(net,getTrimmablePIPsFromPins(net, pins));
 	    for(SitePinInst pin : pins) {
 	        pin.setRouted(false);
 	    }	    
 	}
 	
-	public static Set<PIP> getPIPsDrivenBySourcePin(SitePinInst src) {
+	private static void removePIPsFromNet(Net net, Set<PIP> pipsToRemove) {
+	    if(pipsToRemove.size() > 0) {
+	        List<PIP> updatedPIPs = new ArrayList<>();
+	        for(PIP pip : net.getPIPs()){
+	            if(!pipsToRemove.contains(pip)) updatedPIPs.add(pip);
+	        }
+	        net.setPIPs(updatedPIPs);
+	    }	    
+	}
+	
+	/**
+	 * Unroutes a SitePinInst of a net.  This is desirable when a net has multiple SitePinInst
+	 * source pins (multiple outputs of a Site) and only a particular branch is desired to be 
+	 * unrouted.  If the entire net is to be unrouted, a more efficient method is {@link Net#unroute()}. 
+	 * @param src The source pin of the net from which to remove the routing 
+	 * @return The set of PIPs that were unrouted from the net.
+	 */
+	public static Set<PIP> unrouteSourcePin(SitePinInst src) {
 	    if(!src.isOutPin() || src.getNet() == null) return Collections.emptySet();
 	    Node srcNode = src.getConnectedNode();
-	    Set<PIP> pipsUsed = new HashSet<>();
+	    Set<PIP> pipsToRemove = new HashSet<>();
 	    
 	    Map<Node, List<PIP>> pipMap = new HashMap<>();
 	    for(PIP pip : src.getNet().getPIPs()) {
 	        Node node = pip.getStartNode();
 	        pipMap.computeIfAbsent(node, k -> new ArrayList<>()).add(pip);
+	    }
+	    
+	    Map<Node,SitePinInst> sinkNodes = new HashMap<>();
+	    for(SitePinInst sinkPin : src.getNet().getSinkPins()) {
+	        sinkNodes.put(sinkPin.getConnectedNode(), sinkPin);
 	    }
 	    
 	    Queue<Node> q = new LinkedList<>();
@@ -1063,18 +1078,26 @@ public class DesignTools {
 	        List<PIP> pips = pipMap.get(curr);
 	        if(pips != null) {
 	            for(PIP p : pips) {
-	                q.add(p.getEndNode());
-	                pipsUsed.add(p);
+	                Node endNode = p.getEndNode();
+	                q.add(endNode);
+	                pipsToRemove.add(p);
+	                SitePinInst sink = sinkNodes.get(endNode);
+	                if(sink != null) {
+	                    sink.setRouted(false);
+	                }
 	            }
 	        }
 	    }
 	    
-	    return pipsUsed;
+	    src.setRouted(false);
+        removePIPsFromNet(src.getNet(), pipsToRemove);
+	    return pipsToRemove;
 	}
 	
 	/**
 	 * For the given set of pins, if they were removed, determine which PIPs could be trimmed as 
-	 * they no longer route to any specific sink.
+	 * they no longer route to any specific sink.  This method is intended only for sink pins.  
+	 * See {@link #unrouteSourcePin(SitePinInst)} for handling source pin unroutes.
 	 * @param net The current net
 	 * @param pins The set of pins to remove.
 	 * @return The set of redundant (trimmable) PIPs that cane safely be removed when removing the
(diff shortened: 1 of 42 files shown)


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1032,20 +1032,72 @@ public static boolean removeConnectedRouting(Net net, Node node){
...
+	    
+	    src.setRouted(false);
+        removePIPsFromNet(src.getNet(), pipsToRemove);
+	    return pipsToRemove;
+	}
+	
 	/**
 	 * For the given set of pins, if they were removed, determine which PIPs could be trimmed as 
-	 * they no longer route to any specific sink.
+	 * they no longer route to any specific sink.  This method is intended only for sink pins.
```
clavin-xlnx, line 1099:
```suggestion
	 * they no longer route to any specific sink.  This method only works for sink pins.  
```

eddieh-xlnx (commented):
Overall comment: would it better to have `DesignTools.unroutePins()` handle both source and sink cases internally? This is the case for `Net.removePin()` which unroutes the whole net if it is a source pin, and for sink pins, either the whole net or just its branch (depending on the `preserveOtherRoutes` parameter)

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1028,24 +1028,77 @@ public static boolean removeConnectedRouting(Net net, Node node){
...
+	            if(!pipsToRemove.contains(pip)) updatedPIPs.add(pip);
+	        }
+	        net.setPIPs(updatedPIPs);
+	    }	    
+	}
+	
+	/**
+	 * Unroutes a SitePinInst of a net.  This is desirable when a net has multiple SitePinInst
+	 * source pins (multiple outputs of a Site) and only a particular branch is desired to be 
+	 * unrouted.  If the entire net is to be unrouted, a more efficient method is {@link Net#unroute()}.
```
eddieh-xlnx, line 1055: Would it be wise to do this automatically -- if the net only has one source pin (assuming the alternate source is populated correctly) then it can go ahead and unroute the whole net?

clavin-xlnx:
> Overall comment: would it better to have `DesignTools.unroutePins()` handle both source and sink cases internally? This is the case for `Net.removePin()` which unroutes the whole net if it is a source pin, and for sink pins, either the whole net or just its branch (depending on the `preserveOtherRoutes` parameter)

Indeed, it would be better.  That was my original approach and it got messy because of the unique aspects of the source-removal case.  It wasn't working so I refactored and fixed the code to the current state.  Now that it is working, perhaps you could help me find a more elegant way to merge them.

eddieh-xlnx:
Does this handle routing stubs?
Here's a VU440 route that has one source pin, two sink pins, and a stub:
```
INT_X72Y538/INT.LOGIC_OUTS_E12->>INT_NODE_SINGLE_DOUBLE_76_INT_OUT
INT_X72Y538/INT.INT_NODE_SINGLE_DOUBLE_76_INT_OUT->>SS2_E_BEG3
INT_X72Y536/INT.SS2_E_END3->>INT_NODE_IMUX_71_INT_OUT
INT_X72Y536/INT.SS2_E_END3->>INT_NODE_SINGLE_DOUBLE_109_INT_OUT
INT_X72Y536/INT.INT_NODE_SINGLE_DOUBLE_109_INT_OUT->>WW2_E_BEG4
INT_X71Y536/INT.WW2_E_END4->>INT_NODE_SINGLE_DOUBLE_47_INT_OUT
INT_X71Y536/INT.INT_NODE_SINGLE_DOUBLE_47_INT_OUT->>NN2_E_BEG4
INT_X71Y538/INT.NN2_E_END4->>INT_NODE_SINGLE_DOUBLE_1_INT_OUT
INT_X71Y538/INT.INT_NODE_SINGLE_DOUBLE_1_INT_OUT->>EE2_E_BEG5
INT_X72Y538/INT.EE2_E_END5->>INT_NODE_IMUX_16_INT_OUT
INT_X72Y538/INT.INT_NODE_IMUX_16_INT_OUT->>BOUNCE_E_14_FTN
INT_X72Y538/INT.INT_NODE_IMUX_16_INT_OUT->>BYPASS_E13
```

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx: In theory, yes we should.  But I'm wondering if we can always depend on the alternate source being populated correctly?  I guess we can make the change and find out, but perhaps we make that another PR.

clavin-xlnx:
Since I've tested this code and know that it works, it would be good to get it merged in as a reference point.  But, you make two good points:

1) Merge this functionality into `DesignTools.unroutePins()` and
2) Unroute the whole net if there is only one source pin.

How about we merge this and then I file an issue to resolve the other two issues?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1028,24 +1028,77 @@ public static boolean removeConnectedRouting(Net net, Node node){
...
+	                SitePinInst sink = sinkNodes.get(endNode);
+	                if(sink != null) {
+	                    sink.setRouted(false);
+	                }
+	            }
+	        }
+	    }
+	    
+	    src.setRouted(false);
+        removePIPsFromNet(src.getNet(), pipsToRemove);
```
eddieh-xlnx, line 1094:
```suggestion
	    src.setRouted(false);
	    removePIPsFromNet(src.getNet(), pipsToRemove);
```

test/src/com/xilinx/rapidwright/design/TestDesignTools.java
```diff
@@ -240,4 +237,115 @@ public void testGetTrimmablePIPsFromPins(String pinName) {
...
+            net.addPIP(device.getPIP(pip));
+        }
+        return net;
+    }
+    
+    @Test
+    public void testRemoveSourcePin() {
+        Design design = new Design("test", Device.KCU105);
+        
+        Net net1 = createTestNet(design, "net1", new String[]{
```
eddieh-xlnx, line 254:
```suggestion
        // Net with one source (AQ2) and two sinks (A_I & FX) and a stub (INT_NODE_IMUX_71_INT_OUT)
        Net net1 = createTestNet(design, "net1", new String[]{
```

test/src/com/xilinx/rapidwright/design/TestDesignTools.java
```diff
@@ -240,4 +237,115 @@ public void testGetTrimmablePIPsFromPins(String pinName) {
...
+        
+        Assertions.assertEquals(12, DesignTools.unrouteSourcePin(net1.getSource()).size());
+        
+        Assertions.assertEquals(0, net1.getPIPs().size());
+        for(SitePinInst pin : net1.getPins()) {
+            Assertions.assertFalse(pin.isRouted());
+        }
+
+        
+        Net net2 = createTestNet(design, "net2", new String[]{
```
eddieh-xlnx, line 284:
```suggestion
        // Net with one output (HMUX) and one input (SRST_B2)
        Net net2 = createTestNet(design, "net2", new String[]{
```
??

test/src/com/xilinx/rapidwright/design/TestDesignTools.java
```diff
@@ -240,4 +237,115 @@ public void testGetTrimmablePIPsFromPins(String pinName) {
...
+            Assertions.assertFalse(pin.isRouted());
+        }
+
+        net2.removePin(net2.getSource());
+        net2.removePin(net2.getPins().get(0));
+        design.removeSiteInst(design.getSiteInstFromSiteName("SLICE_X65Y158"));
+        design.removeSiteInst(design.getSiteInstFromSiteName("SLICE_X64Y158"));
+        
+        
+        Net net3 = createTestNet(design, "net3", new String[]{
```
eddieh-xlnx, line 309:
```suggestion
        // Net with two outputs (HMUX primary and H_O alternate) and two sinks (SRST_B2 & B2)
        Net net3 = createTestNet(design, "net3", new String[]{
```
???

test/src/com/xilinx/rapidwright/design/TestDesignTools.java
```diff
@@ -240,4 +237,115 @@ public void testGetTrimmablePIPsFromPins(String pinName) {
...
+            altSrc.setRouted(true);
+            Assertions.assertNotNull(net3.getAlternateSource());
+            Assertions.assertTrue(net3.getAlternateSource().getName().equals("H_O"));
+            si = design.createSiteInst(design.getDevice().getSite("SLICE_X64Y158"));
+            SitePinInst snk = net3.createPin("SRST_B2", si);
+            snk.setRouted(true);
+            SitePinInst altSnk = net3.createPin("B2", si);
+            altSnk.setRouted(true);
+            
+            Set<PIP> unroutedPIPs = DesignTools.unrouteSourcePin(net3.getAlternateSource());
```
eddieh-xlnx, line 343:
```suggestion
            // Unroute just the H_O alternate source
            Set<PIP> unroutedPIPs = DesignTools.unrouteSourcePin(net3.getAlternateSource());
```

eddieh-xlnx (approved):

PR #477, opened 2022-07-18
Fixes issue when connecting passthru module ports 

clavin-xlnx:
Also checks existing `SiteInst` placement to make sure not to overlap module instance placements.


src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -261,9 +283,10 @@ public boolean placeOnOriginalAnchor() {
 	 * does not match the floorplan according to the original module and simply leave it unplaced.  
 	 * Setting to false will cause placement to fail on first mismatch of floorplan placement 
 	 * attempt.
+	 * @param allowOverlap Return false if the module will overlap with existing placed logic.
```
eddieh-xlnx, line 286:
```suggestion
	 * @param allowOverlap True if the module instance is allowed to overlap with existing placed logic.
```

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -236,23 +238,43 @@ public ArrayList<Site> getAllValidPlacements(){
...
-	 * instances and nets within the module instance at their relative offsets of the new site.
+	 * instances and nets within the module instance at their relative offsets of the new site. Note
+	 * that this method allows placement overlap by default.  See 
+	 * {@link #place(Site, boolean, boolean)} to disallow module overlap.
 	 * @param newAnchorSite The new site for the anchor of the module instance.
 	 * @return True if placement was successful, false otherwise.
 	 */
 	public boolean place(Site newAnchorSite){	
-		return place(newAnchorSite, false);
+	    return place(newAnchorSite, false);
```
eddieh-xlnx, line 248: This calls `place()` with `allowOverlap=false` which is contrary to the javadoc which states that overlaps are allowed by default?

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -236,23 +238,43 @@ public ArrayList<Site> getAllValidPlacements(){
...
 	 * Places the module instance on the module's anchor site (original location of the module).  
-	 * This is the same as place(getModule().getAnchor().getSite()). 
+	 * This is the same as place(getModule().getAnchor().getSite()). Note
+	 * that this method allows placement overlap by default.  See 
+	 * {@link #place(Site, boolean, boolean)} to disallow module overlap.
 	 * @return True if the placement was successful, false otherwise.
 	 */
 	public boolean placeOnOriginalAnchor() {
-		return place(module.getAnchor(), false);
+	    return place(module.getAnchor(), false);
```
eddieh-xlnx, line 259: This calls `place()` with `allowOverlap=false` which is contrary to the javadoc which states that overlaps are allowed by default?

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -261,9 +283,10 @@ public boolean placeOnOriginalAnchor() {
 	 * does not match the floorplan according to the original module and simply leave it unplaced.  
 	 * Setting to false will cause placement to fail on first mismatch of floorplan placement 
 	 * attempt.
+	 * @param allowOverlap Return false if the module will overlap with existing placed logic.
 	 * @return True if placement was successful, false otherwise.
 	 */
-	public boolean place(Site newAnchorSite, boolean skipIncompatible){	
+	public boolean place(Site newAnchorSite, boolean skipIncompatible, boolean allowOverlap){
```
eddieh-xlnx, line 289: Would it be more descriptive to call the argument `allowClobbering` or `allowOverwrite` or simply `force` since it doesn't just overlap such that the module inst can still use the unused BELs -- the module inst's SiteInst appears to replace the existing SiteInst?

It's also not clear if that previous SiteInst will get unplaced correctly?

src/com/xilinx/rapidwright/design/ModuleInst.java
eddieh-xlnx: Oh wait, never mind. That's for `skipIncompatible`!

src/com/xilinx/rapidwright/design/ModuleInst.java
eddieh-xlnx: Oh wait, never mind. That's for `skipIncompatible`!

src/com/xilinx/rapidwright/design/ModuleInst.java
clavin-xlnx: Things are technically not lost and RapidWright still has tracking of each `SiteInst` location.  This is how the block placer operates and can recover from placing a module instance on top of another, and then subsequently moving it elsewhere.

eddieh-xlnx: Hmm. Let me wrap my head around that. Where is that overlapping placement tracked? If I asked the `Design` what `SiteInst` is located at a particular site, what answer would I get back just the last-placed `SiteInst`? If so, how would one get the ones "underneath" that one?

clavin-xlnx: There are two stored locations.  One global in `Design`, this would be the one "on top" and the other locally in the `SiteInst` itself.  Clearly this is not a perfect system.  When the block placer legalizes, that is when things are resolved.  So, there is some intermediate situations where as you noted, if checking for placement on a `Site` that previously had two, but now has one `SiteInst` matching placement, it wouldn't be found globally, only locally.   This issue doesn't manifest itself as when writing out a DCP, we iterate over `SiteInst` objects rather than the `Design` map.

eddieh-xlnx: Would you happen to know what holds a reference to the `SiteInst` that gets overwritten? I think it would be a good idea to document this behaviour. Historians from the future will thank you.

clavin-xlnx: There are two maps in the Design class that maintain the two different locations for placement.  The user would see the discrepancy through the APIs `Design.getSiteInstFromSite()` vs `Design.getSiteInsts()`.  Since each `SiteInst` keeps track of its own placement, getting a reference to the `SiteInst` and calling `SiteInst.getSite()` would be consistent in both the overlap/non-overlap case.  However, the `Design.getSiteInstFromSite()` would not be consistent as this is a map from last placement on a particular site to its `SiteInst` and removing an overlap doesn't restore that map to the existing/prior placement.

PR #481, opened 2022-07-25
add a JUnit test runner that runs all tests on LSF

jakobwenzel:


clavin-xlnx (approved):
Not sure if I understand how each part works, but looks ok to me.

PR #486, opened 2022-07-27
Refactor PhysNetlistWriter

eddieh-xlnx:
Also partially revert #253 by allowing routethru cells to be expressed again.


clavin-xlnx (approved):
Things look ok to me.

PR #489, opened 2022-08-01
optimized method to find net leaf source

jakobwenzel:
This adds a function to get the leaf `EDIFHierPortInst` of a `EDIFHierNet`.

This can already be done by searching the list returned by `EDIFHierNet.getLeafHierPortInsts` for outputs, but this method does not have to walk the net's whole hierarchy. Instead, it uses the ports' directions to walk straight to the source.


jakobwenzel:
This currently does not handle inouts.

PR #497, opened 2022-08-09
allow dangling up connections in getLeafHierPortInsts

jakobwenzel:
also introduces better error message on multi-source nets


eddieh-xlnx:
I remember we discussed this PR: whether it covers all cases, and whether it overlaps with functionality already present (but unmerged) in `sashimi`.

eddieh-xlnx:
@jakobwenzel This PR is a little stale now. If you still think this is important, get in touch!

PR #500, opened 2022-08-12
Fix SAT Router Example

clavin-xlnx:
- Allows the random SEED to be set from the command line
- Allows for return code 1 in evRouter (still generated valid solution in this case)

### DIFF
--- a/src/com/xilinx/rapidwright/examples/RunSATRouterExample.java
+++ b/src/com/xilinx/rapidwright/examples/RunSATRouterExample.java
@@ -42,7 +42,7 @@ public class RunSATRouterExample {
 			System.out.println("USAGE: java " + RunSATRouterExample.class.getCanonicalName() + " " 
 							+ "<placed_dcp_filename> <pblock_area_constraint> <output_dcp> [random seed]");
 			return;
-		}		
+		}
 		// Check for Vivado
 		String vivadoPath = FileTools.getVivadoPath();
 		if(vivadoPath == null || vivadoPath.length() == 0){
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/router/SATRouter.java
```diff
@@ -773,7 +773,7 @@ public void route(){
 		createNetsFiles();
 		createPipFile();
 		int result = runEvRouter();
-		if(result != 0){
+		if(result != 0 && result != 1){
```
eddieh-xlnx, line 776: What does return code of 1 mean?

src/com/xilinx/rapidwright/examples/RunSATRouterExample.java
```diff
@@ -38,11 +38,11 @@ public class RunSATRouterExample {
...
 	public static void main(String[] args) {
 		// Check args
-		if(args.length != 3){
+		if(args.length != 3 && args.length != 4){
 			System.out.println("USAGE: java " + RunSATRouterExample.class.getCanonicalName() + " " 
-							+ "<placed_dcp_filename> <pblock_area_constraint> <output_dcp>");
+							+ "<placed_dcp_filename> <pblock_area_constraint> <output_dcp> [random seed]");
 			return;
-		}
+		}
```
eddieh-xlnx, line 45:
```suggestion
		}
```

src/com/xilinx/rapidwright/router/SATRouter.java
clavin-xlnx: From the author of `evRouter`:
> It might not be intuitive but I think I am just using the same return codes as in standard SAT solvers.
> The returns code of evRouter are:
> -	0: the problem is unsatisfiable / unroutable
> -	1: the problem is routed and a solution is found
> -	2: the router cannot find a solution nor prove that it is unroutable with current limits.

PR #501, opened 2022-08-14
Upgrade QtJambi to support Qt6

makslevental:
Upgrade QtJambi in order to support Qt6 in order that the GUIs work on macos. Changes consist primarily of adjusting to the API differences between Qt4 and Qt6. 

With these changes I can use `DeviceBrowser` on my linux machine

Screenshot from 2022-08-13 19-37-27

and in principle should be able to likewise on a mac. Unfortunately, I have an M1 so I'm butting up against building QtJambi JNI libs for arm64 so I haven't been able to test yet. I will testdrive the rest of the GUIs soon. Let me know if there's anything you'd like me to change.

EDIT: after wrestling with QtJambi M1 build I'm able to see the gui on my M1 as well:

image

image

image

image


clavin-xlnx:
Cool!  Thanks for the PR!  I had looked in the past to upgrading, but could never really find suitable replacement jars.  [The current set of jars](https://search.maven.org/artifact/com.xilinx.rapidwright/qtjambi-macosx-gcc/4.5.2_01/jar) we use include the compiled libraries so that users don't have to install Qt separately.  I thought these did not include them, but it looks like they do!  

We'll start testing these, it may take a bit of time, but I think we will try moving to the newer version if we can.  One other issue we had was newer compiled libraries wouldn't run on older Linux (CentOS 7, despite it being old, is still quite popular in the EDA industry).  I tried running it out of the box and ran into an issue:

```
Exception in thread "main" io.qt.QLibraryNotFoundError: Library 'libQt6Core.so.6.3' was not found in 'java.library.path'
        at io.qt.internal.NativeLibraryManager.loadLibrary_helper(NativeLibraryManager.java:1608)
        at io.qt.internal.NativeLibraryManager.loadNativeLibrary(NativeLibraryManager.java:1315)
        at io.qt.internal.NativeLibraryManager.findAndLoadLibrary(NativeLibraryManager.java:1204)
        at io.qt.internal.NativeLibraryManager.loadQtCore(NativeLibraryManager.java:1104)
        at io.qt.internal.QtJambi_LibraryUtilities.<clinit>(QtJambi_LibraryUtilities.java:25)
        at io.qt.internal.QtJambiObject.<clinit>(QtJambiObject.java:39)
        at com.xilinx.rapidwright.MainEntrypoint.main(MainEntrypoint.java:208)

```
Did you set any special variables to get it to run under Linux?

build.gradle
```diff
@@ -11,6 +11,8 @@ plugins {
 
 apply from: 'common.gradle'
 
+applicationDefaultJvmArgs = ["-XstartOnFirstThread"]
```
clavin-xlnx, line 14: Looks like this line is great for Max OSX users, but doesn't work well for others.  Could we guard it so that it only takes effect on Mac OSX?

makslevental:
@clavin-xlnx 

> I thought these did not include them, but it looks like they do!

>Did you set any special variables to get it to run under Linux?

Those jars don't include Qt but they do include the compiled JNI libs. It looks like that error is from not having Qt in your LD_LIBRARY_PATH? I had to set 

```
applicationDefaultJvmArgs = ["/Users/mlevental/dev_projects/Qt/6.3.1/macos/lib"]
```

in `build.gradle` but I think there's an env variable you can set too.

>Looks like this line is great for Max OSX users, but doesn't work well for others. Could we guard it so that it only takes effect on Mac OSX?

Probably using the same gate I used in the dep (I don't know for sure I've never written line of gradle before this weekend!). I'll try and update the PR.

makslevental:
>https://github.com/OmixVisualization/qtjambi/blob/master/www/How-to-bundle-Qt-libraries.md

I saw that but didn't consider that would be an issue - was imagining you'd just expect people to have Qt installed but it makes sense to distribute.

>If licensing is favorable, perhaps we could deploy some of these.

I checked this before attempting the port; it's the same GPLv3/LGPLv21 as that of the implementation of qtjambi that you're currently using.

https://github.com/OmixVisualization/qtjambi/blob/master/LICENSE.GPLv3
https://github.com/OmixVisualization/qtjambi/blob/master/LICENSE.LGPLv21

PR #510, opened 2022-08-24
RWRoute: fixes routethru detection

YunxZhou:
This PR brings a "this" back to fix an oversight, where it means to check if two nodes make a routethru.


src/com/xilinx/rapidwright/rwroute/RoutableNode.java
```diff
@@ -131,10 +131,10 @@ public int setChildren(int globalIndex, Map<Node, Routable> createdRoutable, Set
 		children = new ArrayList<>();
 		List<Node> allDownHillNodes = node.getAllDownhillNodes();
 		
-		for(Node node:allDownHillNodes){		
+		for(Node node : allDownHillNodes){
 			if(reserved.contains(node)) continue;		
 			if(isExcluded(node, timingDriven)) continue;
-			if(routethruHelper.isRouteThru(node, node)) continue;
+			if(routethruHelper.isRouteThru(this.node, node)) continue;
```
eddieh-xlnx, line 137: Maybe to avoid confusion, can you instead rename the for loop variable to be `downhill` or similar?

eddieh-xlnx:
More broadly, can you provide an example of why this check is necessary -- is it just for performance reasons, or is there some breakage it avoids? It looks like this oversight has existed since RWRoute was released, but it doesn't /seem/ to have caused any problems...

src/com/xilinx/rapidwright/rwroute/RoutableNode.java
YunxZhou: > More broadly, can you provide an example of why this check is necessary -- is it just for performance reasons, or is there some breakage it avoids? It looks like this oversight has existed since RWRoute was released, but it doesn't /seem/ to have caused any problems...

Hi Eddie, thanks for the title clarification. Much clearer.

Regarding the routethru check, I was actually wondering about the same... The oversight has been there since Feb. when I was supposed to remove the heavy use of "this" regarding the PR #321 merging RapidStreamRoute. 

We needed the check because of a corner case where two nodes formulate a routethru that was not supported in RW. So we excluded such downhill nodes from the routing resource graph to avoid using them. As far as I remember, the routethru is a lut routethru. I am not sure if there has been updates fixing the routethrus. Perhaps @clavin-xlnx could comment a bit on this?

clavin-xlnx (approved):
@YunxZhou - I'm ok with this PR, but I don't think it fixes anything other than a name change and whitespace.  Is this what you intended?

eddieh-xlnx:
> @YunxZhou - I'm ok with this PR, but I don't think it fixes anything other than a name change and whitespace. Is this what you intended?

Wouldn't this changed line:
```
if(routethruHelper.isRouteThru(node, downhill)) continue;
```
represent a material change? My question is: how material is that since it's been broken since the beginning?

PR #513, opened 2022-08-25
EDIFNetlist.{copy,migrate}CellAndSubCells() to copy into new same-named library

### LINKED ISSUE #512: EDIFNetlist.copyCellAndSubCells() does not copy library correctly
EDIFNetlist.copyCellAndSubCells fails to copy some libraries associated with the given cell.

eddieh-xlnx:
Rather than into the work library, if source library doesn't exist in destination netlist.

Fixes #512.

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFNetlist.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFNetlist.java
@@ -553,7 +553,7 @@ public class EDIFNetlist extends EDIFName {
 			if(cell.getLibrary().isHDIPrimitivesLibrary()){
 				destLib = getHDIPrimitivesLibrary();
 			}else{
-				destLib = getWorkLibrary();
+				destLib = addLibrary(new EDIFLibrary(cell.getLibrary().getName()));
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


eddieh-xlnx:
Should `migrateCellAndSubCells()` get the same treatment?

src/com/xilinx/rapidwright/edif/EDIFNetlist.java
```diff
@@ -553,7 +553,7 @@ private EDIFCell copyCellAndSubCellsWorker(EDIFCell cell, Set<EDIFCell> copiedCe
 			if(cell.getLibrary().isHDIPrimitivesLibrary()){
 				destLib = getHDIPrimitivesLibrary();
 			}else{
-				destLib = getWorkLibrary();
+				destLib = addLibrary(new EDIFLibrary(cell.getLibrary().getName()));
```
clavin-xlnx, line 556: There might need to be a check here to see if the destination netlist already has the same named library.  If it does, we should just add the cell to it.

clavin-xlnx:
> Should `migrateCellAndSubCells()` get the same treatment?

Yes, that would probably be wise.

src/com/xilinx/rapidwright/edif/EDIFNetlist.java
eddieh-xlnx: That does get checked here a few lines up:

https://github.com/Xilinx/RapidWright/blob/c2770a77b7481745abc246b3f6cad61d722bc70f/src/com/xilinx/rapidwright/edif/EDIFNetlist.java#L551-L552

clavin-xlnx: Sorry, I missed that.  Thanks!

clavin-xlnx (approved):

PR #518, opened 2022-08-31
Test for Cell.getAllCorrespondingSitePinNames()

eddieh-xlnx:


eddieh-xlnx:
Actually, I'm not so sure an input BELPin should return an output site pin.

Closing for now.

PR #519, opened 2022-09-01
Retry EDIFHierPortInst.getPhysicalCell() after escaping \ again

eddieh-xlnx:
It's been observed that Vivado can doubly escape backslashes in the physical `Cell` name, so handle that case.


clavin-xlnx:
> @clavin-xlnx Are there any other cases where we go from `EDIFCellInst` to `Cell` that we should also update/all converge into a single pathway?

I'd like to better understand how names are being presented via the design/EDIF.   Java can be unintuitive at times with escape sequences.  Perhaps we need to investigate deeper into how these names are being presented to RapidWright and ensure that we are correctly/consistently parsing them on all fronts.

eddieh-xlnx:
Unclear as to whether it's a logical or physical problem. Closing for now.

PR #533, opened 2022-09-14
Tracks peak memory usage

clavin-xlnx:
Augments the `CodePerfTracker` so that when it is invoked to report memory usage, if the application is running in a Linux environment, it will also report the peak memory usage as reported by the OS (`/proc/<procid>/status`).


clavin-xlnx:
Some example output (not sure if I am sold on the formatting):

```
 XML Parse & Device Load:     3.016s    742.214MBs |   1040.953MBs (curr)   1499.652MBs (peak)
              EDIF Parse:    19.599s   1254.782MBs |   3754.152MBs (curr)   5458.785MBs (peak)
      Read PhysDB Caches:     0.493s    134.568MBs |   4208.629MBs (curr)   5458.785MBs (peak)
   (Read PhysDB Routing):                           (    4.731s)
 (Read PhysDB Placement):                           (    5.233s)
      Read Place & Route:     5.234s   1138.143MBs |   4359.684MBs (curr)   5555.418MBs (peak)
                 *Total*:    28.342s   3269.707MBs |   4538.379MBs (curr)   5555.418MBs (peak)
```

eddieh-xlnx:
It looks like there's not much value to the "current" field? This is the value seen by the OS, which is less accurate than JVM in terms of developer feedback. Peak value is valuable though!

clavin-xlnx:
> It looks like there's not much value to the "current" field? This is the value seen by the OS, which is less accurate than JVM in terms of developer feedback. Peak value is valuable though!

I've removed the current field so it only displays peak now.

src/com/xilinx/rapidwright/tests/CodePerfTracker.java
```diff
@@ -48,6 +52,8 @@ public class CodePerfTracker {
 
     private Map<String,Long> inflightTimes;
 
+    private List<Pair<Long,Long>> totalPeakMemUsages;
```
eddieh-xlnx, line 55: Since we're no longer tracking the current, do we still need to acquire and store `Pair`?

eddieh-xlnx: Generally, I disagree with adding dead code that cannot get exercised in normal operation and might even cause confusion down the line for anyone wanting to modify this code.

If you feel strongly, please leave a comment behind, and perhaps rename `totalPeakMemUsages` to `totalOSMemUsages` or similar.

src/com/xilinx/rapidwright/tests/CodePerfTracker.java
```diff
@@ -151,12 +162,35 @@ public CodePerfTracker stop() {
...
+        }
+
         if (printProgress && isVerbose()) {
             print(idx);
         }
         return this;
     }
 
+    /**
+     * Gets the current and total peak memory usage. Depends on /proc Linux to get
```
eddieh-xlnx, line 176:
```suggestion
     * Gets the current and total peak memory usage. Depends on Linux's /proc to get
```

eddieh-xlnx (approved):

src/com/xilinx/rapidwright/tests/CodePerfTracker.java
clavin-xlnx: Updated so the code is not dead, but can be turned on with a flag.

src/com/xilinx/rapidwright/tests/CodePerfTracker.java
```diff
@@ -151,12 +165,35 @@ public CodePerfTracker stop() {
...
 
+    /**
+     * Gets the current and total peak memory usage. Depends on Linux's /proc to get
+     * values.
+     * 
+     * @return A Pair where the first is current and second is peak memory usage.
+     */
+    private Pair<Long,Long> getTotalOSMemUsage() {
+        if (linuxProcID == null) return null;
+        Pair<Long,Long> totalPeakMemUsage = new Pair<>();
```
eddieh-xlnx, line 186: This needs renaming too

eddieh-xlnx (approved):

PR #537, opened 2022-09-16
Test from #426

clavin-xlnx:
#426 has been fixed in 2022.1.3, moving this test into its own PR.


clavin-xlnx:
I've incorrectly categorized this as fixed in 2022.1.3, it is not.  The fix is in #426.

PR #539, opened 2022-09-16
Fixes for copy cell to DCP

clavin-xlnx:
Fixes issues related to issues in `HardwareContractMaker` in #390.  This accommodates scenarios where the top level hierarchy object is being used.  Also fixes an issue when creating a stand-alone netlist from an existing cell was not updating the top cell reference appropriately.

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFTools.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFTools.java
@@ -980,12 +980,28 @@ public class EDIFTools {
 	 * @return The newly created netlist from the provided cell inst.
 	 */
 	public static EDIFNetlist createNewNetlist(EDIFCellInst cellInst){
+	    List<String> librariesOrder = new ArrayList<>(cellInst.getCellType().getNetlist().getLibrariesMap().keySet());
 	    EDIFNetlist n = new EDIFNetlist(cellInst.getName());
 	    n.generateBuildComments();
 	    EDIFDesign eDesign = new EDIFDesign(cellInst.getName());
 	    n.setDesign(eDesign);
-	    eDesign.setTopCell(cellInst.getCellType());
-	    n.migrateCellAndSubCells(cellInst.getCellType());
+	    EDIFCell topCell = cellInst.getCellType();
+	    n.copyCellAndSubCells(topCell);
+	    eDesign.setTopCell(n.getLibrary(topCell.getLibrary().getName()).getCell(topCell.getName()));
+	    if(n.getLibraries().size() > 2) {
+	        // Put libraries in the same order as source netlist
+	        Map<String, EDIFLibrary> libs = new HashMap<>();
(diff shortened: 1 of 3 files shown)


src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -980,12 +980,28 @@ public static EDIFNetlist createNewNetlist(String topName, boolean addINV){
 	 * @return The newly created netlist from the provided cell inst.
 	 */
 	public static EDIFNetlist createNewNetlist(EDIFCellInst cellInst){
+	    List<String> librariesOrder = new ArrayList<>(cellInst.getCellType().getNetlist().getLibrariesMap().keySet());
```
eddieh-xlnx, line 983: Am I missing where `cellInst.getCellType().getNetlist()` is modified, since I can see that it is copied into the netlist `n` but not where it would be modified?

src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -980,12 +980,28 @@ public static EDIFNetlist createNewNetlist(String topName, boolean addINV){
...
 	    EDIFNetlist n = new EDIFNetlist(cellInst.getName());
 	    n.generateBuildComments();
 	    EDIFDesign eDesign = new EDIFDesign(cellInst.getName());
 	    n.setDesign(eDesign);
-	    eDesign.setTopCell(cellInst.getCellType());
-	    n.migrateCellAndSubCells(cellInst.getCellType());
+	    EDIFCell topCell = cellInst.getCellType();
+	    n.copyCellAndSubCells(topCell);
+	    eDesign.setTopCell(n.getLibrary(topCell.getLibrary().getName()).getCell(topCell.getName()));
+	    if(n.getLibraries().size() > 2) {
```
eddieh-xlnx, line 991: `> 2` for work and primitives library? If so, let's leave a comment.

clavin-xlnx, line 991:
```suggestion
	    // If we have more than the primitives and work library, lets order the libraries in the same way 
	    // from the source netlist
	    if(n.getLibraries().size() > 2) {
```

src/com/xilinx/rapidwright/edif/EDIFTools.java
clavin-xlnx: Not sure why we want to modify the source netlist?  We are just copying from it.

eddieh-xlnx: Right. You copy this netlist's library order here, and my assumption was that you did this because you were worried it would change before/if you needed it later inside the `if` below?

clavin-xlnx: Thanks, I've updated it and the erroneous Javadoc that I think polluted my thinking.

eddieh-xlnx (approved):

PR #540, opened 2022-09-20
UltraScale Incremental Clock Router

clavin-xlnx:
Assumes the horizontal distribution lines are already in place, a common technique in DFX designs.


eddieh-xlnx:
Would it be straightforward to generate a test for this?

clavin-xlnx:
> Would it be straightforward to generate a test for this?

If we had a design that had the need, possibly yes.

eddieh-xlnx:
> If we have a design that had the need, possibly yes.

I was more alluding to -- do you recall what testcase you used to build this feature?

I have a bigger (and the original non-public) testcase that drove this work, but is it straightforward to manufacture a public, clock-specific testcase?

clavin-xlnx:
> I have a bigger (and the original non-public) testcase that drove this work, but is it straightforward to manufacture a public, clock-specific testcase?

The only existing DCP in test/RapidWrightDCP that comes close is the microblaze design, but it targets UltraScale. Because it is missing the leaf clock buffers, this code doesn't work for that design scenario.  I would have to retarget it to UltraScale+.

clavin-xlnx:
Add a test case for incremental clock routing on UltraScale+.

PR #541, opened 2022-09-20
Fixing whitespace and updating to AMD copyright

clavin-xlnx:


eddieh-xlnx:
Thanks for doing this Chris! Maybe this one-fell-swoop is an opportunity to fix another peeve of mine: that of inconsistent spacing.
I've noticed that (your?) IDE prefers to do:
```
if(...){
for(...){
void foo(){
etc.
```
whereas my IDE prefers a little less carbon-friendly (but arguably eyeball-friendly) with:
```
if (...) {
for (...) {
void foo() {
etc.
```
Can we converge on a style? We should synchronize on one style and use an IDE do an auto format on all of the codebase too. Maybe for now we can't/shouldn't easily enforce moving forward, but at least we should have eliminate ambiguity.

clavin-xlnx:
> Can we converge on a style? We should synchronize on one style and use an IDE do an auto format on all of the codebase too. Maybe for now we can't/shouldn't easily enforce moving forward, but at least we should have eliminate ambiguity.

Sure.  In recent years I have transitioned to the latter style with a space between `)` and `{`.  I think the examples in https://github.com/Xilinx/RapidWright/blob/master/CODING_STYLE.md prefer this style, but I don't think we legislate it.  Perhaps we should revisit this document.

clavin-xlnx:
@eddieh-xlnx - I fixed the tab check issue (had to resort to an actual tab character in the command).  All the license headers appear to be up to date and tabs are completely eliminated from the source.  I did clean up whitespace with respect to your comment.  I'm fairly satisfied with it now, so I'd like to move on this one sooner than later just so we don't cause a bunch of merge/conflict headaches later.  Have a look, but doesn't have to be today.

eddieh-xlnx (approved):

PR #542, opened 2022-09-23
Better manage GND/VCC routing of module instances during placement

clavin-xlnx:
When a module instance is placed, it may contain GND and VCC routing.  Because GND and VCC are global nets in a design, the subset of routing PIPs that are in the net need to be relocated during placement, but not the rest of the PIPs.  In the case of moving a module instance, the existing placed PIPs should be relocated (removed and then re-added with the proper offset) and leave all other existing GND/VCC PIPs in place.  This PR attempts to accomplish this.  This may diminish performance for re-placement heavy designs with modules that contain VCC and GND routing.

### DIFF
--- a/src/com/xilinx/rapidwright/design/ModuleInst.java
+++ b/src/com/xilinx/rapidwright/design/ModuleInst.java
@@ -315,7 +315,8 @@ public class ModuleInst extends AbstractModuleInst<Module, Site, ModuleInst>{
 
         // save original placement in case new placement is invalid
         HashMap<SiteInst, Site> originalSites;
-        originalSites = isPlaced() ? new HashMap<SiteInst, Site>() : null;
+        boolean placedPreviously = isPlaced();
+        originalSites = placedPreviously ? new HashMap<SiteInst, Site>() : null;
 
         //=======================================================//
         /* Place instances at new location                       */
@@ -376,7 +377,7 @@ public class ModuleInst extends AbstractModuleInst<Module, Site, ModuleInst>{
         /* Place net at new location                             */
         //=======================================================//
         nextnet: for (Net net : nets) {
-            unrouteNet(net);
+            unrouteNet(net, placedPreviously);
 
             Net templateNet = net.getModuleTemplateNet();
             Set<PIP> pipSet = getUsedStaticPIPs(templateNet);
@@ -417,21 +418,22 @@ public class ModuleInst extends AbstractModuleInst<Module, Site, ModuleInst>{
      * Removes all placement information and unroutes all nets of the module instance.
      */
     public void unplace() {
+        boolean placedPreviously = isPlaced();
         // unplace instances
         for (SiteInst inst : instances) {
             inst.unPlace();
         }
         // unplace nets (remove pips)
         for (Net net : nets) {
-            unrouteNet(net);
+            unrouteNet(net, placedPreviously);
         }
     }
 
-    private void unrouteNet(Net net) {
+    private void unrouteNet(Net net, boolean placedPreviously) {
         if (net.isStaticNet()) {
             // Any static nets that appear in 'nets' is not the exclusive responsibility
             // of this ModuleInst, instead it is shared with everything in 'design'.
-            if (isPlaced()) {
+            if (placedPreviously) {
                 // We need to remove the GND/VCC PIPs inserted by this ModuleInst out of the global design net
                 Net designNet = design.getNet(net.getName());
                 Set<PIP> prevUsed = getUsedStaticPIPs(designNet);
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -65,7 +65,10 @@ public class ModuleInst extends AbstractModuleInst<Module, Site, ModuleInst>{
     private ArrayList<Net> nets;
```
eddieh-xlnx, line 65: Not sure whether this should be in the javadoc:
```suggestion
    // Note that these are references to nets inside 'design' that this ModuleInst
    // inserted and is exclusively responsible for. As such, any static nets present in
    // this list must be handled with care.
    private ArrayList<Net> nets;
```

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -65,7 +65,10 @@ public class ModuleInst extends AbstractModuleInst<Module, Site, ModuleInst>{
...
     /** A list of all NOCClients belonging to this instance **/
     private ArrayList<NOCClient> nocClients;
-
+    /** Keeps track of currently placed GND PIPs */
+    private Set<PIP> gndPIPs;
+    /** Keeps track of currently placed VCC PIPs */
+    private Set<PIP> vccPIPs;
     /**
```
eddieh-xlnx, line 72: Does USED_NET pips need to be saved too?

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -369,8 +372,26 @@ public boolean place(Site newAnchorSite, boolean skipIncompatible, boolean allow
         /* Place net at new location                             */
         //=======================================================//
         nextnet: for (Net net : nets) {
-            net.getPIPs().clear();
+            if (net.isStaticNet()) {
+                if (isPlaced()) {
```
eddieh-xlnx, line 376:
```suggestion
                // Any static nets that appear in 'nets' is not the exclusive responsibility
                // of this ModuleInst, instead it is shared with everything in 'design'.
                if (isPlaced()) {
```

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -369,8 +372,26 @@ public boolean place(Site newAnchorSite, boolean skipIncompatible, boolean allow
...
+                    // We need to remove the old GND/VCC PIPs out of the main design nets
```
eddieh-xlnx, line 377:
```suggestion
                    // We need to remove the GND/VCC PIPs inserted by this ModuleInst out of the global design net
```

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -383,12 +404,13 @@ public boolean place(Site newAnchorSite, boolean skipIncompatible, boolean allow
...
                 newPip.setTile(newPipTile);
-                //if (!newPipTile.hasPIP(newPip)) {
-                //    return false;
-                //}
-                net.addPIP(newPip);
+                if (pipSet != null) {
+                    pipSet.add(newPip);
+                } else {
+                    net.addPIP(newPip);
+                }
```
eddieh-xlnx, line 413: Wouldn't you always want to insert this PIP into the global net...
```suggestion
                }
                net.addPIP(newPip);
```

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -397,7 +419,9 @@ public boolean place(Site newAnchorSite, boolean skipIncompatible, boolean allow
             // singleton net here
             if (templateNet.isStaticNet()) {
                 Net designNet = design.getNet(templateNet.getName());
-                designNet.getPIPs().addAll(net.getPIPs());
+                Set<PIP> newPipSet = new HashSet<>(designNet.getPIPs());
+                newPipSet.addAll(pipSet);
+                designNet.setPIPs(newPipSet);
```
eddieh-xlnx, line 424: ... instead of doing it here?
```suggestion
```

eddieh-xlnx (approved):

eddieh-xlnx:
Remember to clear `gndPIPs` and `vccPIPs` on unplace?

src/com/xilinx/rapidwright/design/ModuleInst.java
clavin-xlnx: Good thought, but it never contains any PIPs.  Usually it is limited to site routing usage.

eddieh-xlnx (approved):

PR #547, opened 2022-09-29
Changes EDIF port name collision behavior

clavin-xlnx:


src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -245,41 +240,45 @@ public EDIFPort addPort(EDIFPort port) {
         port.setParentCell(this);
         EDIFPort collision = ports.put(port.getBusName(), port);
         if (collision != null && port != collision) {
```
eddieh-xlnx, line 242: Assuming that `port` can't be `null`, I think this is sufficient:
```suggestion
        if (port != collision) {
```

src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -245,41 +240,45 @@ public EDIFPort addPort(EDIFPort port) {
...
     }
 
     /**
-     * Gets a port by bus name (see {@link EDIFPort#getBusName()}).  Multi-bit ports need to
-     * have brackets removed unless the {@link EDIFCell} already has a port
-     * with the same name as the bus name of the multi-bit port.  In only this case,
-     * the range would be required in order to distinguish the ambiguity.
-     * See {@link EDIFCell#addPort(EDIFPort)} for more information.
+     * Gets a port by bus name (see {@link EDIFPort#getBusName()}). Multi-bit ports
+     * need to have close square bracket and range removed (for example: "bus[3:0]"
```
eddieh-xlnx, line 252:
```suggestion
     * need to have closing square bracket and range removed (for example: "bus[3:0]"
```

src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -245,41 +240,45 @@ public EDIFPort addPort(EDIFPort port) {
...
+     * -> "bus[". See {@link EDIFCell#addPort(EDIFPort)} for more information.
      *
-     * @param name Bus name of the port to get.
+     * @param busName Bus name (ends with '[' to represent a bussed port) of the
+     *                port to get. Single bit ports use their entire name.
      * @return The port or null if none exists.
      */
-    public EDIFPort getPort(String name) {
+    public EDIFPort getPort(String busName) {
```
eddieh-xlnx, line 259: Param here being named `busName` implies that it only works for busses. That's not true, right?

src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -245,41 +240,45 @@ public EDIFPort addPort(EDIFPort port) {
...
         if (ports == null) return null;
-        return ports.get(name);
+        EDIFPort port = ports.get(busName);
+        if (port == null && busName.charAt(busName.length() - 1) != '[') {
+            port = ports.get(busName + "[");
```
eddieh-xlnx, line 263: While it's fresh, can I get a comment here explaining what condition would cause the search to be retried with a `[` suffix if one didn't exist? Can we expect that our own RapidWright code would be careful enough to preserve the `[` so is this for non-library code?

src/com/xilinx/rapidwright/edif/EDIFNet.java
```diff
@@ -104,22 +104,59 @@ public EDIFPortInst createPortInst(EDIFPort port, int index) {
...
+    /**
+     * Creates a port instance from a name. Navigates port naming issues when bussed
+     * names can collide with single bit port names.
+     * 
+     * @param portInstName Proposed name of the new port instance
+     * @param cell         The cell from which to draw the port
+     * @param inst         If this is not null, the port instance is added to the
+     *                     external facing port connection. If this is null, it will
+     *                     add it to the inward facing port connection.
+     * @return The newly create port instance or null if none could be created on
```
eddieh-xlnx, line 148:
```suggestion
     * @return The newly created port instance or null if none could be created on
```

src/com/xilinx/rapidwright/edif/EDIFNet.java
```diff
@@ -104,22 +104,59 @@ public EDIFPortInst createPortInst(EDIFPort port, int index) {
...
+    /**
+     * Creates a new port instance from a name on the internal port of the provided
+     * cell. It looks up the appropriate port name from the portInstName and
+     * identifies the index if any.
+     * 
+     * @param portInstName The name of the new port instance, including indexed bit
+     *                     if it belongs on a bussed port.
+     * @param cell         The destination cell to receive the port instance (on an
+     *                     inward facing port).
+     * @return The newly create port instance or null if none could be created on
```
eddieh-xlnx, line 132:
```suggestion
     * @return The newly created port instance or null if none could be created on
```

src/com/xilinx/rapidwright/edif/EDIFNet.java
```diff
@@ -104,22 +104,59 @@ public EDIFPortInst createPortInst(EDIFPort port, int index) {
...
+    /**
+     * Creates a new port instance from a name on the external port of the provided
+     * cell instance. It looks up the appropriate port name from the portInstName
+     * and identifies the index if any.
+     * 
+     * @param portInstName The name of the new port instance, including indexed bit
+     *                     if it belongs on a bussed port.
+     * @param cellInst     The destination cell instance to receive the port
+     *                     instance
+     * @return The newly create port instance or null if none could be created on
```
eddieh-xlnx, line 116:
```suggestion
     * @return The newly created port instance or null if none could be created on
```

src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -382,14 +382,29 @@ public static String makeNameEDIFCompatible(String currName) {
...
-     * Strips off bracket index in a bussed name (ex: {@code "data[0]" --> "data"}).
+     * Strips off bracket index in a bussed name (ex:
+     * {@code "data[0]" --> "data["}). Leaves open bracket by default as this is the
+     * key for bussed ports in the port map of {@link EDIFCell}
+     * 
      * @param name Bracketed bussed name.
-     * @return Name of bus with brackets removed
+     * @return Name of bus with index and close bracket removed
      */
     public static String getRootBusName(String name) {
```
eddieh-xlnx, line 392: This method looks similar to `getStemName()`. Perhaps we can deprecate this (it doesn't appear to be used in RapidWright, but is used once elsewhere) while we're here?

src/com/xilinx/rapidwright/edif/compare/EDIFNetlistComparator.java
```diff
@@ -0,0 +1,416 @@
...
+        diffs.add(new EDIFDiff(type, gold, test, parentCell, parentLibrary, notEqualString));
+        diffCount++;
+    }
+
+    public int compareNetlists(EDIFNetlist gold, EDIFNetlist test) {
+        diffMap = new LinkedHashMap<>();
+//        check(gold.getName(), test.getName(), "EDIFNetlist.getName()");
+//        check(gold.getDesign().getName(), test.getDesign().getName(), "EDIFNetlist.getDesign().getName()");
+//        equivalentEDIFPropObject(gold.getDesign(), test.getDesign());
+//        check(gold.getLibraries().size(), test.getLibraries().size(), "EDIFNetlist.getLibraries().size()");
```
eddieh-xlnx, line 306: Valuable?

src/com/xilinx/rapidwright/util/FileTools.java
```diff
@@ -1783,9 +1787,67 @@ public static boolean overrideDataFileDownload() {
     }
 
     /**
-     * For Java 16 and below, calling this method will prevent System.exit() calls from
-     * exiting the JVM and instead throws a {@link SecurityException} in its place.  This method
-     * allows for a check to avoid the JVM WARNING message in Java 17.
+     * Gets all files (Path objects) recursively (including all sub-directories)
```
eddieh-xlnx, line 1790: Needs master merging in?

test/src/com/xilinx/rapidwright/edif/TestEDIFPort.java
```diff
@@ -66,4 +67,49 @@ public void testEDIFPortInternalNets() {
             }
         }
     }
+
+    @Test
+    public void testCreatePort() {
```
eddieh-xlnx, line 72: Does this test all the different collision scenarios we can think of? If so, can you enumerate them as a comment so we don't forget about them?

src/com/xilinx/rapidwright/edif/EDIFCell.java
clavin-xlnx: Correct, if you think it is less confusing, we can return it to being `name`.

src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -245,41 +240,45 @@ public EDIFPort addPort(EDIFPort port) {
...
+     * @param busName Bus name (ends with '[' to represent a bussed port) of the
+     *                port to get. Single bit ports use their entire name.
      * @return The port or null if none exists.
      */
-    public EDIFPort getPort(String name) {
+    public EDIFPort getPort(String busName) {
         if (ports == null) return null;
-        return ports.get(name);
+        EDIFPort port = ports.get(busName);
+        if (port == null && busName.charAt(busName.length() - 1) != '[') {
```
clavin-xlnx, line 262:
```suggestion
        // For callers who have a port name and its unknown if its a bus, attempt a check with adding the '[' suffix
        if (port == null && busName.charAt(busName.length() - 1) != '[') {
```

src/com/xilinx/rapidwright/edif/EDIFCell.java
clavin-xlnx: There are some scenarios where it is unknown.  If the caller passes a name with the `[` suffix there, then we avoid an extra lookup.  Of course, I suppose that would not give signal names such as `bus[[3:0]` a second chance.

src/com/xilinx/rapidwright/edif/EDIFTools.java
clavin-xlnx: Which method do you mean to deprecate (or both)?

src/com/xilinx/rapidwright/util/FileTools.java
clavin-xlnx: Thanks

src/com/xilinx/rapidwright/edif/EDIFTools.java
eddieh-xlnx: Sorry, it looks like I'm confusing `EDIFPort.getStemName()` and `EDIFPort.getBusName()` (what is the difference between these two?) with `EDIFTools.getRootBusName()`.

I'm okay with it if you feel like this out of scope of this PR, but the synonym-y nature of all three is a bit confusing.

test/src/com/xilinx/rapidwright/edif/TestEDIFPort.java
clavin-xlnx: I think with the convention change of adding the `[` suffix to all bussed ports, it alleviates the previous collision challenges.  However, I can imagine that we now have theoretical issues with port names such as `single_bit_port[` and `single_bit_port[[3:0]`.  Do we want to build out support for these?

eddieh-xlnx: > However, I can imagine that we now have theoretical issues with port names such as `single_bit_port[` and `single_bit_port[[3:0]`. Do we want to build out support for these?

Absolutely not! 

I just couldn't figure out if this test tests any collisions, so maybe a quick comment to state which collisions are being tested. Off the top of my head:

`foo` vs `foo[0]` -- both bitty
`foo[0]` vs `foo[1]` -- both bitty
`foo[0]` vs `foo[0][0]` -- both bitty
`foo[0]` vs `foo[0][7:0]` -- one bitty one bussy

Anything else? Particularly the scenario that the current implementation already resolves, plus the new scenario that you discovered recently which necessitated this rewrite.

src/com/xilinx/rapidwright/edif/EDIFTools.java
clavin-xlnx: Deprecating `EDIFPort.getStemName()`

test/src/com/xilinx/rapidwright/edif/TestEDIFPort.java
clavin-xlnx: I don't think there are really any others that we want to test for.  I've updated the test to enumerate and test for these.

src/com/xilinx/rapidwright/edif/EDIFCell.java
eddieh-xlnx: Yes, let's do that.

clavin-xlnx: Done

src/com/xilinx/rapidwright/edif/compare/EDIFNetlistComparator.java
clavin-xlnx, line 306:
```suggestion
```

eddieh-xlnx (approved):

PR #550, opened 2022-09-30
Test that Node.getAll{Uphill,Dowhill}Nodes() are unique

eddieh-xlnx:
Signed-off-by: Eddie Hung <eddie.hung@amd.com>

### DIFF
--- a/test/src/com/xilinx/rapidwright/device/TestNode.java
+++ b/test/src/com/xilinx/rapidwright/device/TestNode.java
@@ -1,3 +1,25 @@
+/*
+ * Copyright (c) 2022, Advanced Micro Devices, Inc.
+ * All rights reserved.
+ *
+ * Author: Eddie Hung, Advanced Micro Devices, Inc.
+ *
+ * This file is part of RapidWright.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


clavin-xlnx:
After some investigation, it was found that (on at least a VU19P) the duplicates that are present occur very rarely (0.02%),  Switching from a `List` based data structure to a `Set` based data structure also comes with a performance penalty of ~46% slowdown which is not really worth removing the duplicates for a such a small subset.  Infrastructure could be augmented to capture the set of nodes that exhibit this behavior and only use a set in those situations, but this will be left to future work.  

To provide a meaningful alternative, we'll overload the `Node.getAll{Uphill,Downhill}Nodes()` APIs with versions that take as input a `Collection<Node>` object type so the caller can choose the desired behavior based on the data structure's underlying fundamentals.

test/src/com/xilinx/rapidwright/device/TestNode.java
```diff
@@ -39,10 +40,10 @@ public void testGetDownhillUphillNodesUnique(String deviceName, String nodeName)
         Device device = Device.getDevice(deviceName);
         Node node = device.getNode(nodeName);
 
-        List<Node> downhill = node.getAllDownhillNodes();
```
eddieh-xlnx: Can we continue to test for non-unique-ness using the existing API so that it'll stick out if anything changes?

clavin-xlnx: Sure, I've updated the test to check both APIs.

eddieh-xlnx: Looks good (I can't approve my own PR)

clavin-xlnx (approved):

Issue #558, opened 2022-10-12
RWRoute RouteHelper fails to find the INT tile entry node from a BITSLICE_CONTROL site pin


clavin-xlnx:
The method `RouteHelper.projectOutputPinToINTNode()` has not been tested in RWRoute for `BITSLICE_CONTROL` site pins and does not find the INT tile entry node for the beginning of a route.  For example, the site pin `TX_T_OUT` when queried, the method returns `null` when it should return the node `INT_INTF_L_CMT_X182Y90/LOGIC_OUTS_R19`.  In Vivado, we can see that the INT entry node (shown in green) does exist:

![image](https://user-images.githubusercontent.com/20803522/195468792-f7a030a7-ca86-4ba1-919e-16139dd713e1.png)

The method needs to account for the pip connection between the two nodes as shown in blue.

EDIT: For future contributors who will work on this -- a test demonstrating this failure is at https://github.com/Xilinx/RapidWright/pull/559

PR #559, opened 2022-10-13
Known failing test for BITSLICE_CONTROL output pin projection

clavin-xlnx:
Demonstrates the failure described in #558.


eddieh-xlnx:
@clavin-xlnx Can't tag you for review since you opened this PR but please take a look.

PR #566, opened 2022-10-31
Test for the fix of Cell.getCorrespondingSitePinName() 

clavin-xlnx:


clavin-xlnx:
> Ack! It looks like the diff is really big. Does some merging need to happen?

Sorry, I forgot you have to "turn it off and then on again" with the target branch.

eddieh-xlnx (approved):

PR #567, opened 2022-10-31
Fix DesignTools.createMissingStaticSitePins() when only LUT6

eddieh-xlnx:
... to not connect A6 to VCC unless there is a corresponding [A-H]5LUT cell placed.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -3030,7 +3030,7 @@ public class DesignTools {
                 }
             }
             String belName = belPin.getBELName();
-            if (DesignTools.isBELALut(belName) && si.getCell(belName.replace('6', '5')) == null) {
+            if (bel.isLUT() && si.getCell(belName.replace('6', '5')) == null) {
                 // Nothing is placed on the 5LUT BEL, no need for site pin
                 return;
             }
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -2939,7 +2939,7 @@ public class DesignTools {
                 if (cell.getName().contains("LOCKED")) continue;// Are there better ways to identify this problem?
 
                 BEL bel = cell.getBEL();
-                if (bel == null || !bel.getName().contains("LUT")) continue;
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3029,6 +3029,11 @@ public static void createMissingStaticSitePins(BELPin belPin, SiteInst si, Cell
                     net = si.getDesign().getGndNet();
                 }
             }
+            String belName = belPin.getBELName();
+            if (DesignTools.isBELALut(belName) && si.getCell(belName.replace('6', '5')) == null) {
```
clavin-xlnx, line 3033: Should we use `belPin.getBEL().isLUT()` over `DesignTools.isBELALut()`?

clavin-xlnx (approved):

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: Should we consider deprecating the latter?

eddieh-xlnx, line 3033:
```suggestion
            String belName = belPin.getBELName();
            if (bel.isLUT() && si.getCell(belName.replace('6', '5')) == null) {
```

clavin-xlnx: I don't think the `bel` variable exists.

eddieh-xlnx: Let's leave it then.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3029,6 +3029,11 @@ public static void createMissingStaticSitePins(BELPin belPin, SiteInst si, Cell
                     net = si.getDesign().getGndNet();
                 }
             }
+            String belName = belPin.getBELName();
+            if (bel.isLUT() && si.getCell(belName.replace('6', '5')) == null) {
```
eddieh-xlnx, line 3033:
```suggestion
            if (belPin.getBEL().isLUT() && si.getCell(belName.replace('6', '5')) == null) {
```

Issue #572, opened 2022-11-03
[Site with dual output site pins (COUT, HQ2)] concurrent.ExecutionException at writeCheckpoint


nqdtan:
I attach the zip file that contains the DCP for reproducing the issue here. Below is my RapidWright code

```java
public class TestCmp {
  public static void main(String[] args) {
    Design design = Design.readCheckpoint("test_cmp.dcp", "test_cmp.edf");
    design.writeCheckpoint("test_cmp_rw.dcp");
  }
}
```

I got the following error

> Exception in thread "main" java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.NullPointerException
	at com.xilinx.rapidwright.util.ParallelismTools.joinFirst(ParallelismTools.java:246)
	at b.b(Unknown Source)
	at b.a(Unknown Source)
	at com.xilinx.rapidwright.design.Design.writeCheckpoint(Unknown Source)
	at com.xilinx.rapidwright.examples.TestCmp.main(TestCmp.java:31)
Caused by: java.util.concurrent.ExecutionException: java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.NullPointerException
	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.util.concurrent.FutureTask.get(FutureTask.java:192)
	at com.xilinx.rapidwright.util.ParallelismTools.joinFirst(ParallelismTools.java:244)
	... 4 more
...

When trying to isolate the problem, I suspect the error might have something to do with a net on which there are two output Site pins coming from the same SLICE. My workaround is replacing one of the PIPs connecting to these Site pins with a fresh one created by the `getPIP()` method from a Device object.

```java
public class TestCmp {                                                              
  public static void main(String[] args) {                                          
    Design design = Design.readCheckpoint("test_cmp.dcp", "test_cmp.edf");
    Device dev = design.getDevice();  

    PIP toRemovePIP = null;                                                         
    Net cmpNet = design.getNet("cmp");                                              
    for (PIP pip : cmpNet.getPIPs()) {                                              
[...]

PR #573, opened 2022-11-03
Test for #572

clavin-xlnx:


eddieh-xlnx (approved):
Just need to update submodule ref to master and good to go!

PR #575, opened 2022-11-09
Rewrite unroute source/sink pin methods in DesignTools

eddieh-xlnx:
Specifically rewriting: `DesignTools.getTrimmablePIPsFromPins()` and modifying `DesignTools.unrouteSourcePin()`.
Also removes unused private method `DesignTools.unroutePin()`

Since it wasn't handling the two new testcases with bidirectional PIPs.

In this rewrite, use `PIP.isReversed()` to get the correct driving/sink node.

### DIFF
--- a/test/src/com/xilinx/rapidwright/design/TestDesignTools.java
+++ b/test/src/com/xilinx/rapidwright/design/TestDesignTools.java
@@ -479,6 +479,44 @@ public class TestDesignTools {
         }
     }
 
+    @Test
+    public void testUnrouteSourcePinBidir() {
+        Design design = new Design("test", "xcvu19p-fsva3824-1-e");
+
+        Net net = createTestNet(design, "net", new String[]{
+                "INT_X193Y606/INT.LOGIC_OUTS_W27->INT_NODE_SDQ_87_INT_OUT0",
+                "INT_X193Y606/INT.INT_NODE_SDQ_87_INT_OUT0->>NN1_W_BEG6",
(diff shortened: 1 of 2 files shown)


test/src/com/xilinx/rapidwright/design/TestDesignTools.java
```diff
@@ -309,6 +311,11 @@ public void testGetTrimmablePIPsFromPinsBidir(boolean unrouteAll) {
                 "INT_X102Y429/INT.INODE_W_BLN_60_FT1->>IMUX_W2",                // E1 input
         });
 
+        for (PIP pip : net.getPIPs()) {
+            if (pip.toString().equals("INT_X102Y428/INT.INT_NODE_IMUX_50_INT_OUT0<<->>BYPASS_W14"))
+                pip.setIsReversed(true);
```
clavin-xlnx, line 316: I guess we are relying on the source of the PIPs to set this flag correctly for this change to work.  Will that always be true?

eddieh-xlnx: It looks like that, if it came directly from a Vivado DCP, then this flag should be set correctly. In other cases (e.g. when a user creates a PIP) then it is incumbent on them to set the flag.

A future enhancement would provide a helper method that can analyze a list of PIPs and update such flags.

clavin-xlnx: I have filed #579 and #580 to help us track this issue and ensure this doesn't become a problem.

clavin-xlnx (approved):

PR #577, opened 2022-11-10
Known failing test for EDIFHierPortInst.getRoutedSitePinInst()

eddieh-xlnx:
Of the five testcases in `TestEDIFHierPortInst.testGetRoutedSitePinInst()`, the last situation:
```
            // Cell pin placed onto a E5LUT/O5 -- its net does not exit the site but
            // another net is using EMUX
            "processor/stack_loop[4].upper_stack.stack_pointer_lut/LUT5/O,null",
```
fails because it returns a SPI belonging to some other net.

There appears to be a similar method: `DesignTools.getRoutedSitePin()` which appears to behave correctly for all five testcases (except it returns a site pin name, and not a `SitePinInst` object).

### DIFF
--- a/test/src/com/xilinx/rapidwright/design/TestDesignTools.java
+++ b/test/src/com/xilinx/rapidwright/design/TestDesignTools.java
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2021-2022, Xilinx, Inc.
- * Copyright (c) 2022-2023, Advanced Micro Devices, Inc.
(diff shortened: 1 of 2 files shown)


eddieh-xlnx:
Since there are two methods:
* `EDIFHierPortInst.getRoutedSitePinInst()`
* `DesignTools.getRoutedSitePin()`
that take different approaches to get very similar information, is it worth converging onto just one approach?

clavin-xlnx (approved):

PR #587, opened 2022-11-30
Memory Reduction for RWRoute

clavin-xlnx:
RouteNode memory reduction: 
1) Set HashMap init size to 1; 
2) Removes baseCost (compute on-the-fly); 
3) Combine isTarget flag and RouteNodeType into a single variable

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/RouteNode.java
+++ b/src/com/xilinx/rapidwright/rwroute/RouteNode.java
@@ -51,17 +51,17 @@ abstract public class RouteNode implements Comparable<RouteNode> {
 
     /** The associated {@link Node} instance */
     protected Node node;
-    /** The type of a rnode*/
-    private RouteNodeType type;
     /** The tileXCoordinate and tileYCoordinate of the INT tile that the associated node stops at */
     private short endTileXCoordinate;
     private short endTileYCoordinate;
     /** The wirelength of a rnode */
     private short length;
-    /** The base cost of a rnode */
-    private float baseCost;
-    /** A flag to indicate if this rnode is the target */
-    private boolean isTarget;
+    /** Bit mask used for storing the target flag */
+    private static final int IS_TARGET_MASK = 0x80;
+    /** Bit mask to store RouteNodeType ordinal */
+    private static final int ROUTE_NODE_TYPE_MASK = 0x07;
+    /** Stores a flag to indicate that this node is the target and also node type */
+    private byte isTargetAndType;
     /** The children (downhill rnodes) of this rnode */
     protected RouteNode[] children;
 
@@ -95,13 +95,12 @@ abstract public class RouteNode implements Comparable<RouteNode> {
         setType(type);
         children = null;
         setEndTileXYCoordinates();
-        setBaseCost(type);
         presentCongestionCost = 1;
         historicalCongestionCost = 1;
         usersConnectionCounts = null;
         driversCounts = null;
         assert(prev == null);
-        assert(!isTarget);
+        assert(!isTarget());
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/rwroute/RouteNode.java
```diff
@@ -451,47 +453,48 @@ public void resetChildren() {
...
     private void setType(RouteNodeType type) {
-        this.type = type;
+        RouteNodeType typeToSet = type;
         if (type == RouteNodeType.WIRE) {
             // NOTE: IntentCode is device-dependent
             IntentCode ic = node.getIntentCode();
             switch (ic) {
                 case NODE_PINBOUNCE:
-                    this.type = RouteNodeType.PINBOUNCE;
+                    typeToSet = RouteNodeType.PINBOUNCE;
```
eddieh-xlnx, line 462: I would get rid of `typeToSet` entirely and overwrite `type` for clarity. It looks like we don't read from `type` except in the very first `if`?

src/com/xilinx/rapidwright/rwroute/RouteNode.java
```diff
@@ -51,17 +51,17 @@ abstract public class RouteNode implements Comparable<RouteNode> {
 
     /** The associated {@link Node} instance */
     protected Node node;
-    /** The type of a rnode*/
-    private RouteNodeType type;
     /** The tileXCoordinate and tileYCoordinate of the INT tile that the associated node stops at */
     private short endTileXCoordinate;
     private short endTileYCoordinate;
     /** The wirelength of a rnode */
     private short length;
```
eddieh-xlnx, line 58: Is there any gain in making `length` a byte instead of a `short`? More generally, I think the longest (non clock) wire we have is a SLL which is 60 tiles. I'm unclear as to where there are a longer clock wires, but in any case, I don't believe clock routing uses the `RouteNode` class anyway.

If you do reduce it to a `byte`, let's add an assertion before setting it to ensure this invariant doesn't get tripped in the future.

src/com/xilinx/rapidwright/rwroute/RouteNode.java
```diff
@@ -432,7 +371,70 @@ public short getEndTileYCoordinate() {
      * @return The base cost of a RouteNode Object.
      */
     public float getBaseCost() {
-        return baseCost;
+        float _baseCost = 0;
```
eddieh-xlnx, line 374: I would be curious to benchmark this properly -- we're potentially adding multiple branches, `getIntentCode()`, and floating point multiply to this.

Another approach I can think of is to have a `float[][]` array where the first dimension is `TileTypeEnum` and the second is `wireIndex` in order to get the base cost. Assuming the cost truly is the same for all tile types. Perhaps the same could be done for `length` too, assuming the same invariant applies.

eddieh-xlnx (commented):
Another value we can trade memory for runtime is could be `presentCongestionCost` as that solely depends on occupancy and `config.getPresentCongestionMultiplier()`?

Other low hanging fruit would be to get rid of the `driverCounts` reference. This member is only ever used during post-routing fix up, and shouldn't be carried all the way through the most expensive part of routing. This requires refactoring the post-route fix up code too though, so maybe in a different PR?

For completeness, another opt is to get rid of `lowerBoundTotalPathCost` from the node. This is only ever used when in the priority queue (for comparator use); it's of no consequence once the rnode has been popped.

src/com/xilinx/rapidwright/rwroute/RouteNode.java
clavin-xlnx: In the follow up to this optimization, I've found that this this method is called less than once per node pushed.  
```
Nodes pushed:                        605400311
Nodes popped:                        278681998
getBaseCost() calls:                 604688664
```
I'm open to optimizing it further.  For example, `node.getAllDownhillNodes().isEmpty()` can be precomputed offline and cached.

clavin-xlnx:
Revisit later in the future.

PR #589, opened 2022-12-07
EDIF -> LogicalNetlist converter

kboronski-ant:
There's an easy way to convert LogicalNetlist to EDIF (https://github.com/Xilinx/RapidWright/pull/565), but there's no easy way to do it the other way around.

This PR adds a capability to do that.


eddieh-xlnx:
Can you merge in/rebase on latest master please? #591 is now in which should make CI green again.

PR #592, opened 2023-01-03
Add support to parse gzipped EDIF files

clavin-xlnx:
This will allow the EDIF Parser (both serial and parallel) to accept gzipped EDIF files for parsing.  It adds this capability by checking the file extension, if it ends with `.gz`, it is assumed gzip compressed and will temporarily decompress it in the same directory as the original file and parse it.  Once parsing is completed, the temporary decompressed EDIF file is deleted.  This will overwrite an existing EDIF file if the provided `.edf.gz` file is given for parsing.

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFTools.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFTools.java
@@ -755,7 +755,7 @@ public class EDIFTools {
     public static EDIFNetlist loadEDIFFile(Path fileName, int maxThreads) {
         try {
             final long size = Files.size(fileName);
-            if (ParallelEDIFParser.calcThreads(size, maxThreads) > 1) {
+            if (ParallelEDIFParser.calcThreads(size, maxThreads, fileName.toString().endsWith(".gz")) > 1) {
                 try (ParallelEDIFParser p = new ParallelEDIFParser(fileName)) {
                     return p.parseEDIFNetlist();
                 }
--- a/src/com/xilinx/rapidwright/edif/ParallelEDIFParser.java
+++ b/src/com/xilinx/rapidwright/edif/ParallelEDIFParser.java
@@ -40,6 +40,7 @@ import java.util.stream.Collectors;
 
 import com.xilinx.rapidwright.device.Device;
 import com.xilinx.rapidwright.tests.CodePerfTracker;
+import com.xilinx.rapidwright.util.FileTools;
 import com.xilinx.rapidwright.util.ParallelismTools;
 import com.xilinx.rapidwright.util.StringPool;
 import com.xilinx.rapidwright.util.function.InputStreamSupplier;
@@ -86,15 +87,16 @@ public class ParallelEDIFParser implements AutoCloseable{
         return new ParallelEDIFParserWorker(fileName, inputStreamSupplier.get(), offset, uniquifier, maxTokenLength, cache);
     }
 
-    public static int calcThreads(long fileSize, int maxThreads) {
-        int maxUsefulThreads = Math.max((int) (fileSize / MIN_BYTES_PER_THREAD),1);
+    public static int calcThreads(long fileSize, int maxThreads, boolean isGzipped) {
+        long sizeThreshold = isGzipped ? (MIN_BYTES_PER_THREAD >> 4) : MIN_BYTES_PER_THREAD;
+        int maxUsefulThreads = Math.max((int) (fileSize / sizeThreshold), 1);
         return Math.min(maxUsefulThreads, Math.min(ParallelismTools.maxParallelism(), maxThreads));
     }
 
 
     protected void initializeWorkers() throws IOException {
         workers.clear();
-        int threads = calcThreads(fileSize, maxThreads);
+        int threads = calcThreads(fileSize, maxThreads, fileName.toString().endsWith(".gz"));
         long offsetPerThread = fileSize / threads;
         for (int i=0;i<threads;i++) {
             ParallelEDIFParserWorker worker = makeWorker(i*offsetPerThread);
@@ -105,7 +107,11 @@ public class ParallelEDIFParser implements AutoCloseable{
     private int numberOfThreads;
 
     public EDIFNetlist parseEDIFNetlist() throws IOException {
-        return parseEDIFNetlist(CodePerfTracker.SILENT);
+        EDIFNetlist netlist = parseEDIFNetlist(CodePerfTracker.SILENT);
+        if (this.fileName.toString().endsWith(".gz")) {
+            Files.delete(FileTools.getDecompressedGZIPFileName(fileName));
+        }
+        return netlist;
     }
 
     public EDIFNetlist parseEDIFNetlist(CodePerfTracker t) throws IOException {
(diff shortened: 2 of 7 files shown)


clavin-xlnx:
@eddieh-xlnx - Should we support gzipped files that don't end in `.gz`? Currently, this implementation does not.

clavin-xlnx:
Sorry @eddieh-xlnx , just found a performance bug, will mark as ready for review when fixed.

clavin-xlnx:
Some performance numbers:

image

src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -755,7 +755,7 @@ public static EDIFNetlist loadEDIFFile(Path fileName) {
     public static EDIFNetlist loadEDIFFile(Path fileName, int maxThreads) {
         try {
             final long size = Files.size(fileName);
-            if (ParallelEDIFParser.calcThreads(size, maxThreads) > 1) {
+            if (ParallelEDIFParser.calcThreads(size, maxThreads, fileName.toString().endsWith(".gz")) > 1) {
```
eddieh-xlnx, line 758:
```suggestion
            long estimatedUncompressedSize = size * EDIF_GZIP_RATIO; 
            if (ParallelEDIFParser.calcThreads(estimatedUncompressedSize, maxThreads) > 1) {
```
(After moving the `EDIF_GZIP_RATIO` member over from `ParallelEDIFParser`)
Could this be more meaningful/generic?

src/com/xilinx/rapidwright/edif/ParallelEDIFParser.java
```diff
@@ -105,7 +110,11 @@ protected void initializeWorkers() throws IOException {
     private int numberOfThreads;
 
     public EDIFNetlist parseEDIFNetlist() throws IOException {
-        return parseEDIFNetlist(CodePerfTracker.SILENT);
+        EDIFNetlist netlist = parseEDIFNetlist(CodePerfTracker.SILENT);
+        if (this.fileName.toString().endsWith(".gz")) {
```
eddieh-xlnx, line 114:
```suggestion
        if (fileName.toString().endsWith(".gz")) {
```

src/com/xilinx/rapidwright/util/FileTools.java
```diff
@@ -1832,5 +1833,62 @@ public static void main(String[] args) {
...
+     */
+    public static Path decompressGZIPFile(Path gzipFile) {
+        String fileNameStr = gzipFile.toString();
+        if (!fileNameStr.endsWith(".gz")) return null;
+        Path target = FileTools.getDecompressedGZIPFileName(gzipFile);
+        // Using a larger buffer size for GZIPInputStream improved runtime 5-10%
+        try (GZIPInputStream gis = new GZIPInputStream(new FileInputStream(fileNameStr), 65536)) {
+            Files.copy(gis, target, StandardCopyOption.REPLACE_EXISTING);
+        } catch (FileNotFoundException e) {
+            throw new UncheckedIOException(e);
```
eddieh-xlnx, line 1855:
```suggestion
```
`FileNotFoundException` is a subclass of `IOException`

src/com/xilinx/rapidwright/util/FileTools.java
```diff
@@ -1832,5 +1833,62 @@ public static void main(String[] args) {
...
+     * Compresses the provided file using GZIP (adds '.gz' extension) 
+     * @param uncompressedFile The path to the uncompressed file
+     * @return The path to the compressed file which has a '.gz' extension
+     */
+    public static Path compressFileUsingGZIP(Path uncompressedFile) {
+        Path compressedFile = Paths.get(uncompressedFile.toString() + ".gz");
+        try (GZIPOutputStream gos = new GZIPOutputStream(new FileOutputStream(compressedFile.toFile()), 65536)) {
+            Files.copy(uncompressedFile, gos);
+        } catch (FileNotFoundException e) {
+            throw new UncheckedIOException(e);
```
eddieh-xlnx, line 1872:
```suggestion
```

test/src/com/xilinx/rapidwright/edif/TestEDIFParser.java
```diff
@@ -119,4 +122,30 @@ public void loadEDIFSingleThreaded() throws IOException {
...
+
+    @Test
+    public void testGZIPEDIFParsing(@TempDir Path tempDir) throws IOException {
+        Path src = tempDir.resolve(input.getFileName());
+        Files.copy(input, src);
+        Path compressed = FileTools.compressFileUsingGZIP(src);
+        Assertions.assertTrue(compressed.toString().endsWith(".gz"));
+        Assertions.assertTrue(src.toFile().exists());
+        Files.delete(src);
+        EDIFTools.loadEDIFFile(compressed);
```
eddieh-xlnx, line 134:
```suggestion
        Assertions.assertFalse(src.toFile().exists());
        Assertions.assertNotNull(EDIFTools.loadEDIFFile(compressed));
```

test/src/com/xilinx/rapidwright/edif/TestEDIFParser.java
```diff
@@ -119,4 +122,30 @@ public void loadEDIFSingleThreaded() throws IOException {
...
+    public void testGZIPEDIFParsingParallel(@TempDir Path tempDir) throws IOException {
+        Path src = tempDir.resolve(input.getFileName());
+        Files.copy(input, src);
+        Path compressed = FileTools.compressFileUsingGZIP(src);
+        Assertions.assertTrue(compressed.toString().endsWith(".gz"));
+        Assertions.assertTrue(src.toFile().exists());
+        Files.delete(src);
+        try (ParallelEDIFParser p = new ParallelEDIFParser(compressed)) {
+            p.parseEDIFNetlist();
+        }
```
eddieh-xlnx, line 148:
```suggestion
        Assertions.assertFalse(src.toFile().exists());
        try (ParallelEDIFParser p = new ParallelEDIFParser(compressed)) {
            Assertions.assertNotNull(p.parseEDIFNetlist());
        }
```

src/com/xilinx/rapidwright/edif/EDIFTools.java
clavin-xlnx: Not sure that this works in the uncompressed case.

eddieh-xlnx: Oh yes, sorry! What about gating the inflation using the `.gz` check? My overall suggestion was that we can re-inflate the size back to the uncompressed value before calling the EDIF parser.

clavin-xlnx: I don't think I understand the motivation of why this code needs to be modified.  The only thing this check does it decide between parallel parsing and serial parsing of the file based on size.

eddieh-xlnx: To me, it's a little easier to understand/general if we always provide the uncompressed size to a method than to have something specific happen inside the parser. Scenarios like what if the exact size of the file becomes known, or if we have to support some other compression standard, becomes less intrusive to support. Only a suggestion, not anything to hold up this PR. Let's move on.

PR #595, opened 2023-01-06
Add DesignTools.isNetDrivenByHierPort() method

eddieh-xlnx:
> Determine if a Net is driven by a hierarchical port, created as part of an out-of-context synthesis flow, for example.

It is important for a router to identify such nets, and to reserve their site pin nodes, lest those nodes are claimed during the routing of other nets. When this happens, Vivado will flag it up as a conflict.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -2891,7 +2891,11 @@ public class DesignTools {
             }
             String belName = belPin.getBELName();
             if (LUTTools.isCellALUT(cell) &&
+                    // No cell placed in the 5LUT spot
+                    si.getCell(belName.replace('6', '5')) == null &&
+                    // No net originally present on input sitewire
                     netOnSiteWire != net &&
+                    // No net present on output sitewire
                     si.getNetFromSiteWire(belName.charAt(0) + "5LUT_O5") == null) {
                 // LUT input siteWire has no net attached, nor does the LUT output sitewire: no need for site pin
                 return;
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -3240,12 +3240,6 @@ public class DesignTools {
             return false;
         }
 
-        Cell cell = net.getDesign().getCell(epi.getName());
-        if (cell != null) {
-            // Hier ports have no corresponding cell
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3206,4 +3206,46 @@ public static void createModuleInstsFromModuleImplsInsts(Design design, Collecti
...
+            return false;
+        }
+
+        EDIFPortInst epi = sourcePorts.get(0);
+        if (!epi.isTopLevelPort()) {
+            // Hier ports must be top level ports
+            return false;
+        }
+
+        Cell cell = net.getDesign().getCell(epi.getName());
```
eddieh-xlnx, line 3243: @clavin-xlnx Is it correct to assume that the corresponding `Cell` of a top-level `EDIFPortInst` would be the named the same?

clavin-xlnx: In general, that tends to be the behavior of designs generated by Vivado.  However, I don't think it is strictly enforced or required.

eddieh-xlnx: Would you recommend removing this check, or is there some other way of going from top-level `EDIFPortInst` to `Cell`? It is an overly-cautious check since you would expect `Net.getSource()` to not be `null` if there was a I/O `Cell`...

clavin-xlnx: I'm ok to remove it.  At this point in the code, `epi` is a top level port and there would be no way to get the cell from it as it references its parent instance, not a cell instance in the context of the cell.

clavin-xlnx (approved):

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3224,4 +3206,46 @@ public static void createModuleInstsFromModuleImplsInsts(Design design, Collecti
...
+        if (cell != null) {
+            // Hier ports have no corresponding cell
+            return false;
+        }
```
eddieh-xlnx, line 3247:
```suggestion
```

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: Done in de02d9e!

PR #599, opened 2023-01-10
Make loading gzipped EDIF decompress in memory by default, on to disk by option

clavin-xlnx:
There is a tradeoff when using gzipped EDIF files that, because RapidWright has a very fast parallel parser, decompressing onto disk first is much faster (~2x or more).  However, this temporarily uses ~18x more disk space in order to do so.  This change makes the default method to decompress (more slowly) in memory in order to prevent out of disk space errors.  If users would like the additional parsing speed with gzipped EDIF, they can set the parameter `RW_DECOMPRESS_GZIPPED_EDIF=1` (either via an Environment Variable or through the jvm command line `-Dname=value` parameter).

### DIFF
--- a/src/com/xilinx/rapidwright/util/Params.java
+++ b/src/com/xilinx/rapidwright/util/Params.java
@@ -1,0 +1,43 @@
+package com.xilinx.rapidwright.util;
+
+/**
+ * Aims to be a centralized helper class to manage global RapidWright settings.
+ */
+public class Params {
+
+    /**
+     * Flag to have RapidWright decompress gzipped EDIF files to disk prior to
+     * parsing. This is a tradeoff where pre-decompression improves runtime of the
+     * default method which is to decompress in memory. The disadvantage is that
+     * this will temporarily consume ~18x more disk space than original gzipped EDIF
+     * file, but loading of the EDIF file will be ~2x faster.
+     */
+    public static boolean RW_DECOMPRESS_GZIPPED_EDIF = isParamSet("RW_DECOMPRESS_GZIPPED_EDIF");
+    
+    /**
+     * Checks if the named RapidWright parameter is set
+     * 
+     * @param key Name of the global RapidWright parameter
+     * @return True if the parameter is set (as defined by {@link #isSet(String)}),
+     *         false otherwise
+     */
+    public static boolean isParamSet(String key) {
+        return isSet(System.getenv(key)) || isSet(System.getProperty(key));
+    }
+
+    /**
+     * Checks if a parameter is set by examining the provided value.
+     * 
+     * @param value An environment variable or JVM parameter value
+     * @return True if (1) value is not null, (2) is not an empty string, (3) is not
+     *         0 and (4) is not false (case-insensitive).
+     */
+    public static boolean isSet(String value) {
+        return !( value == null 
+               || value.length() == 0 
+               || value.equals("0") 
+               || value.toLowerCase().equals("false")
+               );         
+    }
+    
+}
--- a/src/com/xilinx/rapidwright/util/Params.java
+++ b/src/com/xilinx/rapidwright/util/Params.java
@@ -1,9 +1,34 @@
+/*
+ *
+ * Copyright (c) 2023, Advanced Micro Devices, Inc.
+ * All rights reserved.
+ *
+ * Author: Chris Lavin, AMD AECG Research Labs.
+ *
+ * This file is part of RapidWright.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
 package com.xilinx.rapidwright.util;
 
 /**
  * Aims to be a centralized helper class to manage global RapidWright settings.
  */
 public class Params {
+
+    public static String RW_DECOMPRESS_GZIPPED_EDIF_NAME = "RW_DECOMPRESS_GZIPPED_EDIF";
 
     /**
      * Flag to have RapidWright decompress gzipped EDIF files to disk prior to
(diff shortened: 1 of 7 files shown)


test/src/com/xilinx/rapidwright/edif/TestEDIFParser.java
```diff
@@ -123,8 +124,14 @@ public void loadEDIFSingleThreaded() throws IOException {
...
     }
 
-    @Test
-    public void testGZIPEDIFParsing(@TempDir Path tempDir) throws IOException {
+    @ParameterizedTest
+    @ValueSource(booleans = { false, true })
+    public void testGZIPEDIFParsing(boolean decompressToDisk, @TempDir Path tempDir)
+            throws IOException {
+        if (decompressToDisk) {
+            System.setProperty("RW_DECOMPRESS_GZIPPED_EDIF", "1");
```
clavin-xlnx, line 132: I'm torn about using a String to reference the parameter here (and below in the sister test).  Usually this is bad practice, but I think the name of the boolean parameter should probably be the same as the name of the interface parameter.  We could add a separate string parameter `String RW_DECOMPRESS_GZIPPED_EDIF_NAME` for example, and reference that, but it doesn't seem the best.  Another option might be to create a named Boolean class (String+boolean), but that seems like overkill. Ideally, we would never change the name of a paramter (just create a new one), but never say never, I guess.

src/com/xilinx/rapidwright/util/Params.java
```diff
@@ -0,0 +1,43 @@
...
+     * Flag to have RapidWright decompress gzipped EDIF files to disk prior to
+     * parsing. This is a tradeoff where pre-decompression improves runtime of the
+     * default method which is to decompress in memory. The disadvantage is that
+     * this will temporarily consume ~18x more disk space than original gzipped EDIF
+     * file, but loading of the EDIF file will be ~2x faster.
+     */
+    public static boolean RW_DECOMPRESS_GZIPPED_EDIF = isParamSet("RW_DECOMPRESS_GZIPPED_EDIF");
+    
+    /**
+     * Checks if the named RapidWright parameter is set
```
clavin-xlnx, line 18:
```suggestion
     * Checks if the named RapidWright parameter is set either via an environment variable or jvm parameter
```

eddieh-xlnx:
As I understand it, there are now two modes of operation:
1. (Original) Temporarily decompress the gzipped EDIF to disk, parse this uncompressed EDIF (where possible, in parallel), and then delete this temp file.
2. (This PR) Parse this gzipped EDIF by decompressing in memory. Where the parallel parser is triggered, each thread continues to `InputStream.skip()` to the (estimated) offset in the file to start parsing. Unlike a file stream which can be arbitrarily-seeked efficiently, for a gzipped stream, data must be decompressed and discarded until that offset in the file is reached which is the cause for this slowdown.
Is that right?

src/com/xilinx/rapidwright/util/function/InputStreamSupplier.java
```diff
@@ -49,7 +52,7 @@ static InputStreamSupplier fromPath(Path p) {
     public static InputStream getInputStream(Path fileName) {
         InputStream in = null;
         try {
-            if (fileName.toString().endsWith(".gz")) {
+            if (fileName.toString().endsWith(".gz") && Params.RW_DECOMPRESS_GZIPPED_EDIF) {
```
eddieh-xlnx, line 55: Should another condition for decompressing to disk be that we are invoking the parallel EDIF parser?

eddieh-xlnx, line 55: `InputStreamSupplier` seems to be a generic utility that could be used outside of the EDIF parser, so maybe `decompressToDisk` should be an method parameter (which would also address the comment too)

src/com/xilinx/rapidwright/util/Params.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+package com.xilinx.rapidwright.util;
+
+/**
+ * Aims to be a centralized helper class to manage global RapidWright settings.
+ */
+public class Params {
+
+    public static String RW_DECOMPRESS_GZIPPED_EDIF_NAME = "RW_DECOMPRESS_GZIPPED_EDIF";
```
eddieh-xlnx, line 31: Does a better name for this parameter exist? I fear that this name could (at a glance) be inferred to mean that RapidWright will refuse to open any gzipped EDIFs at all unless enabled. Ideas:
```
RW_DECOMPRESS_GZIPPED_EDIF_TO_DISK
RW_GZIPPED_EDIF_USE_DISK
RW_FAST_GZIP_EDIF_PARSING
```

src/com/xilinx/rapidwright/util/function/InputStreamSupplier.java
clavin-xlnx: Both serial and parallel parsers are faster if decompressed to disk.  I think I treat them both the same.

eddieh-xlnx: I did not know that. I retract!

clavin-xlnx: Updated the comment and added the new parameter.

src/com/xilinx/rapidwright/util/Params.java
clavin-xlnx: Refactored to `RW_DECOMPRESS_GZIPPED_EDIF_TO_DISK`

src/com/xilinx/rapidwright/util/function/InputStreamSupplier.java
```diff
@@ -34,25 +34,29 @@
...
+     *                         the InputStream.
+     * @return An InputStream of the file, or of a decompressed copy of a gzipped
+     *         file.
      */
-    public static InputStream getInputStream(Path fileName) {
+    public static InputStream getInputStream(Path fileName, boolean decompressToDisk) {
         InputStream in = null;
         try {
-            if (fileName.toString().endsWith(".gz") && Params.RW_DECOMPRESS_GZIPPED_EDIF) {
+            if (decompressToDisk) {
```
eddieh-xlnx, line 59: I think we should preserve the `fileName.toString().endsWith(".gz")` check in this method as that determines whether to return a `GzipInputStream`, especially since it's done below too. `decompressToDisk` would be a no-op for non-gzipped files.

eddieh-xlnx (approved):

src/com/xilinx/rapidwright/util/Params.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Aims to be a centralized helper class to manage global RapidWright settings.
+ */
+public class Params {
+
+    public static String RW_DECOMPRESS_GZIPPED_EDIF_TO_DISK_NAME = "RW_DECOMPRESS_GZIPPED_EDIF_TO_DISK";
+
+    /**
+     * Flag to have RapidWright decompress gzipped EDIF files to disk prior to
+     * parsing. This is a tradeoff where pre-decompression improves runtime of the
```
eddieh-xlnx, line 35:
```suggestion
     * parsing. This is a tradeoff where pre-decompression improves runtime over the
```

src/com/xilinx/rapidwright/util/function/InputStreamSupplier.java
clavin-xlnx: Added the check back, thanks.

eddieh-xlnx (approved):

PR #604, opened 2023-01-17
Classes to support BEL attributes

clavin-xlnx:
Relevant to the discussion in #598.


src/com/xilinx/rapidwright/design/AttrSite.java
```diff
@@ -0,0 +1,99 @@
...
+
+import com.xilinx.rapidwright.device.BEL;
+import com.xilinx.rapidwright.device.Site;
+import com.xilinx.rapidwright.device.SiteTypeEnum;
+
+/**
+ * Represents a site and its configurable attributes.
+ * 
+ */
+public class AttrSite {
```
eddieh-xlnx, line 36: What's the reason for this over `SiteAttr` which seems to be more consistent with other classes?

clavin-xlnx: Its not a great name, but I think `SiteAttr` is misleading as there are not real attributes being applied to the site, it is just collecting configurations that belong to it. What do you think of `ConfiguredSite`?

eddieh-xlnx: I like `ConfiguredSite` more than `AttrSite`. My suggestion would be `SiteConfig`, but if you feel more strongly about `ConfiguredSite` then let's go with that.

src/com/xilinx/rapidwright/design/SiteConfig.java
```diff
@@ -42,11 +42,11 @@ public class AttrSite {
     private Map<BEL, Map<String, BELAttr>> belAttrs;
 
 
-    public static AttrSite getAttrSite(Site site, SiteTypeEnum type) {
-        return new AttrSite(site, type);
+    public static SiteConfig getAttrSite(Site site, SiteTypeEnum type) {
```
eddieh-xlnx, line 45:
```suggestion
    public static SiteConfig createSiteConfig(Site site, SiteTypeEnum type) {
```

`getSiteConfig` implies get-if-exists behaviour, whereas it looks like this is all about creating new instances?

eddieh-xlnx (approved):

src/com/xilinx/rapidwright/design/SiteConfig.java
clavin-xlnx: Thanks for catching this, yes create is more appropriate.

Issue #603, opened 2023-01-17
Names of hierarchical Cells and Nets in 2022.2 DCPs parsed incorrectly


jakobwenzel:
RapidWright seems to parse [this Vivado 2022.2 DCP](https://github.com/Xilinx/RapidWright/files/10436275/design.zip) incorrectly. The names of cells and nets located at deeper levels of hierarchy are incorrect. For cells, corruption seems to start at the second level, while for nets it seems to start at the third.

When writing the design back to disk, RapidWright produces a corrupted DCP. Vivado can no longer read placement/routing of affected cells/nets.

The DCP contains these cells (corrupted names in bold):

Correct Name        |  Name in RapidWright
-|-
r_reg | r_reg
inst/r_reg |**r_reginst**
inst/inst/r_reg | **r_reginstinst**

And these nonstatic nets (corrupted names in bold):
Correct Name | Name in RapidWright
-|-
clk | clk
i | i
inst/inst/o | **instinst/o**
inst/r | inst/r
r | r

<details><summary>Regenerating the DCP (collapsed)</summary>
<p>

Save this as `design.v`:
```verilog
module testing #(parameter DEPTH=3) (input wire i, output wire o, input wire clk);
reg r;
always @(posedge clk) r <= i;
generate
	if (DEPTH<=1) begin
		assign o=r;
	end else begin
		testing #(DEPTH-1) inst(r, o, clk);
	end
endgenerate
endmodule
```
Run these commands in Vivado:
```tcl
read_verilog design.v
synth_design -mode out_of_context -top testing -part XC7A200T-SBG484-1 -flatten_hierarchy none
place_design
route_design
write_checkpoint -force design.dcp
write_edif -force design.edf
```

This minimal example uses `-flatten_hierarchy none` to preserve hierarchy. On larger designs, I've seen this issue occur without the flag.

RapidWright code:
```java
        Design design = Design.readCheckpoint(args[0], args[1]);
        design.writeCheckpoint("saveAgain.dcp");

        System.out.println("cells");
        for (Cell cell : design.getCells()) {
            System.out.println(cell.getName());
        }

        System.out.println("\nnets");
        for (Net net : design.getNets()) {
            if (!net.isStaticNet() && !net.getName().equals(Net.USED_NET)) {
                System.out.println(net.getName());
            }
        }
```

</p>
</details>

jakobwenzel:
Thanks for the quick fix! I've tried it with some bigger designs and it works perfectly.

[...]

PR #605, opened 2023-01-17
Test for fix of issue #603

clavin-xlnx:


test/src/com/xilinx/rapidwright/design/TestDCPLoad.java
```diff
@@ -125,6 +127,13 @@ public void testDCPFromVivado2022_1(String dcp) {
     @ParameterizedTest
     @ValueSource(strings = {"picoblaze_2022.2.dcp"})
     public void testDCPFromVivado2022_2(String dcp) {
-        RapidWrightDCP.loadDCP(dcp);
+        Design design = RapidWrightDCP.loadDCP(dcp);
+        for (Cell c : design.getCells()) {
+            if (c.getName().contains("processor")) {
```
eddieh-xlnx, line 132: Are you restricting it to `processor` for runtime reasons, or is there a reason this check would fail for non-`processor`-containing paths?

clavin-xlnx: This is an issue with hierarchy.  There are some `Cell`s that should be excluded at the top level (`PAD` or port cells) that don't have a hierarchical equivalent. All cells with hierarchy contain `processor` so this was an easy way to satisfy the scenario.

eddieh-xlnx (approved):

eddieh-xlnx (approved):

PR #610, opened 2023-02-18
DesignTools.getPortInstsFromSitePinInst() to cope with LUT RTs

eddieh-xlnx:
Currently, if a LUT routethru services more than one cell (e.g. LUT output fans to two FFs) then only one `EDIFHierPortInst` will be returned.

Fix this here.

### DIFF
--- a/test/src/com/xilinx/rapidwright/design/TestDesignTools.java
+++ b/test/src/com/xilinx/rapidwright/design/TestDesignTools.java
@@ -23,6 +23,23 @@
 
 package com.xilinx.rapidwright.design;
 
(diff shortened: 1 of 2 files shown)


test/src/com/xilinx/rapidwright/design/TestDesignTools.java
```diff
@@ -34,6 +34,7 @@
 import java.util.Map.Entry;
 import java.util.Set;
 
+import com.xilinx.rapidwright.edif.EDIFPortInst;
```
eddieh-xlnx, line 37: Reorder

clavin-xlnx (approved):

PR #616, opened 2023-03-02
Handle EDIF export bussed port name collision with identically named bit ports

clavin-xlnx:
This fixes an issue when RapidWright writes out an EDIF netlist that contains a cell with two ports, one bussed and the other a single bit that have the same root name, for example:
```
my_port
my_port[1:0]
```
This is a legal set of ports, but when the EDIF is exported from RapidWright, the EDIF rename collides on port instance references.  This PR resolves the issue by explicitly using the `busName` of the `EDIFPort` for any bussed ports which contains a trailing `[` character that then gets translated to an `_` when it is legalized for EDIF.  This differentiates it from its single bit counterpart.

### DIFF
--- a/test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
+++ b/test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
@@ -298,9 +298,11 @@ class TestEDIFNetlist {
         EDIFCellInst ff = top.createChildCellInst("ff", Design.getPrimitivesLibrary().getCell("FDRE"));
         origNetlist.getHDIPrimitivesLibrary().addCell(ff.getCellType());
 
+        String portName = "unfortunate_name";
+
         // Create two ports, one single-bit and another bussed with the same root name
-        EDIFPort port0 = top.createPort("unfortunate_name", EDIFDirection.INOUT, 1);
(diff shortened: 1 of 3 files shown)


test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
```diff
@@ -287,4 +288,33 @@ public void testMigrateCellsAndSubCells(boolean uniquify) {
...
+        EDIFNet net1 = top.createNet("net1");
+        net1.createPortInst(port1, 1);
+        net1.createPortInst("R", ff);
+
+        Path tempFile = path.resolve("test.edf");
+        origNetlist.exportEDIF(tempFile);
+
+        EDIFNetlist testNetlist = EDIFTools.readEdifFile(tempFile);
+        EDIFNetlistComparator comparer = new EDIFNetlistComparator();
+        Assertions.assertEquals(0, comparer.compareNetlists(origNetlist, testNetlist));
```
eddieh-xlnx, line 318: This relies on the `EDIFNetlistComparator` functioning perfectly now and into the future.

Since you've already read `testNetlist` from disk, is it possible to also do a quick check to see if both ports continue to exist, and whether they are attached to the right net?

clavin-xlnx: Added an explicit check.

eddieh-xlnx (approved):

PR #617, opened 2023-03-03
Add option to return the number of diffs as exit code in EDIFNetlistComparator

clavin-xlnx:


src/com/xilinx/rapidwright/edif/compare/EDIFNetlistComparator.java
```diff
@@ -408,5 +421,9 @@ public static void main(String[] args) {
             comparator.printDiffReport(System.out);
         }
         t.stop().printSummary();
+        
+        if (setExitCodeAsDiffCount) {
+            System.exit(diffs);
```
eddieh-xlnx, line 426: https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html states that the exit code is [0,255]. Perhaps we want to saturate at 255?
Alternatively, exiting with `0` for success and `1` in all other cases should be sufficient.

clavin-xlnx: I will change it to `0` for no diffs and `1` for 1 or more diffs.  This is common behavior for `diff`.

eddieh-xlnx (approved):

PR #618, opened 2023-03-06
[RWRoute] Static net router + RouteNode enhancements

eddieh-xlnx:
Static net router to now reuse existing used resources when performing partial routing.

Refactor `RouteNode` functionality into `RouteNodeInfo`, fixing coordinate and wirelength computations as well as improving clarity and robustness.

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
+++ b/src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
@@ -33,7 +33,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Queue;
 import java.util.Set;
-import java.util.function.Predicate;
+import java.util.function.Function;
 
 import com.xilinx.rapidwright.design.Design;
 import com.xilinx.rapidwright.design.Net;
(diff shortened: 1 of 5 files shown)


src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -279,15 +279,21 @@ private static ClockRegion findCentroid(Net clk, Device device) {
         return device.getClockRegion(center.y, center.x);
     }
 
+    public enum NodeStatus {
```
clavin-xlnx, line 282: I know its a hassle, but consider possibly moving this to its own file.  Nested classes/enums create some untidiness in compilation.

src/com/xilinx/rapidwright/rwroute/RouteNodeInfo.java
```diff
@@ -0,0 +1,180 @@
...
+        if (endTile == null) {
+            endTile = node.getTile();
+        }
+
+        RouteNodeType type = getType(node, endTile, pinfeedIntoLaguna);
+        short endTileXCoordinate = getEndTileXCoordinate(node, type, (short) endTile.getTileXCoordinate());
+        short endTileYCoordinate = (short) endTile.getTileYCoordinate();
+        short length = getLength(baseTile, type, endTileXCoordinate, endTileYCoordinate);
+
+        return new RouteNodeInfo(type, endTileXCoordinate, endTileYCoordinate, length);
```
clavin-xlnx, line 83: This seems like it may get called frequently and creating a 'throw-away' object each time `get()` is called may impact performance.  Is there a way we can have a thread-safe static variable that doesn't require `new` on each call?

eddieh-xlnx: The only thread-safe variable is a [`ThreadLocal`](https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html) which I am under the impression requires a hash lookup -- do you know of any others?

I'm aware that there are a lot of create-objects-and-throw-it-away (e.g. `Node.getDownhillNodes()`) operations so I was hoping that the JVM is efficient with this pattern or that it can optimize the construction some way.

Otherwise, in the current implementation of RWRoute, routing node creation is actually single-threaded currently, so writing into a static variable won't be a problem until we need to make it thread safe.

clavin-xlnx: Ok, let's leave it for now.  Has performance changed at all based on this PR?

clavin-xlnx (approved):

PR #619, opened 2023-03-09
[EDIF] Fix for skipping past EOF, improve tests

eddieh-xlnx:
When parsing a gzipped EDIF, we can only estimate its uncompressed size in order to decide where each worker thread should start parsing. The possibility exists that estimate is so incorrect, and when there are enough workers, that one worker is asked to start at an offset that is past the end of the uncompressed file.

Handle this situation, and improve the tests to catch this.

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFTokenizer.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFTokenizer.java
@@ -443,9 +443,18 @@ public class EDIFTokenizer implements AutoCloseable {
         }
         long actual = 0;
         while (actual < i) {
-            actual += in.skip(i-actual);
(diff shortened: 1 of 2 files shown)


clavin-xlnx (approved):
Thanks!  Note that the test will hang when the fix is not present (not saying we need to create an elaborate test to force a fail, though).

eddieh-xlnx:
> Note that the test will hang when the fix is not present (not saying we need to create an elaborate test to force a fail, though).

Correct, it will just hang. There may be a way to set a per-test timeout.
Also a little frustrating is that reverting the fix does not cause any failures on GitHub Actions. It only fails locally, so we'll just live with that for now.

PR #624, opened 2023-03-10
Fixes RelocationTools to correspond to new Cell.equals() behavior

clavin-xlnx:
Now that routethru cells are distinguished by their routethru flag, the `RelocationTools` check was no longer handling the routethrus correctly.


eddieh-xlnx (approved):
Passes all tests!

PR #633, opened 2023-03-27
Fix PBlockGenerator Parser

clavin-xlnx:
In more recent utilization reports, some lines appear more than once and overwrite the original parsed values.  This is a quick fix to avoid this issue.


eddieh-xlnx (approved):
Looks reasonable and low impact, but haven't tested it myself.

PR #643, opened 2023-03-30
More compact data file, all family coverage with comprehensive test

clavin-xlnx:
In a new release, all family unisim placement data will be aliased by architecture upon load of the `unisim_data.dat` file.  This will minimize the needed size of the file and provide broader coverage automatically.

### DIFF
--- a/test/src/com/xilinx/rapidwright/device/TestUnisimPlacements.java
+++ b/test/src/com/xilinx/rapidwright/device/TestUnisimPlacements.java
@@ -35,6 +35,10 @@ import com.xilinx.rapidwright.edif.EDIFCell;
 
 public class TestUnisimPlacements {
 
+    /**
+     * Set of unsupported unisim families (for now) in 2022.2.3. Hopefully will add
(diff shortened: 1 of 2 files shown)


test/src/com/xilinx/rapidwright/device/TestUnisimPlacements.java
```diff
@@ -32,9 +35,24 @@
 
 public class TestUnisimPlacements {
 
+    private static Set<FamilyType> unsupportedTypes = null;
+
+    static {
+        unsupportedTypes = new HashSet<>();
```
eddieh-xlnx, line 41: What's the significance of these unsupported families? Are they not currently supported by RapidWright at all? Or do they just cause this particular test to fail?

clavin-xlnx: Just added a comment to clarify.  Adding support is a bit more involved, will hopefully add in 2023.1.0.

eddieh-xlnx (approved):

PR #644, opened 2023-04-03
Use ProcessBuilder.inheritIO() instead of StreamGobbler

eddieh-xlnx:
Doing so *seems* to eliminate some spurious failures where the `java` process exits with code 1 with no apparent output.


clavin-xlnx (approved):
Cool, is this something we can use to replace `StreamGobbler` entirely?  Should we deprecate the class in favor of this technique?

eddieh-xlnx:
> Cool, is this something we can use to replace `StreamGobbler` entirely? Should we deprecate the class in favor of this technique?

I think that's a good idea for a later PR. Currently, the only other use of this in this repo appears to be `FileTools.runCommand()` which appears to only be called by `com.xilinx.rapidwright.debug.createILADesign()` to call Vivado.

Issue #410, opened 2022-05-07
" Too many words to allocate: 333129474" beyond the limit of capnp?


the-centry:
==============================================================================
==                Device Resources Dump: xcvu7p-flvb2104-1-e                ==
==============================================================================
Loading device from file xcvu7p-flvb2104-1-e
             Load Device:     3.590s    440.991MBs
           populateEnums:     0.458s     52.941MBs
               SiteTypes:     0.082s      4.029MBs
               TileTypes:     0.200s      4.374MBs
                   Tiles:     0.494s     44.722MBs
INFO: Building uncommon Wire->Node cache...
      This might take a few seconds for large devices on the first call.  
      It is generally triggered when getting the Node from an uncommon Wire object.  
      To avoid printing this message, set Device.QUIET_MESSAGE=true or set the ENVIRONMENT variable RW_QUIET_MESSAGE=1.
INFO: Finished building uncommon Wire->Node cache
Exception in thread "main" java.lang.RuntimeException: Too many words to allocate: 333129474
	at org.capnproto.BuilderArena.allocate(BuilderArena.java:130)
	at org.capnproto.WireHelpers.allocate(WireHelpers.java:80)
	at org.capnproto.WireHelpers.initStructListPointer(WireHelpers.java:509)
	at org.capnproto.StructList$Factory.initFromPointerBuilder(StructList.java:81)
	at org.capnproto.StructList$Factory.initFromPointerBuilder(StructList.java:30)
	at org.capnproto.StructBuilder._initPointerField(StructBuilder.java:194)
	at com.xilinx.rapidwright.interchange.DeviceResources$Device$Builder.initWires(DeviceResources.java:906)
	at com.xilinx.rapidwright.interchange.DeviceResourcesWriter.writeAllWiresAndNodesToBuilder(DeviceResourcesWriter.java:807)
	at com.xilinx.rapidwright.interchange.DeviceResourcesWriter.writeDeviceResourcesFile(DeviceResourcesWriter.java:280)
	at com.xilinx.rapidwright.interchange.DeviceResourcesExample.main(DeviceResourcesExample.java:29)

**Second**
If the source code of  [rapidwright_jars.zip](https://github.com/Xilinx/RapidWright/releases/download/v2022.1.0-beta/rapidwright_jars.zip) is opened for users ?

clavin-xlnx:
[...]

Issue #176, opened 2021-05-06
IOException thrown from org.capnproto.Serialize.read()


mkurc-ant:
This is a followup (presumably) issue of https://github.com/Xilinx/RapidWright/issues/171.

With #171 fixed when trying to generate device resources database for some of the largest FPGAs the following exception is now thrown:
```
Exception in thread "main" java.io.IOException: too many segments
	at org.capnproto.Serialize.read(Serialize.java:68)
	at com.xilinx.rapidwright.interchange.Interchange.readInterchangeFile(Interchange.java:76)
	at com.xilinx.rapidwright.interchange.DeviceResourcesVerifier.verifyDeviceResources(DeviceResourcesVerifier.java:124)
	at com.xilinx.rapidwright.interchange.DeviceResourcesExample.main(DeviceResourcesExample.java:35)
```

I've confirmed the issue for the following devices:
 - `xczu11eg-ffvc1156-1-e`
 - `xczu15eg-ffvc900-1-e`
 - `xczu17eg-ffvb1517-1-e`
 - `xczu19eg-ffvb1517-1-e`
 - `xc7vx550tffg1927-1`
 - `xc7vx690tffg1927-1`
 - `xc7vx980tffg1930-1`

clavin-xlnx:
The bug was in capnproto-java and the author kindly fixed the issue and created an updated release which can be downloaded here:
```
wget https://repo1.maven.org/maven2/org/capnproto/runtime/0.1.6/runtime-0.1.6.jar -O $(RAPIDWRIGHT_PATH)/jars/runtime-0.1.6.jar
```
You'll have to update the `CLASSPATH` environment variable to point to the new 0,1.6 jar instead of 0.1.4.  Re-running with the update appears to solve the `too many segments` issue.

clavin-xlnx:
Correction, the issue is solved for at least `xczu11eg-ffvc1156-1-e`, but I get another error with `xczu19eg-ffvb1517-1-e`:
```
Exception in thread "main" java.lang.IllegalArgumentException: capacity < 0: (-1862524376 < 0)
	at java.base/java.nio.Buffer.createCapacityException(Buffer.java:256)
	at java.base/java.nio.ByteBuffer.allocate(ByteBuffer.java:347)
	at org.capnproto.Serialize.makeByteBuffer(Serialize.java:34)
	at org.capnproto.Serialize.read(Serialize.java:89)
[...]

PR #658, opened 2023-04-26
Enables large FPGA Interchange devices for placement

clavin-xlnx:
This adds an option (`--skip_route_resources`) to `DeviceResourcesExample` that will skip adding wires and nodes to the device model FPGA Interchange file.  Currently there are a couple of issues (#410, also see https://github.com/Xilinx/RapidWright/issues/176#issuecomment-1030410003) that would prevent large designs from being written out with routing.  

This also fixes a performance bug that caused the `Cell <-> BEL pin map:` phase to take almost 20 minutes, with this change it now takes about 2 minutes.

The output from a run with the following parameters:

`java -Xmx32000m com.xilinx.rapidwright.interchange.DeviceResourcesExample xcvu19p --skip_route_resources`

```
            Prims&Macros:     0.812s     65.209MBs          |   1652.207MBs (peak)
    Cell <-> BEL pin map:   140.047s      0.565MBs          |   5249.344MBs (peak)
                Packages:     0.032s      0.179MBs          |   5253.449MBs (peak)
               Constants:    73.127s   2530.729MBs          |   8781.207MBs (peak)
              Wire Types:     0.011s     -0.373MBs          |   8781.207MBs (peak)
                 Strings:     0.699s     94.670MBs          |   8781.207MBs (peak)
              Write File:     5.147s      0.010MBs          |   8781.207MBs (peak)
             Verify file:   779.180s   3381.796MBs          |  24665.059MBs (peak)
                 *Total*:  1004.774s   6878.560MBs          |  24665.059MBs (peak)
```

### DIFF
--- a/src/com/xilinx/rapidwright/interchange/EnumerateCellBelMapping.java
+++ b/src/com/xilinx/rapidwright/interchange/EnumerateCellBelMapping.java
@@ -741,34 +741,42 @@ public class EnumerateCellBelMapping {
             }
 
             for (List<String> parameters : getParametersFor(series, cell.getName())) {
+                HashSet<Map.Entry<String, String>> pinMapping = null;
+
                 for (Site site : siteMap.get(siteType)) {
-                    SiteInst siteInst = design.createSiteInst("test_site", siteType, site);
+                    if (pinMapping == null) {
+                        SiteInst siteInst = design.createSiteInst("test_site", siteType, site);
 
-                    String[] parameterArray = parameters.toArray(new String[parameters.size()]);
-                    physCell = design.createAndPlaceCell("test", Unisim.valueOf(cell.getName()), site.getName() + "/" + bel, parameterArray);
+                        String[] parameterArray = parameters.toArray(new String[parameters.size()]);
+                        physCell = design.createAndPlaceCell("test", Unisim.valueOf(cell.getName()),
+                                site.getName() + "/" + bel, parameterArray);
 
-                    // Build complete P2L map.
-                    Map<String, String> pinMap = new HashMap<String, String>();
+                        // Build complete P2L map.
+                        Map<String, String> pinMap = new HashMap<String, String>();
 
-                    for (Map.Entry<String, Set<String>> pinPair : physCell.getPinMappingsL2P().entrySet()) {
-                        for (String physPin : pinPair.getValue()) {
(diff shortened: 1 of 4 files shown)


src/com/xilinx/rapidwright/interchange/EnumerateCellBelMapping.java
```diff
@@ -741,34 +741,42 @@ public static void populateCellBelPin(Enumerator<String> allStrings, Map<SiteTyp
...
-                    pinMap.putAll(physCell.getPinMappingsP2L());
+                        pinMap.putAll(physCell.getPinMappingsP2L());
 
-                    HashSet<Map.Entry<String, String>> pinMapping = new HashSet<Map.Entry<String, String>>();
-                    for (Map.Entry<String, String> pinPair : pinMap.entrySet()) {
-                        pinMapping.add(new AbstractMap.SimpleEntry<String, String>(pinPair.getValue(), pinPair.getKey()));
-                    }
+                        pinMapping = new HashSet<Map.Entry<String, String>>();
+                        for (Map.Entry<String, String> pinPair : pinMap.entrySet()) {
+                            pinMapping.add(
```
eddieh-xlnx, line 767: I'm not sure I understand what's going on here. `pinMapping` appears a `Set` of a pair of Strings -- is there need for it to be a set at all?

`pinMap` seems to be (unique) physical -> logical mapping created from logical -> physical mapping. Then `pinMapping` seems to reverse that again?

clavin-xlnx: The previous author of this code probably saw `Map.Entry<>` as a Java proxy for the Python tuple construct and used it judiciously.  This entire section of code should be re-written, but I believe that is beyond the scope of this PR.  The only change here that is relevant is that we are avoiding the re-computation of the map for every site when the map will be identical for all sites of the same type.

eddieh-xlnx: Okay, let's make a note that this may be sped up further.

eddieh-xlnx (approved):

src/com/xilinx/rapidwright/interchange/EnumerateCellBelMapping.java
clavin-xlnx: See #659

Issue #632, opened 2023-03-22
Detect unsupported architectures in RWRoute


clavin-xlnx:
After loading Series 7 devices, RWRoute proceeds without checking if the architecture is supported.  Since RWRoute does not support Series 7 devices, it should probably produce an error message indicating that the architecture is unsupported upfront.

eddieh-xlnx:
Fixed in https://github.com/Xilinx/RapidWright/pull/663

PR #663, opened 2023-05-02
Fix for #632 - Detect unsupported RWRoute devices

clavin-xlnx:
Address issue in #632 by detecting upfront if a targeted device is supported by RWRoute.

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/RWRoute.java
+++ b/src/com/xilinx/rapidwright/rwroute/RWRoute.java
@@ -27,6 +27,7 @@ package com.xilinx.rapidwright.rwroute;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -155,12 +157,31 @@
     /** A map storing routes from CLK_OUT to different INT tiles that connect to sink pins of a global clock net */
     protected Map<String, List<String>> routesToSinkINTTiles;
 
+    public static final Set<Series> SUPPORTED_SERIES;
+
+    static {
+        SUPPORTED_SERIES = new HashSet<>();
```
eddieh-xlnx, line 163: Consider using `EnumSet` here.

eddieh-xlnx (approved):

src/com/xilinx/rapidwright/rwroute/RWRoute.java
clavin-xlnx: Thanks, updated.

PR #667, opened 2023-05-05
Add EDIFHierNet.getLeafHierPortInsts() overload

eddieh-xlnx:


src/com/xilinx/rapidwright/edif/EDIFHierNet.java
```diff
@@ -201,19 +201,29 @@ public List<EDIFHierPortInst> getLeafHierPortInsts() {
...
+     * @param includeSourcePins A flag to include source pins in the result.
+     * @param includeSinkPins A flag to include sink pins in the result.
      * @param visited An initial set of EDIFHierNet-s that have already been visited and will not
      * be visited again. Pre-populating this set can be useful for blocking traversal.
      * @return The list of all leaf cell port instances connected to this hierarchical net and its
      * aliases.
      */
-    public List<EDIFHierPortInst> getLeafHierPortInsts(boolean includeSourcePins, Set<EDIFHierNet> visited) {
+    public List<EDIFHierPortInst> getLeafHierPortInsts(boolean includeSourcePins, boolean includeSinkPins, Set<EDIFHierNet> visited) {
+        assert(includeSourcePins || includeSinkPins);
```
clavin-xlnx, line 225: I'm not sure I agree that both flags being `false` indicates an error state.  If I was looking for just `INOUT` ports, I could see that being useful.  Perhaps we could clarify in the Javadoc that when both flags are false, it will return any `INOUT` ports?

eddieh-xlnx: Hmm. According to my testing, `EDIFPort{Inst,HierInst}.isInput()` and `EDIFPort{Inst,HierInst}.isOutput()` will return false for `INOUT`-s. This looks like the only way to tell an inout. Does this match your understanding?

For this method, it looks like inout-s (on blackbox cells, as I don't think it makes sense for a FPGA primitive to have a inout that isn't connected to a top-level port) and thus are never picked up. This should probably change.

clavin-xlnx: Ok, let's change the comment to declare that this method will never return `INOUT` ports/portInsts and that setting both flags to `false` will always return an empty list.

clavin-xlnx (approved):

PR #668, opened 2023-05-05
Test for setting/adding duplicate Net source pins

eddieh-xlnx:
Test that:
* `Net.addPin()` does not a source pin that is already present
* `Net.setSource()` fires an assertion if to be same as the current alternate source
* `Net.setAlternateSource()` fires an assertion if to be same as current source


test/src/com/xilinx/rapidwright/design/TestNet.java
```diff
@@ -187,4 +187,66 @@ public void testCreatePinInvalid() {
...
+        String pinName = "H_O";
+        boolean samePin = altPinName.equals(pinName);
+        SitePinInst spi1 = new SitePinInst(pinName, si);
+        Assertions.assertTrue(net.addPin(spi1));
+        SitePinInst spi2 = new SitePinInst(altPinName, si);
+        Assertions.assertEquals(samePin, spi1.equals(spi2));
+        Assertions.assertEquals(!samePin, net.addPin(spi2));
+
+        Assertions.assertSame(spi1, net.getSource());
+        if (altPinName.equals(pinName)) {
```
clavin-xlnx, line 206:
```suggestion
        if (samePin) {
```

clavin-xlnx (approved):

PR #671, opened 2023-05-05
EDIFNetlist.cellInstIOStandardFallback to collect set of IOSTANDARDs

eddieh-xlnx:
Instead of throwing a `RuntimeError` when there is a conflict.

This can happen in non-uniquified netlists, and may not necessarily be a problem if such IOSTANDARDs are all compatible -- i.e. they do not produce a conflicting answer as to whether a macro should be expanded.
e.g. `LVDS` and `LVDS_25` would both cause `OBUFDS` to not be expanded, and so the fact there are two IOSTANDARDs that map back to the `EDIFCellInst` is not a problem.


src/com/xilinx/rapidwright/edif/EDIFNetlist.java
```diff
@@ -1610,9 +1612,19 @@ public void expandMacroUnisims(Series series) {
...
-                                if (exception.getSecond().contains(ioStandard)) {
+                                Boolean expand = null;
+                                for (EDIFPropertyValue value : getIOStandards(inst)) {
+                                    IOStandard ioStandard = IOStandard.valueOf(value.getValue());
+                                    boolean contained = exception.getSecond().contains(ioStandard);
+                                    if (expand == null) {
+                                        expand = contained;
+                                    } else {
+                                        if (expand != contained) {
+                                            throw new RuntimeException();
```
clavin-xlnx, line 1623: Perhaps we could add a message to this exception?

clavin-xlnx (approved):

src/com/xilinx/rapidwright/edif/EDIFNetlist.java
eddieh-xlnx: Ooh yes, forgot about that one!

PR #674, opened 2023-05-17
Ignore TestCheckOpenFilesInstalled.test if outside of gradle

zakn-amd:
PR #397 modified the gradle environment such that every testcase is automatically extended with methods to ensure the test closes all of the files it opens. However when run outside of gradle (e.g. in an IDE) the testcase that checks to see if this extension is running obviously fails. This commit adds a conditional to skip the test if run outside of gradle.

### DIFF
--- a/test/shared/com/xilinx/rapidwright/support/CheckOpenFilesExtension.java
+++ b/test/shared/com/xilinx/rapidwright/support/CheckOpenFilesExtension.java
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2021-2022, Xilinx, Inc.
- * Copyright (c) 2022, Advanced Micro Devices, Inc.
+ * Copyright (c) 2022-2023, Advanced Micro Devices, Inc.
  * All rights reserved.
  *
  * Author: Eddie Hung, Xilinx Research Labs.
@@ -32,13 +32,17 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.extension.AfterTestExecutionCallback;
 import org.junit.jupiter.api.extension.BeforeTestExecutionCallback;
+import org.junit.jupiter.api.extension.ConditionEvaluationResult;
+import org.junit.jupiter.api.extension.ExecutionCondition;
 import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.engine.config.JupiterConfiguration;
 
 /**
  * Check that a testcase does leak open files
@@ -137,10 +141,18 @@ public class CheckOpenFilesExtension implements BeforeTestExecutionCallback, Aft
         Assertions.assertNotNull(getBeforeList(extensionContext), "Open Files Checker Extension does not seem to be running");
     }
 
-    public static class CheckOpenFilesWorkingExtension implements AfterTestExecutionCallback {
+    public static class CheckOpenFilesWorkingExtension implements AfterTestExecutionCallback, ExecutionCondition {
         @Override
         public void afterTestExecution(ExtensionContext context) {
             assertExtensionInstalled(context);
+        }
+
+        @Override
+        public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+            Optional<String> extensions = context.getConfigurationParameter(JupiterConfiguration.EXTENSIONS_AUTODETECTION_ENABLED_PROPERTY_NAME);
+            if ("true".equals(extensions.orElse("false")))
+                return ConditionEvaluationResult.enabled(JupiterConfiguration.EXTENSIONS_AUTODETECTION_ENABLED_PROPERTY_NAME + " == true");
+            return ConditionEvaluationResult.disabled(JupiterConfiguration.EXTENSIONS_AUTODETECTION_ENABLED_PROPERTY_NAME + " != true");
         }
     }
 }
--- a/test/shared/com/xilinx/rapidwright/support/CheckOpenFilesExtension.java
+++ b/test/shared/com/xilinx/rapidwright/support/CheckOpenFilesExtension.java
@@ -36,6 +36,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.extension.AfterTestExecutionCallback;
 import org.junit.jupiter.api.extension.BeforeTestExecutionCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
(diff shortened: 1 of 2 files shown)


test/shared/com/xilinx/rapidwright/support/CheckOpenFilesExtension.java
```diff
@@ -140,6 +141,9 @@ public static void assertExtensionInstalled(ExtensionContext extensionContext) {
     public static class CheckOpenFilesWorkingExtension implements AfterTestExecutionCallback {
         @Override
         public void afterTestExecution(ExtensionContext context) {
+            // if "gradle" not in command line ignore this test
+            boolean gradleStart = ManagementFactory.getRuntimeMXBean().getInputArguments().stream().anyMatch(n -> n.contains("gradle"));
+            Assumptions.assumeTrue(gradleStart);
```
eddieh-xlnx, line 146: Two comments:
1. Would it be better to stick this assumption into the actual `TestCheckOpenFilesInstalled` invocation, as opposed to the callback which all (Gradle-executed) tests will call?
2. I wonder if there is a better way to check for "am I executed from Gradle"? (Let me do some research here too)

eddieh-xlnx: Re: 2, I'm seeing three System properties that mention `gradle` and only exist when run from Gradle:
```
org.gradle.test.worker
org.gradle.native
org.gradle.internal.worker.tmpdir
```

zakn-amd: re 1: adding the assumption to TestCheckOpenFilesInstalled.test() doesn't work due to how Junit handles before/after anotations (https://github.com/junit-team/junit5/issues/2919#issuecomment-1132651012). This assumption was added to the afterTestExecution() method in the nested class CheckOpenFilesWorkingExtension. AFAIK Only TestCheckOpenFilesInstalled.test() is extended with CheckOpenFilesExtension.CheckOpenFilesWorkingExtension.class so only this test will run the gradle check; all other tests are auto-extended with CheckOpenFilesExtension.class which doesn't contain the overriding method with the assumption.

re 2: this is pretty janky, but I couldn't think of anything better (and simple). I think it might be possible to use some OS commands to get the parent process's PID and then convert that to a command line and check to see if the parent process is gradle but this is extremely not portable.

eddieh-xlnx: Consider something like this?
```
public class TestCheckOpenFilesInstalled {
    @Test
    @EnabledIf("runningInGradle")
    @ExtendWith(CheckOpenFilesExtension.CheckOpenFilesWorkingExtension.class)
    public void test() throws IOException {
        //Actual test is in extension
    }

    private boolean runningInGradle() {
        return System.getProperty("org.gradle.test.worker") != null;
    }
}
```

zakn-amd: `@EnabledIf` certainly seems like a better option. Picking one system property is maybe a little less general than reading the whole command line but presumably there's no way a gradle test can be run without defining a worker thread.

eddieh-xlnx: A more robust option would be to have our Gradle invocation define a custom property, and then detect that.

zakn-amd:
changed this to use @EnableIf. To check if we're running in gradle I set it to search through all system properties to check for any mention of "gradle". I think this is better than setting up gradle to define a custom property, since I believe this extension is enabled with a gradle property. If the extension is not working because custom gradle properties are not set correctly then this test might just be skipped instead of failing.

eddieh-xlnx:
> I think this is better than setting up gradle to define a custom property, since I believe this extension is enabled with a gradle property.

There is a small danger here is that we if anyone anywhere in the future sets a property containing the substring `gradle` -- such as `some.future.application.definitely_not_gradle` then this will trigger.

> If the extension is not working because custom gradle properties are not set correctly then this test might just be skipped instead of failing.

I would say that's a more tolerable risk. If Gradle fails to set custom properties, a lot more (e.g. LSF) should be failing that just skipping this test.

test/src/com/xilinx/rapidwright/TestCheckOpenFilesInstalled.java
```diff
@@ -27,12 +27,18 @@
...
 public class TestCheckOpenFilesInstalled {
     @Test
+    @EnabledIf("runningInGradle")
     @ExtendWith(CheckOpenFilesExtension.CheckOpenFilesWorkingExtension.class)
     public void test() throws IOException {
         //Actual test is in extension
     }
+
+    private boolean runningInGradle() {
+        return System.getProperty("junit.jupiter.extensions.autodetection.enabled") != null;
```
eddieh-xlnx, line 42: Pedantic:
```suggestion
        return System.getProperty("junit.jupiter.extensions.autodetection.enabled").equals("true");
```

eddieh-xlnx (changes requested):

test/src/com/xilinx/rapidwright/TestCheckOpenFilesInstalled.java
zakn-amd: I don't think this works when running outside of gradle; since `null.equals("true")` gives the error: "Cannot invoke equals(String) on the primitive type null"

eddieh-xlnx: Ooh, yes! Try the other way around:
```
return "true".equals(System.getProperty("junit.jupiter.extensions.autodetection.enabled"));
```

eddieh-xlnx:
Looks like this doesn't work when the test is invoked through the `testLsf` path. Can you cherry-pick c6b4ed1d4114cb1a07d8d1b123a65408934f31b6 which tries yet another way to resolve this please?

test/shared/com/xilinx/rapidwright/support/CheckOpenFilesExtension.java
```diff
@@ -137,10 +141,18 @@ public static void assertExtensionInstalled(ExtensionContext extensionContext) {
...
-    public static class CheckOpenFilesWorkingExtension implements AfterTestExecutionCallback {
+    public static class CheckOpenFilesWorkingExtension implements AfterTestExecutionCallback, ExecutionCondition {
         @Override
         public void afterTestExecution(ExtensionContext context) {
             assertExtensionInstalled(context);
         }
+
+        @Override
+        public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+            Optional<String> extensions = context.getConfigurationParameter(JupiterConfiguration.EXTENSIONS_AUTODETECTION_ENABLED_PROPERTY_NAME);
```
eddieh-xlnx, line 152:
```suggestion
            // Only run this extension (used by TestCheckOpenFilesInstalled) when extensions
            // are enabled which should be the case when invoked by through Gradle, but not
            // necessarily through an IDE
            Optional<String> extensions = context.getConfigurationParameter(JupiterConfiguration.EXTENSIONS_AUTODETECTION_ENABLED_PROPERTY_NAME);
```

PR #646, opened 2023-04-06
EDIFNetlist.getIOStandard() to inherit IOStandard from EDIFNet

eddieh-xlnx:
Anecdotally, it's been observed that Vivado propagates the `IOStandard` property from the `EDIFNet` feeding a top-level output port to its driving `EDIFCellInst` I/O buffer. Do so here too, with `EDIFNetlist.getIOStandard()` which backs onto the `EDIFTools.getIOStandard()` static helper.

This `IOStandard` turns out to be important for determining whether certain macros should be expanded (see https://github.com/Xilinx/RapidWright/pull/505).

PR #675, opened 2023-05-19
[EDIF] EDIFNetlist.collapseMacroUnisims() to not clobber cell

eddieh-xlnx:
.. plus optimize `expandMacroUnisims()`.

Currently, the following scenario is not considered:
* Design has `OBUFDS` cell instantiations
* Only some of those `OBUFDS` cells are expanded into `OBUFDS_DUAL_BUF`
* Collapsing causes the `OBUFDS_DUAL_BUF` cell to be renamed to `OBUFDS`, without awareness that `OBUFDS` already exists.
* Now there are two `OBUFDS` cells being instantiated in the netlist, but only one in the library.

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFNetlist.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFNetlist.java
@@ -259,7 +259,7 @@ public class EDIFNetlist extends EDIFName {
         if (cell == null) {
             cell = Design.getUnisimCell(unisim);
         }
-        return lib.addCell(cell);
+        return new EDIFCell(lib, cell, unisim.name());
     }
 
     public EDIFLibrary getWorkLibrary() {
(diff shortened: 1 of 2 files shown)


eddieh-xlnx:
> I recall you found an issue regarding the attribute being on a port/net would have to be artificially migrated to the instance for the exception to be triggered. Will that case need consideration here?

I think you mean https://github.com/Xilinx/RapidWright/pull/646, along with its derivative PRs? If so, I think this PR is orthogonal: #646 and the like deal with `IOStandard` properties on the `EDIFCellInst` (or top-level `EDIFNet` it is connected to). This PR deals with the `EDIFCell` definition.

clavin-xlnx:
> I think you mean #646, along with its derivative PRs? If so, I think this PR is orthogonal: #646 and the like deal with `IOStandard` properties on the `EDIFCellInst` (or top-level `EDIFNet` it is connected to). This PR deals with the `EDIFCell` definition.

Good, thanks for clarifying.

PR #676, opened 2023-05-19
[Interchange] Split Enumerator into {Identity,String} variants

eddieh-xlnx:
`StringEnumerator` is the same as the old `Enumerator<String>`.

`IdentityEnumerator` uses an `IdentityHashMap` (i.e. Java reference "pointer") to uniquely enumerate all objects.

As a result, can get rid of `EDIFEnumerable` which requires expensive string concatenation and hashing.


src/com/xilinx/rapidwright/interchange/IdentityEnumerator.java
```diff
@@ -78,7 +65,7 @@ public void ensureSize(int size) {
 
     @Override
     public T get(int index) {
-        if (size() -1 < index) return null;
+        // if (size() -1 < index) return null;
```
clavin-xlnx, line 68: Ok to remove?

clavin-xlnx (approved):
Thanks for making this change.  Did this lead to any measurable speedup?

eddieh-xlnx:
> Thanks for making this change. Did this lead to any measurable speedup?

Yup: it brought runtime down of WriteLogNetlist from 43s down to 35s on *a really big design*.  There are savings for the "Populate Enums" phase as well as the actually lookups that follow. There are likely memory savings too, since the string keys need not be kept in memory either.

PR #680, opened 2023-05-22
Switch all data files to Zstandard compression

clavin-xlnx:


src/com/xilinx/rapidwright/util/FileTools.java
```diff
@@ -1672,6 +1698,23 @@ public static boolean isDataBinary(InputStream is) {
...
+        
+    public static boolean isFileGzipped(Path path) {
+        try (InputStream in = Files.newInputStream(path)) {
+            boolean isGzipped = false;
+            try {
+                isGzipped = in.read() == GZIP_START_BYTES[0] 
+                         && in.read() == GZIP_START_BYTES[1];
+            } catch (IOException e) {
+                throw new UncheckedIOException(e);
+            }
```
eddieh-xlnx, line 1711: Do you need this inner `try/catch`?

clavin-xlnx: Thanks, I've simplified it.

eddieh-xlnx (approved):

clavin-xlnx:
To summarize the impact of this change for the data files, device model file sizes were reduced by a geomean of 32% with a geomean reduction in load time by 11%.  For the device node cache files, the savings were even more pronounced, 52% smaller file sizes with a 108% reduction in load times.

image

image

image

image

mithro:
QQ - Did you do any test comparing SSD verse spinning rust? I guess it would be even more pronounced / better on slower media.

clavin-xlnx:
> QQ - Did you do any test comparing SSD verse spinning rust? I guess it would be even more pronounced / better on slower media.

We didn't try testing specific storage technologies.  These tests were all on network attached storage (not sure of the physical hardware) which is fairly robust when not under significant load.  We did some similar tests with a RAMDisk and found the performance mostly the same.  But yes, presumably the improvement should be more significant on slower storage.

PR #682, opened 2023-05-23
Update unisim test for 2023.1.0

clavin-xlnx:


eddieh-xlnx (approved):
Is this test expected to passing because it's pulling in the new 2023.1-staging device files already?

clavin-xlnx:
> Is this test expected to passing because it's pulling in the new 2023.1-staging device files already?

Yes, it passes now with the latest release candidate jar for 2023.1.

PR #684, opened 2023-05-25
Add rapidwright.util.VivadoTools

zakn-amd:
Small utility class that contains a method to spawn a vivado process that consumes a tcl script, and a method to parse vivado's stdout. Also includes a basic testcase to exercise these methods.

Similar to com.xilinx.rapidwright.ipi.BlockUpdater.runVivadoTasks, but more generic and could probably be used to replace this (and possibly others)

### DIFF
--- a/src/com/xilinx/rapidwright/util/VivadoTools.java
+++ b/src/com/xilinx/rapidwright/util/VivadoTools.java
@@ -86,12 +86,7 @@ public class VivadoTools {
             fullyRoutedNets = parseLog(log, "# of fully routed nets");
             netsWithRoutingErrors = parseLog(log, "# of nets with routing errors");
 
-            String[] toDel = workdirHandle.list();
-            for (String s : toDel) {
-                File f = new File(workdirHandle.getPath(), s);
-                f.delete();
-            }
-            workdirHandle.delete();
+            FileTools.deleteFolder(workdirHandle.toString());
         }
     }
 


src/com/xilinx/rapidwright/util/RunVivado.java
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (c) 2021-2022, Xilinx, Inc.
```
eddieh-xlnx, line 2: Don't think this needs Xilinx copyright anymore. For new files, there's a `make ensure_headers` target that should generate the copyright header for you. Can you try that please?

src/com/xilinx/rapidwright/util/RunVivado.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Utility methods to run a tcl script through vivado and parse the stdout
+ * @author zakn
+ *
+ */
+public class RunVivado {
+    public static List<String> searchVivadoLog(List<String> log, String key) {
+        List<String> results = new ArrayList<>();
+        for(String l : log) {
+            if(l.contains(key)) {
```
eddieh-xlnx, line 38: Looks like you're missing a space between `for` or `if` and the open bracket; see https://github.com/Xilinx/RapidWright/blob/master/CODING_STYLE.md

It could be your Eclipse settings doing this too.

src/com/xilinx/rapidwright/util/RunVivado.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Utility methods to run a tcl script through vivado and parse the stdout
+ * @author zakn
+ *
+ */
+public class RunVivado {
```
eddieh-xlnx, line 34: Maybe call this class `Vivado` to reflect that it (will) contain all sorts of Vivado utilities.

src/com/xilinx/rapidwright/util/RunVivado.java
```diff
@@ -0,0 +1,77 @@
...
+        while (!j.isFinished()) {
+            if(verbose) {
+                System.out.println("Vivado running");
+            }
+            try {
+                Thread.sleep(5000);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
+        }
```
eddieh-xlnx, line 65: I'm not sure a `Job` here is the best tool here, since it does not output interactively to the screen -- I would prefer to see the progress on the screen, rather than wait until it terminates and then the entire output to appear. Consider using:
https://github.com/Xilinx/RapidWright/blob/45adb4f852b944dd3aaff14803edf8be8567cecd/src/com/xilinx/rapidwright/util/FileTools.java#L1553-L1560

FYI: There is a `JobQueue` class that takes care of this wait-until-done for you. Example:
https://github.com/Xilinx/RapidWright/blob/45adb4f852b944dd3aaff14803edf8be8567cecd/src/com/xilinx/rapidwright/interchange/GenerateInterchangeDevices.java#L45-L53
Typically, it's used for when there is more than one job being run in parallel.

src/com/xilinx/rapidwright/util/RunVivado.java
```diff
@@ -0,0 +1,77 @@
...
+        List<String> results = new ArrayList<>();
+        for(String l : log) {
+            if(l.contains(key)) {
+                results.add(l);
+            }
+        }
+        return results;
+    }
+
+    public static List<String> runVivadoTask(String runDir, String tclScript, boolean verbose) {
```
eddieh-xlnx, line 45: Please add Javadoc for this method. I would also suggest naming it the more specific `runTcl{File,Command}`.

It is ambiguous without looking at the method body whether the `tclScript` parameter is a Tcl script file, or a Tcl command. Ideally, we would have both, with the latter building on the former by writing the command to a temp file and sourcing that.

test/src/com/xilinx/rapidwright/util/TestRunVivado.java
```diff
@@ -0,0 +1,88 @@
...
+    @Test
+    public void testOpenDcpReadLog(@TempDir Path tempDir) throws IOException {
+        Assumptions.assumeTrue(FileTools.isVivadoOnPath());
+        // XXX
+        // grabbing a handle to one of the example dcps requires to some
+        // unfortunate path hacking to work in and out of gradle. inside gradle
+        // we get a relative path, and we must convert to an absolute path
+        // since vivado is run in the tempDir. outside of gradle we get an
+        // absolute path and no conversion is needed.
+        final Path dcpPath = Path.of(RapidWrightDCP.getString("picoblaze_partial.dcp"));
```
eddieh-xlnx, line 48: `RapidWrightDCP.getPath()` exists.

test/src/com/xilinx/rapidwright/util/TestRunVivado.java
```diff
@@ -0,0 +1,88 @@
...
+        int parsed = Integer.parseInt(results.get(0).replaceAll("[^\\d]", ""));
+        Assertions.assertEquals(12144, parsed);
+
+        // check to see that a non existent key phrase doesn't break things
+        results = RunVivado.searchVivadoLog(log, "This key shouldn't be in the log!");
+        Assertions.assertEquals(0, results.size());
+
+        // check to see everything works if we get multiple matches
+        results = RunVivado.searchVivadoLog(log, "# of");
+        Assertions.assertEquals(9, results.size());
```
eddieh-xlnx, line 84: It'd be great to see this be embedded this parse `report_route_status` capability (be the first to be) embedded inside the Vivado class so that anyone who wants to use can do so with just one call.

I'm imagining something like:
```
Design design = ...;
ReportRouteStatusResult result = Vivado.reportRouteStatus(design); // Note that this takes a `Design` object, and will automatically write it out as a DCP that can be read by Vivado.
if (result.numUnroutedNets == 0 && result.numNetsWithRoutingErrors == 0) {
  // Good
} else {
  // Bad
}
// ReportRouteStatusResult would also contain all the other entries too.
```
Perhaps that if condition above could even be a `isFullyRouted()` method?

eddieh-xlnx (commented):
Thanks for building this! This is a great start, but I think we can take this PR even further to be useful right out of the box.

src/com/xilinx/rapidwright/util/RunVivado.java
zakn-amd: Calling this class just `Vivado` seems wrong since it is an interface/collection of utilities to access vivado. Maybe `VivadoUtils`?

eddieh-xlnx: Maybe `VivadoTools` then?

src/com/xilinx/rapidwright/util/RunVivado.java
zakn-amd: `FileTools.runCommand` intercepts std.out and std.err in order to print the output, and as a result I don't think there's an easy way to post process it (e.g. to parse the output of `report_route_status`). Perhaps I missed something obvious, but that's why I went with a `job`. I agree that printing the output live would be better though. I'll see if I can make that happen.

eddieh-xlnx: Consider that Vivado has `-log` argument that allows you to override the output log file name (which gets written on top of printing to stdout). Even more specific is that `report_route_status` has a `-file` option which writes its output to a file (but looks like it doesn't print to screen?)

zakn-amd: Reading the vivado logfile could work. Right now I'm also using the `setRunDir()` method of `LocalJob` to ensure that output logs can be put in a `tempDir`. I'm not sure if there's a similar way to do this with `runCommand`, it seems like we'd just pollute the root directory with vivado logs otherwise. Although perhaps this isn't that big of an issue.

src/com/xilinx/rapidwright/util/RunVivado.java
zakn-amd: added Javadoc and overloaded this method to support single tcl commands or a script: https://github.com/Xilinx/RapidWright/pull/684/commits/23c2bd5ebab8579a71e576d8f85be5897fa2f463

test/src/com/xilinx/rapidwright/util/TestRunVivado.java
zakn-amd: https://github.com/Xilinx/RapidWright/pull/684/commits/df344218fd3dca6a21fb9a06785b7892bac75954

test/src/com/xilinx/rapidwright/util/TestRunVivado.java
zakn-amd: https://github.com/Xilinx/RapidWright/pull/684/commits/c5e778de407ec9062ed95d57a0eb1ae1145fade2 here's a first pass at implementing this functionality

src/com/xilinx/rapidwright/util/VivadoTools.java
```diff
@@ -0,0 +1,179 @@
...
+     */
+    public static List<String> runTcl(Path outputLog, Path tclScript, boolean verbose) {
+        final String vivadoCmd = "vivado -log " + outputLog.toString() + " -mode batch -source "
+                + tclScript.toString();
+        FileTools.runCommand(vivadoCmd, verbose);
+        List<String> log = new ArrayList<>();
+        log = FileTools.getLinesFromTextFile(outputLog.toString());
+        return log;
+    }
+}
```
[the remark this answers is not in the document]
zakn-amd: https://github.com/Xilinx/RapidWright/pull/684/commits/baca3e7d9c1bef9026a21aa5ec5e3e01d65bf671

src/com/xilinx/rapidwright/util/VivadoTools.java
```diff
@@ -0,0 +1,179 @@
...
+        for (String l : log) {
+            if (l.contains(key)) {
+                results.add(l);
+            }
+        }
+        return results;
+    }
+
+    /**
+     * method to run a single tcl command in vivado
```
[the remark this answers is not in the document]
zakn-amd: https://github.com/Xilinx/RapidWright/pull/684/commits/baca3e7d9c1bef9026a21aa5ec5e3e01d65bf671

src/com/xilinx/rapidwright/util/VivadoTools.java
```diff
@@ -0,0 +1,179 @@
...
+     * builds a tcl script to open the Design d in vivado and run the command
+     * `report_route_status`
+     * 
+     * @param d       the design to be analyzed
+     * @param workdir a temporary directory to run vivado in (will contain logs,
+     *                dcp, etc)
+     * @return ReportRouteStatus object with fields initialized with the output of
+     *         `report_route_status`
+     */
+    public static ReportRouteStatusResult reportRouteStatus(Design d, Path workdir) {
```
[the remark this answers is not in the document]
zakn-amd: https://github.com/Xilinx/RapidWright/pull/684/commits/baca3e7d9c1bef9026a21aa5ec5e3e01d65bf671 refactored to make/delete a work directory inplace

eddieh-xlnx (approved):

src/com/xilinx/rapidwright/util/VivadoTools.java
```diff
@@ -0,0 +1,151 @@
...
+import java.io.File;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+
+import com.xilinx.rapidwright.design.Design;
+
+/**
+ * Utility methods to run access vivado and parse logs
```
clavin-xlnx, line 34:
```suggestion
 * Utility methods to provide access to vivado and parse logs
```

src/com/xilinx/rapidwright/util/VivadoTools.java
```diff
@@ -0,0 +1,151 @@
...
+ * 
+ * @author zakn
```
clavin-xlnx, line 36:
```suggestion
```
Author name is in the top comment.

src/com/xilinx/rapidwright/util/VivadoTools.java
```diff
@@ -0,0 +1,151 @@
...
+ *
+ */
+public class VivadoTools {
```
clavin-xlnx, line 39: Is the idea to have multiple classes going forward dedicated to each task?  If so, I would suggest that we create a dedicated package for Vivado instead of putting them all into the same class.  We have tried to avoid inner classes because they tend to complicate compilation and a few other areas.

Alternatively, the class could be converted into a static method and just return a `Map<String,Integer>` that contains the results.  This would be simpler and would probably be the preferred method in my opinion.

test/src/com/xilinx/rapidwright/util/TestVivadoTools.java
```diff
@@ -0,0 +1,62 @@
...
+    @Test
+    public void testReportRouteStatus() throws IOException {
+        Assumptions.assumeTrue(FileTools.isVivadoOnPath());
+        String dcp = RapidWrightDCP.getPath("picoblaze_partial.dcp").toString();
+        Design d = Design.readCheckpoint(dcp);
+        VivadoTools.ReportRouteStatusResult r = new VivadoTools.ReportRouteStatusResult(d);
+
+        Assertions.assertEquals(12144, r.unroutedNets);
+    }
+
```
clavin-xlnx, line 60:
```suggestion
```

clavin-xlnx (commented):
Thanks for developing this!

src/com/xilinx/rapidwright/util/VivadoTools.java
zakn-amd: @eddieh-xlnx has been directing some of the architecture decisions here, but I think the idea is to add classes for future vivado tasks, so maybe a separate package is warranted.

Returning a `Map<String,Integer>` is perhaps little more awkward, since the user then needs to know exactly what vivado output log key phrase they want (i.e. `"# of unrouted nets"`). I was mostly following the proposed system from here: https://github.com/Xilinx/RapidWright/pull/684#discussion_r1205979575

clavin-xlnx: > @eddieh-xlnx has been directing some of the architecture decisions here, but I think the idea is to add classes for future vivado tasks, so maybe a separate package is warranted.

Ok, sure.  Let's keep it the way it is for now and we can revisit it in the future if the file seems becomes unwieldly.  

> Returning a `Map<String,Integer>` is perhaps little more awkward, since the user then needs to know exactly what vivado output log key phrase they want (i.e. `"# of unrouted nets"`). I was mostly following the proposed system from here: [#684 (comment)](https://github.com/Xilinx/RapidWright/pull/684#discussion_r1205979575)

Sure, I can see how that is a bit cumbersome and error prone.  For simplicity, let's keep it as is.

src/com/xilinx/rapidwright/util/VivadoTools.java
```diff
@@ -0,0 +1,149 @@
...
+            unroutedNets = parseLog(log, "# of unrouted nets");
+            fullyRoutedNets = parseLog(log, "# of fully routed nets");
+            netsWithRoutingErrors = parseLog(log, "# of nets with routing errors");
+
+            String[] toDel = workdirHandle.list();
+            for (String s : toDel) {
+                File f = new File(workdirHandle.getPath(), s);
+                f.delete();
+            }
+            workdirHandle.delete();
```
clavin-xlnx, line 94: Would [`FileTools.deleteFolder()`](https://github.com/Xilinx/RapidWright/blob/6b107b77bd85d8edd72da70ecbefb962222ab026/src/com/xilinx/rapidwright/util/FileTools.java#L846) work here?

clavin-xlnx (approved):

src/com/xilinx/rapidwright/util/VivadoTools.java
zakn-amd: yes it would: https://github.com/Xilinx/RapidWright/pull/684/commits/de5a93a7f84c8597b203adc68b01f0a7dd3d6855

PR #687, opened 2023-05-30
Adds GitHub Action Smoke Test for RapidWright Wrapper

clavin-xlnx:


eddieh-xlnx:
Can you try with the "#!/bin/bash` shebang instead?

clavin-xlnx:
> Can you try with the `#!/bin/bash` shebang for `rapidwright` instead?

Should we try to stick to the wider compatible `/bin/sh`?  Gradle seems to handle this task in `sh` with:

https://github.com/Xilinx/RapidWright/blob/38c4b68008f3ae4cfec4e1363a4e1635688aca4a/gradlew#L27-L40

eddieh-xlnx:
> > Can you try with the `#!/bin/bash` shebang for `rapidwright` instead?
> 
> Should we try to stick to the wider compatible `/bin/sh`? `gradlew` seems to handle this task in `sh` with:

Yes, that's a good idea. I learned there's a material difference between single bracket and double bracket today!

As for the latest failure, looks like we need to use `$0` instead of `${BASH_SOURCE}` too
https://stackoverflow.com/questions/29832037/how-to-get-script-directory-in-posix-sh

clavin-xlnx:
> As for the latest failure, looks like we need to use `$0` instead of `${BASH_SOURCE}` too: https://stackoverflow.com/questions/29832037/how-to-get-script-directory-in-posix-sh

I reused the logic from `gradlew`, not sure if that is something we want to separate out.  I also had to restructure the script so that the `--list-apps` option to return 0 for GitHub Actions not to fail.

eddieh-xlnx (approved):

PR #690, opened 2023-06-02
Updates rapidwright.bat script to work with symlink

clavin-xlnx:


clavin-xlnx:
Couldn't find a more elegant way to check if we are running from the symlink... (`%~dp0` gives the directory of the symlink, not the actual script as  ) so open to suggestions.  The equivalent functionality we re-used from `gradlew` is: 

```
# Attempt to set APP_HOME
# Resolve links: $0 may be a link
PRG="$0"
# Need this for relative symlinks.
while [ -h "$PRG" ] ; do
    ls=`ls -ld "$PRG"`
    link=`expr "$ls" : '.*-> \(.*\)$'`
    if expr "$link" : '/.*' > /dev/null; then
        PRG="$link"
    else
        PRG=`dirname "$PRG"`"/$link"
    fi
done
SAVED="`pwd`"
cd "`dirname \"$PRG\"`/" >/dev/null
APP_HOME="`pwd -P`"
cd "$SAVED" >/dev/null
```

The `gradlew.bat` doesn't appear to even try.

eddieh-xlnx (approved):
I don't have RapidWright set up on Windows so I'll take your word that this works! Perhaps we should re-introduce Windows just for a smoke test?

The other option that exists is we just duplicate the batch file and put it into `bin/`.

PR #692, opened 2023-06-02
Refactor wrapper scripts

clavin-xlnx:


build.gradle
```diff
@@ -49,16 +50,33 @@ sourceSets {
...
     from configurations.api
     into 'jars'
     //testFixturesImplementation contains the output jar, as well as an implicit dependency on the task.
     //Filter to remove the jar itself, getFiles() to remove the implicit dependency
     from configurations.testFixturesImplementation.filter { !tasks.getByName("jar").outputs.files.contains(it) } .getFiles()
     into 'jars'
+    dependsOn preserveScripts
+
 }
 compileJava.dependsOn(copyJars)
```
eddieh-xlnx, line 68:
```suggestion
}
compileJava.dependsOn(copyJars)
compileJava.dependsOn(preserveScripts)
```
for consistency/clarity.

build.gradle
```diff
@@ -49,16 +50,33 @@ sourceSets {
   }
 }
 
+task preserveScripts(type: Copy) {
```
eddieh-xlnx, line 53: Would `*Bin` be a better moniker than `*Scripts`? It does sit in the `bin` directory after all.

eddieh-xlnx (approved):

build.gradle
clavin-xlnx: I thought this wouldn't work because I thought the deletion was being triggered during the `copyJars` task, but I just tried it again and it appears to work.

build.gradle
clavin-xlnx: Sure, I've updated the names accordingly.

PR #695, opened 2023-06-05
Relevant tests to check against Vivado's report_route_status

eddieh-xlnx:
CC: @zakn-amd

### DIFF
--- a/src/com/xilinx/rapidwright/util/ReportRouteStatusResult.java
+++ b/src/com/xilinx/rapidwright/util/ReportRouteStatusResult.java
@@ -1,0 +1,54 @@
+package com.xilinx.rapidwright.util;
+
+import com.xilinx.rapidwright.design.Design;
+
+import java.io.File;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ReportRouteStatusResult {
+
+    public final int logicalNets;
+    public final int netsNotNeedingRouting;
+    public final int internallyRoutedNets;
+    public final int netsWithNoLoads;
+    public final int implicitlyRoutedPorts;
+    public final int routableNets;
+    public final int unroutedNets;
+    public final int fullyRoutedNets;
+    public final int netsWithRoutingErrors;
+
+    private static int parseLog(List<String> log, String key) {
+        List<String> matchingLines = VivadoTools.searchVivadoLog(log, key);
+        if (matchingLines.isEmpty()) {
+            return 0;
+        }
+        // Consider first match only
+        return Integer.parseInt(matchingLines.get(0).replaceAll("[^\\d]", ""));
+    }
+
+    /**
+     * Analyze a Tcl script to open the Design d in vivado and run the command
+     * `report_route_status`
+     *
+     * @param log the design to be analyzed
+     */
+    ReportRouteStatusResult(List<String> log) {
+        logicalNets = parseLog(log, "# of logical nets");
+        netsNotNeedingRouting = parseLog(log, "# of nets not needing routing");
+        internallyRoutedNets = parseLog(log, "# of internally routed nets");
+        netsWithNoLoads = parseLog(log, "# of nets with no loads");
+        implicitlyRoutedPorts = parseLog(log, "# of implicitly routed ports");
+        routableNets = parseLog(log, "# of routable nets");
+        unroutedNets = parseLog(log, "# of unrouted nets");
+        fullyRoutedNets = parseLog(log, "# of fully routed nets");
+        netsWithRoutingErrors = parseLog(log, "# of nets with routing errors");
+    }
+
+    public boolean isFullyRouted() {
+        return unroutedNets == 0 && netsWithRoutingErrors == 0;
+    }
+
+}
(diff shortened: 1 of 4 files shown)


src/com/xilinx/rapidwright/util/ReportRouteStatusResult.java
```diff
@@ -0,0 +1,70 @@
...
+        // Consider first match only
+        return Integer.parseInt(matchingLines.get(0).replaceAll("[^\\d]", ""));
+    }
+
+    /**
+     * Analyze a Tcl script to open the Design d in vivado and run the command
+     * `report_route_status`
+     *
+     * @param log the design to be analyzed
+     */
```
eddieh-xlnx, line 53:
```suggestion
    /**
     * Analyze a log file produced by Vivado's `report_route_status`
     * command.
     *
     * @param log the List<String> of lines to be analyzed
     */
```

zakn-amd:
seems like https://github.com/Xilinx/RapidWright/blob/aeba8f683e13de452cb1f6db47ba9a4097a8e914/test/src/com/xilinx/rapidwright/util/TestVivadoTools.java#L25 and https://github.com/Xilinx/RapidWright/blob/aeba8f683e13de452cb1f6db47ba9a4097a8e914/test/src/com/xilinx/rapidwright/util/TestVivadoTools.java#L27 are never used and can be removed.

src/com/xilinx/rapidwright/util/VivadoTools.java
```diff
@@ -137,8 +85,67 @@ public static List<String> runTcl(Path outputLog, Path tclScript, boolean verbos
...
+        final Path outputLog = workdir.resolve("outputLog.log");
+
+        StringBuilder sb = new StringBuilder();
+        sb.append("open_checkpoint " + dcp + "; ");
+        sb.append(REPORT_ROUTE_STATUS);
+
+        List<String> log = VivadoTools.runTcl(outputLog, sb.toString(), true);
+        return new ReportRouteStatusResult(log);
     }
 }
```
zakn-amd, line 151:
```suggestion
    /**
     * Run Vivado's `report_route_status` command on the provided Design object
     * and return its result as a ReportRouteStatusResult object.
     *
     * @param design Design object to report on.
     * @return ReportRouteStatusResult object.
     */
    public static ReportRouteStatusResult reportRouteStatus(Design design) {
        return reportRouteStatus(design, null, null);
    }

    /**
     * Run Vivado's `report_route_status` command on the provided DCP path
     * and return its result as a ReportRouteStatusResult object.
     *
     * @param dcp Path to DCP to report on.
     * @return ReportRouteStatusResult object.
     */
    public static ReportRouteStatusResult reportRouteStatus(Path dcp) {
        return reportRouteStatus(null, dcp, null);
    }

    /**
     * Run Vivado's `report_route_status` command on the provided DCP path
     * and return its result as a ReportRouteStatusResult object.
     *
     * @param dcp Path to DCP to report on.
     * @param workdir Directory to work within.
     * @return ReportRouteStatusResult object.
     */
    public static ReportRouteStatusResult reportRouteStatus(Design design, Path dcp, Path workdir) {
        boolean clean = false;
        if (workdir == null) {
            workdir = FileSystems.getDefault()
                    .getPath("vivadoToolsWorkdir" + FileTools.getUniqueProcessAndHostID());
            File workdirHandle = new File(workdir.toString());
            workdirHandle.mkdirs();
            clean = true;
        }
        if (dcp == null) {
            dcp = workdir.resolve("checkpoint.dcp");
            design.writeCheckpoint(dcp);
        }

        final Path outputLog = workdir.resolve("outputLog.log");

        StringBuilder sb = new StringBuilder();
        sb.append("open_checkpoint " + dcp + "; ");
        sb.append(REPORT_ROUTE_STATUS);

        List<String> log = VivadoTools.runTcl(outputLog, sb.toString(), true);

        if (clean) {
            FileTools.deleteFolder(workdir.toString());
        }
        return new ReportRouteStatusResult(log);
    }
}
```
Could do something like this to try and remove the duplication of the working directory creation/deletion in the `reportRouteStatus(*)` methods

eddieh-xlnx: Good attempt! Feels a little imprecise to have a method that takes both a `Path` and a `Design` though, for the sake of reusing the make-temp-directory code though.

zakn-amd: yeah this is a little rough. I guess there are basically 4 combinations to care about: 
1. design + workdir 
2. design + no workdir (in your commit)
3. dcp path + workdir (in your commit)
4. dcp path + no workdir (in your commit)
we could have all of these be wrappers for a private function with an imprecise signature, but then that one function would be filled with some confusing branching to make everything work. Its just a shame to have two methods that only differ by two lines.

eddieh-xlnx: I would say that 2 and and 4 are the main ones to care about for a push-button approach. 3 came about as a byproduct of the common code in 2 and 4. There's also the cognitive load of understanding what a method does -- duplicating a few lines is worthwhile if it means the common code is truly common.

zakn-amd (approved):

PR #696, opened 2023-06-06
[RWRoute] Fix routeStaticNets() to not use other nets' nodes

eddieh-xlnx:
Discovered by https://github.com/Xilinx/RapidWright/pull/691 which changed `RWRoute.staticNetsAndRoutingTargets` to be an `IdentityHashMap` which has a non-deterministic iteration order.

This exposed a bug whereby the static net router would -- when routing static nets in a particular order -- allow one static net to overuse a node that was actually the pin-node for the other static net.

Fix this oversight, and make static net routing deterministic.

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/RWRoute.java
+++ b/src/com/xilinx/rapidwright/rwroute/RWRoute.java
@@ -437,7 +437,7 @@ public class RWRoute{
             }
         }
 
-        // Annotate all static pins with the net they're associated with to ensure that one
+        // Annotate all static pin nodes with the net they're associated with to ensure that one
         // net cannot unknowingly use a node needed by the other net
         Map<Node,Net> preservedStaticNodes = new HashMap<>();
         for (Map.Entry<Net,List<SitePinInst>> e : staticNetAndRoutingTargets.entrySet()) {
@@ -449,7 +449,7 @@ public class RWRoute{
             }
         }
 
-        // Iterate through both static nets in a stable order (not guaranteed by IdentityHashMap())
+        // Iterate through both static nets in a stable order (not guaranteed by IdentityHashMap)
         for (Net staticNet : Arrays.asList(design.getGndNet(), design.getVccNet())) {
             List<SitePinInst> pins = staticNetAndRoutingTargets.get(staticNet);
             if (pins == null) {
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -443,29 +437,51 @@ protected void routeStaticNets() {
...
-            GlobalSignalRouting.routeStaticNet(net,
+            Net staticNet = e.getKey();
+            for (SitePinInst sink : e.getValue()) {
+                Node node = sink.getConnectedNode();
+                preservedStaticNodes.put(node, staticNet);
+                assert(!routingGraph.isPreserved(node));
+            }
+        }
+
+        // Iterate through both static nets in a stable order (not guaranteed by IdentityHashMap())
```
eddieh-xlnx, line 452:
```suggestion
        // Iterate through both static nets in a stable order (not guaranteed by IdentityHashMap)
```

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -443,29 +437,51 @@ protected void routeStaticNets() {
             }
         }
 
+        // Annotate all static pins with the net they're associated with to ensure that one
```
eddieh-xlnx, line 440:
```suggestion
        // Annotate all static pin nodes with the net they're associated with to ensure that one
```

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -443,29 +437,51 @@ protected void routeStaticNets() {
...
+            }
+        }
+
+        // Iterate through both static nets in a stable order (not guaranteed by IdentityHashMap)
+        for (Net staticNet : Arrays.asList(design.getGndNet(), design.getVccNet())) {
+            List<SitePinInst> pins = staticNetAndRoutingTargets.get(staticNet);
+            if (pins == null) {
+                continue;
+            }
+            System.out.println("INFO: Route " + pins.size() + " pins of " + staticNet);
```
clavin-xlnx, line 458: Should this be something like "INFO: Routing " or "INFO: To route "?

clavin-xlnx (approved):

src/com/xilinx/rapidwright/rwroute/RWRoute.java
eddieh-xlnx, line 458:
```suggestion
            System.out.println("INFO: Routing " + pins.size() + " pins of " + staticNet);
```

PR #699, opened 2023-06-07
Linux RapidWright wrapper to not change working directory

eddieh-xlnx:
Found by @zakn-amd.

It appears that the new `rapidwright` wrapper on Linux changes the current working directory (to the root of the git repository) , thus when an application writes to the current directory it appears there as opposed to where it was invoked.

### DIFF
--- a/.github/workflows/test-rapidwright-wrapper.yml
+++ b/.github/workflows/test-rapidwright-wrapper.yml
@@ -35,7 +35,10 @@ jobs:
         run: bin/rapidwright com.xilinx.rapidwright.rwroute.RWRoute
 
       - name: Check Jython Command Execution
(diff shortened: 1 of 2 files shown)


eddieh-xlnx:
Looks like my use of `test` is failing on Windows. My first approach was to do it with Jython and use the exit code, however, due to https://github.com/Xilinx/RapidWright/pull/284 exits are blocked.

clavin-xlnx:
> Looks like my use of `test` is failing on Windows. My first approach was to do it entirely in Jython and use the exit code, however, due to #284 exits are blocked.

#700 should allow a fully Jython-native solution, we really only want to prevent unintentional exits from the interactive shell.

eddieh-xlnx:
Rebased on #700, and using a pure Python solution it appears to work!

clavin-xlnx (approved):

PR #706, opened 2023-06-14
UltraScale Incremental Clock Router fixes

eddieh-xlnx:
* Adds ability to be aware of (and thus re-use) existing routing resources already being used to route clock net
* Fixes computation of centroid for incremental case
* Fixes preservation of resources used by existing clock and static routes


clavin-xlnx:
There are some changes in the review that are part of already merged PRs--perhaps merge in `master` to `incr_clk_routing`?

clavin-xlnx (approved):

PR #710, opened 2023-06-15
Test that SitePinInst.detachSiteInst() preserves sitewire

eddieh-xlnx:
When `SitePinInst` being detached is not actually the SPI that is attached to the `SiteInst`.


clavin-xlnx (approved):
You've done a better job than me about this, but do we need to update the copyright year?

PR #715, opened 2023-06-16
Add test for Tile.getWireConnections()

eddieh-xlnx:
This test fails because `Tile.getWireConnections()` throws an exception.

@clavin-xlnx Can you take a look please?


eddieh-xlnx:
> I am more inclined to simply remove support for `xcku025` because of the number of issues we run into with it.

The reason I'm on `xcku025` was because of RapidWrightDCP's `microblazeAndILA_3pblocks.dcp` testcase. We'd have to migrate that away too.

clavin-xlnx:
@eddieh-xlnx - please see https://github.com/eddieh-xlnx/RapidWrightDCP/pull/23

clavin-xlnx (approved):

PR #721, opened 2023-06-21
1.8 compliance

clavin-xlnx:
`Path.of()` appeared in Java 11 and is not in Java 1.8, the equivalent is `Paths.get()`.  This is not caught in `gradlew` or Eclipse when setting compiler compliance to 1.8.  This was only detected with using a 1.8 JDK.


eddieh-xlnx (approved):
Good catch! Is there an automated way we can catch this earlier next time?

clavin-xlnx:
> Good catch! Is there an automated way we can catch this earlier next time?

The only way I know so far, is to use an actual 1.8 JDK.  It seems like all the `-source` and `-target` options do not catch this.

PR #729, opened 2023-06-27
Unroute site routing when removing a cell

clavin-xlnx:
Improves the behavior of `DesignTools.makeBlackBox()` such that is more cleanly removes site routing when a black boxed components removed cells share sites with logic that will remain in the design.  This adds a new method that aims to correctly unroute site routing for a given pin of a placed and routed cell.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -1492,7 +1492,7 @@ public class DesignTools {
     }
 
     /**
-     * Unroutes the site routing connected the provided cell's logical pin.
+     * Unroutes the site routing connected to the provided cell's logical pin.
      * Preserves other parts of the net if used by other sinks in the site if an
      * input. For the unrouting to be successful, this method depends on the site
      * routing to be consistent.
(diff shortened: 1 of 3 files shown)


eddieh-xlnx (approved):

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1490,6 +1490,157 @@ public static void makeBlackBox(Design d, String hierarchicalCellName) {
...
+    /**
+     * Unroutes the site routing connected to the provided cell's logical pin.
+     * Preserves other parts of the net if used by other sinks in the site if an
+     * input. For the unrouting to be successful, this method depends on the site
+     * routing to be consistent.
+     * 
+     * @param cell           The cell of the pin
+     * @param logicalPinName The logical pin name source or sink to have routing
+     *                       removed.
+     * @returns A list of site pins that should be removed as part of the unroute.
```
eddieh-xlnx, line 1503:
```suggestion
     * @returns A list of site pins that should also be removed from inter-site routing to complete the unroute.
```

PR #703, opened 2023-06-13
Fix DesignTools.makePhysNetNamesConsistent() 

eddieh-xlnx:
By transforming logical static nets named `<const{0,1}>` into physical static nets `GLOBAL_LOGIC{0,1}`

Fixes #701.

CC: @haydenc-amd

PR #734, opened 2023-06-29
DesignTools.makePhysNetNamesConsistent() to consider */<const{0,1}>

eddieh-xlnx:
Look at the non-hier logical net name to determine if it's a static net.

https://github.com/Xilinx/RapidWright/pull/703 previously only considered `<const{0,1}>` logical nets in the top cell; this PR considers their appearance in any cell.

Greater discussion should be had as to whether we should rely solely on the assumption that all logical nets with a name of `<const{0,1}>` are GND/VCC nets. A more robust solution would be to examine the this logical net to find its source pin, and make sure it really is driven by the `GND` or `VCC` primitive; however, this would be a linear search on every single processed net.

We could take a shortcut and use a different assumption that static nets must be driven by a cell instance called `GND` or `VCC` (allowing a binary search into `EDIFNet.portInsts`) but I'm not sure that's much better.

Alternatively, `EDIFNetlist.generateParentNetMap()` already does computation to collect a list GND and VCC logical port instances, perhaps it can be extended to track static nets too?

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -2810,16 +2810,17 @@ public class DesignTools {
                 continue;
             }
             if (!hierNet.equals(parentHierNet)) {
-                String parentHierNetName = parentHierNet.getHierarchicalNetName();
+                String parentNetName = parentHierNet.getNet().getName();
                 Net parentPhysNet;
-                if (parentHierNetName.equals(EDIFTools.LOGICAL_VCC_NET_NAME)) {
-                    parentHierNetName = Net.VCC_NET;
+                // Assume that a net named <const1> or <const0> is always a VCC or GND net
+                if (parentNetName.equals(EDIFTools.LOGICAL_VCC_NET_NAME)) {
+                    parentNetName = Net.VCC_NET;
                     parentPhysNet = design.getVccNet();
-                } else if (parentHierNetName.equals(EDIFTools.LOGICAL_GND_NET_NAME)) {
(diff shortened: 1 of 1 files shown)


clavin-xlnx:
> Alternatively, `EDIFNetlist.generateParentNetMap()` already does computation to collect a list GND and VCC logical port instances, perhaps it can be extended to track static nets too?

There are a few lists of pins that are generated as part of `EDIFNetlist.generateParentNetMap()`:

https://github.com/Xilinx/RapidWright/blob/beae92fd0648484891b298bcdb99fee87a10202f/src/com/xilinx/rapidwright/edif/EDIFNetlist.java#L95-L96

These should contain at least every sink pin connected to GND and VCC throughout the design.  I don't think it would be too hard to add an entry for all the logical nets, however, the storage of that might be non-trivial for some designs and could impact performance.

clavin-xlnx (approved):

clavin-xlnx:
> Greater discussion should be had as to whether we should rely solely on the assumption that all logical nets with a name of `<const{0,1}>` are GND/VCC nets. A more robust solution would be to examine the this logical net to find its source pin, and make sure it really is driven by the `GND` or `VCC` primitive; however, this would be a linear search on every single processed net.
> 
> We could take a shortcut and use a different assumption that static nets must be driven by a cell instance called `GND` or `VCC` (allowing a binary search into `EDIFNet.portInsts`) but I'm not sure that's much better.

I think we can rely on it for now.  It is legal to create nets with those names that aren't driven by GND/VCC, but let's wait and see if we encounter a legitimate design in the wild before we start building out support for that use case.

PR #751, opened 2023-07-12
Fix 748 - Updates RAM32X1S property to correct default

### LINKED ISSUE #748: RW reports CLK_INVERTED as 1'b1 when it is reported as 1'b0 in Vivado
This error happens if just a couple designs out of the 35 designs that I am running. Basically, sometimes on a RAM32S primitive, rapidwright reports that the CLK_INVERTED property value is "1'b1", when in Vivado it is reported as "1'b0". The device viewer also shows the net being routed through the normal clk input on the slice.
Attached is python code/dcp to replicate and a screenshot of the vivado cell properties,
I am using Vivado 2023.1, but I do believe I ran into this error with checkpoints generated in Vivado 2022.2, I am just now getting around to submitting the issue.
[clk_inverted_py.txt](https://github.com/Xilinx/RapidWright/files/12015330/clk_inverted_py.txt)
vivado_viewer
[design.zip](https://github.com/Xilinx/RapidWright/files/12015332/design.zip)
[design_edf.txt](https://github.com/Xilinx/RapidWright/files/12015333/design_edf.txt)

clavin-xlnx:
Fixes #748 by updating `unisim_data.dat` with the appropriate setting for `IS_CLK_INVERTED`.

### DIFF
--- a/src/com/xilinx/rapidwright/util/DataVersions.java
+++ b/src/com/xilinx/rapidwright/util/DataVersions.java
@@ -351,6 +351,6 @@ public class DataVersions {
         dataVersionMap.put("data/devices/zynquplusrfsoc/xqzu49dr_db.dat", new Pair<>("xqzu49dr-db-dat", "ff145122f2caea588120ac4da439d838"));
         dataVersionMap.put("data/partdump.csv", new Pair<>("partdump-csv", "f5aa2d4c08196e330576bda0e644772a"));
         dataVersionMap.put("data/parts.db", new Pair<>("parts-db", "f5d8a26178d35bd1a68c441b37a04279"));
-        dataVersionMap.put("data/unisim_data.dat", new Pair<>("unisim-data-dat", "ce8410650ad739c8630b12c3152693a3"));
+        dataVersionMap.put("data/unisim_data.dat", new Pair<>("unisim-data-dat", "85322379ad298d5a710d24a731d151e8"));
     }
 }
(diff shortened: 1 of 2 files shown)


eddieh-xlnx:
Looking back at #748, what should RapidWright do if we instantiate a `RAM32X1S_1` instead here? I'm seeing that Vivado emits this:
```
create_cell -reference RAM32X1S_1 inst
WARNING: [Coretcl 2-1024] Primitives of type 'RAM32X1S_1' are not supported by the current part and have been retargeted to 'RAM32X1S'.
```

However, RapidWright doesn't retarget (not reasonable to expect it to retarget to `RAM32X1S`?) but it might be reasonable to expand this directly to the end result (where `inst/SP` is a `RAMS32`) with the `IS_CLK_INVERTED` set correctly?

I'll push my modified test; we can revert if it's out of scope of this PR.

eddieh-xlnx:
That's reasonable, let's revert my test. In this case, would we expect the `IS_CLK_INVERTED` property of a `RAM32X1S_1` instance in RapidWright to be unreliable then?

clavin-xlnx:
> That's reasonable, let's revert my test. In this case, would we expect the `IS_CLK_INVERTED` property of a `RAM32X1S_1` instance created by/loaded into RapidWright to be unreliable then?

That is correct.

eddieh-xlnx (approved):
Changed unsupported macros into a canary test.

PR #754, opened 2023-07-15
Test that Cell.getProperty() returns null when no EDIFCellInst

eddieh-xlnx:


test/src/com/xilinx/rapidwright/design/TestCell.java
```diff
@@ -74,4 +74,11 @@ public void testGetCorrespondingSitePinName() {
         
         Assertions.assertEquals(pin.getName(), cell.getCorrespondingSitePinName("D"));
     }
+
+    @Test
+    public void testGetPropertyNoEDIFCellInst() {
+        Cell cell = new Cell("cell");
+        Assertions.assertNull(cell.getEDIFCellInst());
+        Assertions.assertNull(cell.getProperty("a_property_that_doesnt_exist"));
```
eddieh-xlnx, line 82:
```suggestion
        Assertions.assertNull(cell.getProperty("any_property"));
```

clavin-xlnx (approved):

PR #755, opened 2023-07-15
Add DesignTools.getAllRoutedSitePinsFromPhysicalPin()

eddieh-xlnx:
Prior, the only method was `DesignTools.getRoutedSitePinFromPhysicalPin()` which (a) only returned the first site pin, and (b) returned the other output site pin by adding it to the `Net` and thus having a side effect which has caused problems.

This side effect was actually causing an assumption in RWRoute (that the `COUT` pin was always the primary source, with a `HMUX` pin a possible alternate) to break since `DesignTools.getRoutedSitePinFromPhysicalPin()` actually set `HMUX` as the primary source. 

Note that `DesignTools.createMissingSitePinInsts()` -- which calls `DesignTools.getRoutedSitePinFromPhysicalPin()` -- will now not infer any dual output pins. If those are needed, they can be explicitly gotten, as RWRoute does.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -1313,8 +1313,7 @@ public class DesignTools {
             }
             if (otherUser == false) {
                 // Unroute site routing back to pin and remove site pin
-                String sitePinName = getRoutedSitePinFromPhysicalPin(cell, net, pin.getName());
-                if (sitePinName != null) {
+                for (String sitePinName : getAllRoutedSitePinsFromPhysicalPin(cell, net, pin.getName())) {
                     BELPin sitePortBelPin = siteInst.getSite().getBELPin(sitePinName);
                     assert(sitePortBelPin.isSitePort());
                     boolean outputSitePin = sitePortBelPin.isInput(); // Input BELPin means output SitePin
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -2121,13 +2121,12 @@ public class DesignTools {
     }
 
     /**
-     * Gets the site pin that is currently routed to the specified cell pin.  If
+     * Gets the first site pin that is currently routed to the specified cell pin.  If
      * the site instance is not routed, it will return null.
-     * Side Effect: It will set alternative source site pins on the net if present.
      * @param cell The cell with the pin of interest.
      * @param net The physical net to which this pin belongs
      * @param logicalPinName The logical pin name of the cell to query.
-     * @return The name of the site pin on the cell's site to which the pin is routed.
+     * @return The name of the site pin name on the cell's site to which the pin is routed.
      */
     public static String getRoutedSitePin(Cell cell, Net net, String logicalPinName) {
         String belPinName = cell.getPhysicalPinMapping(logicalPinName);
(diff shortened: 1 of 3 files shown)


eddieh-xlnx:
Depends on a feature fixed in upcoming 2023.1.2 to do with dual output nets.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2121,33 +2120,45 @@ public static List<SitePinInst> createMissingSitePinInsts(Design design, Net net
...
     /**
-     * Gets the site pin that is currently routed to the specified cell pin.  If
+     * Gets the first site pin that is currently routed to the specified cell pin.  If
      * the site instance is not routed, it will return null.
-     * Side Effect: It will set alternative source site pins on the net if present.
      * @param cell The cell with the pin of interest.
      * @param net The physical net to which this pin belongs
      * @param logicalPinName The logical pin name of the cell to query.
-     * @return The name of the site pin on the cell's site to which the pin is routed.
+     * @return The name of the site pin name on the cell's site to which the pin is routed.
```
eddieh-xlnx, line 2128:
```suggestion
     * @return The name of the first site pin on the cell's site to which the pin is routed.
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2121,33 +2120,45 @@ public static List<SitePinInst> createMissingSitePinInsts(Design design, Net net
...
     /**
-     * Gets the site pin that is currently routed to the specified cell pin.  If
+     * Gets the first site pin that is currently routed to the specified cell pin.  If
      * the site instance is not routed, it will return null.
-     * Side Effect: It will set alternative source site pins on the net if present.
      * @param cell The cell with the pin of interest.
      * @param net The physical net to which this pin belongs
      * @param belPinName The physical pin name of the cell
-     * @return The name of the site pin on the cell's site to which the pin is routed.
+     * @return The name of the first site pin name on the cell's site to which the pin is routed.
```
eddieh-xlnx, line 2141:
```suggestion
     * @return The name of the first site pin on the cell's site to which the pin is routed.
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2156,18 +2167,18 @@ public static String getRoutedSitePinFromPhysicalPin(Cell cell, Net net, String
...
                 BELPin source = curr.getSourcePin();
-                if (source == null) return null;
+                if (source == null) return Collections.emptyList();
                 if (source.isSitePort()) {
-                    return source.getName();
+                    return Collections.singletonList(source.getName());
                 } else if (source.getBEL().getBELClass() == BELClass.RBEL) {
                     SitePIP sitePIP = inst.getUsedSitePIP(source.getBELName());
-                    if (sitePIP == null) return null;
+                    if (sitePIP == null) return Collections.emptyList();
```
eddieh-xlnx, line 2181:
```suggestion
                    if (sitePIP == null) continue;
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3208,7 +3202,9 @@ public static void createMissingStaticSitePins(BELPin belPin, SiteInst si, Cell
         Net net = (netOnSiteWire != null) ? netOnSiteWire : si.getDesign().getVccNet();
         if (net.isStaticNet()) {
             // SRL16Es that have been transformed from SRLC32E require GND on their A6 pin
-            if (cell.getType().equals("SRL16E") && siteWireName.endsWith("6")) {
+            if (cell.getType().equals("SRL16E") && siteWireName.endsWith("6") &&
+                    //
```
clavin-xlnx, line 3206: Missing thought?

clavin-xlnx (approved):

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: This is a workaround for the symptom in https://github.com/Xilinx/RapidWright/pull/754 that shouldn't have slipped in here. Good spot!

PR #759, opened 2023-07-17
[PartialRouter] Improve incremental global routing

eddieh-xlnx:
1. PartialRouter with `softPreserve` enabled allows the global router to "steal" nodes from preserved non-global nets in order to complete its route. When this happens, the new `PartialRouter.unpreserveCongestedNets()` will recognize those victim non-globals and force them to be rerouted.
2. RWRoute and PartialRouter to call `preserveNet` for all nets, irrespective of whether it has any PIPs on or not, in order to preserve all nodes associated with its pins.
3. Global router to be more careful about stealing nodes that may serve as a used site pin
4. Do not rely on `SitePinInst.isRouted()` during initialization, since that flag is not currently set correctly during `readCheckpoint`.

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/PartialRouter.java
+++ b/src/com/xilinx/rapidwright/rwroute/PartialRouter.java
@@ -292,7 +292,7 @@ public class PartialRouter extends RWRoute {
                         unpreserveNets.add(preservedNet);
                     }
 
-                    // Redo preserving clk
+                    // Preserve node for global net
                     Net oldNet = routingGraph.preserve(node, net);
                     if (oldNet != null) {
                         // oldNet/preservedNet is not a routable net (e.g. driven by hier port)
@@ -309,8 +309,8 @@ public class PartialRouter extends RWRoute {
                         rnode.clearPrev();
 
                         // Increment this RouteNode with a null net (since global nets have
-                        // no corresponding NetWrapper) in order to flag it as being permanently
-                        // overused
+                        // no corresponding NetWrapper) in order to flag it as being irreversibly
+                        // used by a global net thus forcing the non-global to find another path
                         rnode.incrementUser(null);
                     } else {
                         assert(oldNet != null);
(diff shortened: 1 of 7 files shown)


src/com/xilinx/rapidwright/rwroute/PartialRouter.java
```diff
@@ -218,47 +243,82 @@ protected void routeGlobalClkNets() {
...
+                        continue;
+                    }
+                    if (preservedNet == null) {
+                        // Assume this node has already been unpreserved
+                    } else if (RouterHelper.isRoutableNetWithSourceSinks(preservedNet)) {
+                        unpreserveNet(preservedNet);
+                        unpreserveNets.add(preservedNet);
+                    }
+
+                    // Redo preserving clk
```
clavin-xlnx, line 295: This comment might need to be reworded based on the change from `clkNets -> globalNets`.

clavin-xlnx (approved):

src/com/xilinx/rapidwright/rwroute/PartialRouter.java
eddieh-xlnx, line 295:
```suggestion
                    // Preserve node for global net
```

src/com/xilinx/rapidwright/rwroute/PartialRouter.java
```diff
@@ -218,47 +243,82 @@ protected void routeGlobalClkNets() {
...
-                }
-            }
 
-            if (!unpreserveNets.isEmpty()) {
-                System.out.println("INFO: Unpreserving " + unpreserveNets.size() + " nets due to clock congestion");
-                for (Net net : unpreserveNets) {
-                    System.out.println("\t" + net);
+                        // Increment this RouteNode with a null net (since global nets have
+                        // no corresponding NetWrapper) in order to flag it as being permanently
+                        // overused
```
eddieh-xlnx, line 313:
```suggestion
                        // no corresponding NetWrapper) in order to flag it as being irreversibly
                        // used by a global net
```

PR #760, opened 2023-07-19
Shell creation improvements to enable lock_design and timing closure preservation

clavin-xlnx:
- When creating a blackbox from the command line, also will prohibit partially used half slices from being used as it seems the Vivado placer struggles to put cells in those locations without considering downstream congestion issues
- Adds lock_design methods  (netlist, placement and routing) that attempt to mimic the `lock_design -level {netlist, placement, routing}` Tcl command in Vivado.  Some minor variations exist, but should accomplish the same end

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -1671,7 +1671,8 @@ public class DesignTools {
      * In Vivado, this can be accomplished by running: (1)
      * {@code update_design -cells <name> -black_box} or (2) by deleting all of the
      * cells and nets insides of a cell instance. Method (2) is more likely to have
-     * complications.
+     * complications. This also unroutes both GND and VCC nets to avoid
+     * implementation issues by Vivado in subsequent place and route runs.
      * 
      * @param d                The current design
      * @param hierarchicalCell The hierarchical cell to become a black box.
@@ -3619,8 +3620,9 @@ public class DesignTools {
     }
 
     /**
-     * Locks the logical netlist of the design using the DONT_TOUCH property.
-     * {@link EDIFTools#lockNetlist(EDIFNetlist)}.
+     * Locks the logical netlist of the design using the DONT_TOUCH property. This
+     * strives to be as close as possible to what Vivado's 'lock_design -level
+     * netlist' does to lock the design. {@link EDIFTools#lockNetlist(EDIFNetlist)}.
      * 
      * @param design The design of the netlist to lock.
      */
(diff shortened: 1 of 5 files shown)


src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -1485,4 +1488,62 @@ Iterable<Entry<T,U>> sortIfStable(Map<T,U> collection, boolean stable) {
...
+
+    /**
+     * Traversal helper method to {@link #lockNetlist(EDIFNetlist)} and
+     * {@link #unlockNetlist(EDIFNetlist)}.
+     * 
+     * @param netlist    The netlist on which to apply the action.
+     * @param lockAction The method of action (lock or unlock) to apply to
+     *                   {@link EDIFPropertyObject} objects.
+     */
+    private static void lockNetlist(EDIFNetlist netlist, Function<EDIFPropertyObject, EDIFPropertyValue> lockAction) {
```
eddieh-xlnx, line 1524: Looks like this lambda doesn't need to return anything?

```suggestion
    private static void lockNetlist(EDIFNetlist netlist, Consumer<EDIFPropertyValue> lockAction) {
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3613,4 +3617,243 @@ public static boolean isNetDrivenByHierPort(Net net) {
 
         return true;
     }
+
+    /**
+     * Locks the logical netlist of the design using the DONT_TOUCH property.
```
eddieh-xlnx, line 3622: I think it would be documenting whether this (and the following methods) are identical/not identical to Vivado `lock_design`?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1667,18 +1667,22 @@ public static List<SitePinInst> unrouteCellPinSiteRouting(Cell cell, String logi
...
-     * this can be accomplished by running: (1) {@code update_design -cells <name> -black_box} or (2)
-     * by deleting all of the cells and nets insides of a cell instance.  Method (2) is
-     * more likely to have complications.
-     * @param d The current design
+     * associated placement and routing information associated with that instance.
+     * In Vivado, this can be accomplished by running: (1)
+     * {@code update_design -cells <name> -black_box} or (2) by deleting all of the
+     * cells and nets insides of a cell instance. Method (2) is more likely to have
+     * complications.
+     *
```
eddieh-xlnx, line 1675: Document that this this will unroute all ground/vcc routing too?

eddieh-xlnx (approved):

PR #762, opened 2023-07-20
Add DesignTools.updatePinsIsRouted()

eddieh-xlnx:
Walk forward -- using a Net's PIPs -- from all its sources to determine which of its sinks are reachable and thus should be marked routed.

Works for static nets too.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -3636,17 +3636,28 @@ public class DesignTools {
         return true;
     }
 
+    /**
+     * Update the SitePinInst.isRouted() value of all sink pins on the given
+     * Net. A pin will be marked as being routed if it is reachable from the
+     * Net's source pins (or in the case of a static net, a node tied to GND
+     * or VCC) when following the Net's PIPs.
+     * @param net Net on which pins are to be updated.
+     */
     public static void updatePinsIsRouted(Net net) {
         Queue<Node> queue = new ArrayDeque<>();
         Map<Node, List<Node>> node2fanout = new HashMap<>();
+        Map<Node, Node> bidirNode2node = new HashMap<>();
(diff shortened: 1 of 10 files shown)


clavin-xlnx (approved):
Thanks for writing this!

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3601,4 +3602,65 @@ public static boolean isNetDrivenByHierPort(Net net) {
...
+            Node node = queue.poll();
+            SitePinInst spi = node2spi.get(node);
+            if (spi != null) {
+                spi.setRouted(true);
+            }
+
+            List<Node> fanouts = node2fanout.get(node);
+            if (fanouts != null) {
+                for (Node fanout : fanouts) {
+                    // In the case of bidir PIPs, do not go back on yourself by re-enqueue
```
eddieh-xlnx, line 3657:
```suggestion
                    // In the case of bidir PIPs, do not go back on yourself by re-enqueuing
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3890,5 +3891,67 @@ public static void addProhibitConstraint(Design design, List<String> belLocation
...
                     "set_property PROHIBIT true [get_bels { " + sb.toString() + "} ]");
         }
-    }  
+    }
+
+    /**
+     * Update the SitePinInst.isRouted() value of all sink pins on the given
+     * Net. A pin will be marked as being routed if it is reachable from the
+     * Net's source pins (or in the case of a static net, a node tied to GND
+     * or VCC) when following the Net's PIPs.
```
eddieh-xlnx, line 3900:
```suggestion
     * Net's source pins (or in the case of static nets, also from nodes
     * tied to GND or VCC) when following the Net's PIPs.
```

PR #763, opened 2023-07-20
Fix module relocation in the face of tile node depopulation

clavin-xlnx:
Some tiles around IO columns can have certain nodes depopulated from the standard set.  This breaks the relocatability of some modules when the routes use those nodes in other contexts.  This change detects the depopulated nodes and ensures the module cannot be placed in those locations.

For example: 
The tile `INT_X94Y899` in the `xcvu13p` device is fully populated:
image

However, the tile `INT_X74Y589` in the same device is not:
image

### DIFF
--- a/src/com/xilinx/rapidwright/design/ModuleInst.java
+++ b/src/com/xilinx/rapidwright/design/ModuleInst.java
@@ -389,6 +389,11 @@ public class ModuleInst extends AbstractModuleInst<Module, Site, ModuleInst>{
                 if (pipSet != null) {
                     pipSet.add(newPip);
                 }
+                // Some tiles have nodes that are depopulated, we need to detect those
+                if (newPip.getEndNode().getAllDownhillPIPs().size() == 0
+                        && pip.getEndNode().getAllDownhillPIPs().size() != 0) {
+                    return false;
+                }
                 net.addPIP(newPip);
             }
         }
--- a/src/com/xilinx/rapidwright/design/ModuleInst.java
+++ b/src/com/xilinx/rapidwright/design/ModuleInst.java
@@ -34,6 +34,7 @@ import com.xilinx.rapidwright.design.blocks.PBlock;
 import com.xilinx.rapidwright.design.blocks.PBlockRange;
 import com.xilinx.rapidwright.design.noc.NOCClient;
 import com.xilinx.rapidwright.device.Device;
+import com.xilinx.rapidwright.device.Node;
 import com.xilinx.rapidwright.device.PIP;
 import com.xilinx.rapidwright.device.Site;
 import com.xilinx.rapidwright.device.SiteTypeEnum;
@@ -390,7 +391,8 @@ public class ModuleInst extends AbstractModuleInst<Module, Site, ModuleInst>{
                     pipSet.add(newPip);
                 }
                 // Some tiles have nodes that are depopulated, we need to detect those
-                if (newPip.getEndNode().getAllDownhillPIPs().size() == 0
+                Node endNode = newPip.getEndNode();
+                if (endNode != null && endNode.getAllDownhillPIPs().size() == 0
                         && pip.getEndNode().getAllDownhillPIPs().size() != 0) {
                     return false;
                 }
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -389,6 +390,12 @@ public boolean place(Site newAnchorSite, boolean skipIncompatible, boolean allow
                 if (pipSet != null) {
                     pipSet.add(newPip);
                 }
+                // Some tiles have nodes that are depopulated, we need to detect those
+                Node endNode = newPip.getEndNode();
+                if (endNode != null && endNode.getAllDownhillPIPs().size() == 0
+                        && pip.getEndNode().getAllDownhillPIPs().size() != 0) {
```
eddieh-xlnx, line 396: Is looking at the existence of downhill PIPs the best way to check for depopulation? It seems that the new PIP has its tile forcibly updated, but is there a more general way to check for the validity of this new PIP?

clavin-xlnx: I'm open to other suggestions, but I couldn't find another way.

eddieh-xlnx: I can't see another way without adding a new method onto `Node` either, and it's not clear to me what should be checked when inside too. I have a nagging worry about performance here since every placement attempt (which `getAllValidPlacements` does) that gets through to routing will involve examining all downhill nodes of all PIPs at least once.

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -389,6 +390,12 @@ public boolean place(Site newAnchorSite, boolean skipIncompatible, boolean allow
                 if (pipSet != null) {
                     pipSet.add(newPip);
                 }
+                // Some tiles have nodes that are depopulated, we need to detect those
+                Node endNode = newPip.getEndNode();
+                if (endNode != null && endNode.getAllDownhillPIPs().size() == 0
```
eddieh-xlnx, line 395: Do you know if `Node`s can be partially depopulated, or that they must either be fully depopulated or fully populated with no middle ground?

eddieh-xlnx (approved):
This fixes an immediate need, so let's merge it, and revisit performance if we see it become a problem.

src/com/xilinx/rapidwright/design/ModuleInst.java
clavin-xlnx: I know performance is an issue, but it may not matter for this particular application.  `getAllValidPlacements()` is or can be called offline and results cached in the module.  Then, during placement, only the collisions between it and the logic already placed needs to be checked.  The block placer does a number of optimizations during placement to avoid ever calling the actual `place()` method because it is expensive.  For example, for a modestly small module, it takes ~3.2 secs to call `getAllValidPlacements()`, it calls `place()` 117120 times, 38400 being successful.  Before this PR, it took ~2.5 secs. 

I spent some time looking into ways to cache the architectural special cases, but I'm not sure the effort is warranted yet.

src/com/xilinx/rapidwright/design/ModuleInst.java
clavin-xlnx: The problem here arises because the module uses a node in the original placed context that spans a tile type that is different in the problematic placed context.  The tile type simply doesn't have the connection as it did in the original context.  That means the tile/wire participates in a different node.

eddieh-xlnx: Sorry, let me reprhase my question: Can a node simply have one arc depopulated, or must it have all arcs depopulated (which is what the above code checks for)? The former would be painful to identify.

clavin-xlnx: I have not performed an exhaustive architectural search so I don't know.  There are actually many differences when comparing one `INT` tiles node fanout to another.  The reason this one became an issue was because the fanout involved a tile that had no other presence from the module except the fact that this node passed through it.  Most of the time, the module's footprint is pervasive in that it covers all the surrounding tiles so that if there is a change in the fanout, the tile type will have already been compared and the candidate location would be disqualified.

PR #771, opened 2023-07-27
Add flag to filter static sources

clavin-xlnx:


clavin-xlnx:
Similar changes were added in https://github.com/Xilinx/RapidWright/pull/1006.

Issue #579, opened 2022-11-10
RWRoute to set PIP.setIsReversed() on routed nets


clavin-xlnx:
When using a bidirectional PIP in the direction opposite its name, it will need `PIP.setIsReversed()` set in order for other methods to work correctly (see #575).

clavin-xlnx:
Fixed in #774

PR #774, opened 2023-08-02
Set reversed flag on bi-directional PIPs used from end->start

clavin-xlnx:
Previously, RWRoute was not setting the isReversed() flag on `PIP` objects.  This PR makes it so that all `PIP`s are properly set.  This also found that when routing static nets, the `PIP` creation process was using a less efficient method as it was presenting the connection nodes in order (rather than reverse order in typical routing).  This also modifies how static `PIP`s are created to present the driver/load ordering correctly.

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
+++ b/src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
@@ -374,7 +374,7 @@ public class GlobalSignalRouting {
                         if (debug) System.out.println("  " + routingNode.toString());
                         routingNode = routingNode.getPrev();
                     }
-                    netPIPs.addAll(RouterHelper.getPIPsFromNodes(pathNodes));
+                    netPIPs.addAll(RouterHelper.getPIPsFromNodes(pathNodes, true));
 
                     // If the source is an output site pin, put it aside for consideration
                     // to add as a new source pin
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -374,7 +374,7 @@ public static void routeStaticNet(Net currNet,
                         if (debug) System.out.println("  " + routingNode.toString());
                         routingNode = routingNode.getPrev();
                     }
-                    netPIPs.addAll(RouterHelper.getPIPsFromNodes(pathNodes));
+                    netPIPs.addAll(RouterHelper.getPIPsFromNodes(pathNodes, true));
```
eddieh-xlnx, line 377:
```suggestion

                    // Note that the static net router goes backward from sinks to sources,
                    // requiring the srcToSinkOrder parameter to be set to true below
                    netPIPs.addAll(RouterHelper.getPIPsFromNodes(pathNodes, true));
```

eddieh-xlnx (approved):

eddieh-xlnx:
I believe this PR fix https://github.com/Xilinx/RapidWright/issues/579 right?

clavin-xlnx:
> I believe this PR fix #579 right?

Yes

PR #778, opened 2023-08-08
Make PartialRouter.getUnroutedPins() public

eddieh-xlnx:


clavin-xlnx (approved):
Should this be relocated to a more generic class?

eddieh-xlnx:
> Should this (eventually) be relocated to a more generic class?

This is only necessary in the very short-term. The proper way to do this would be to use https://github.com/Xilinx/RapidWright/pull/762 to set `SitePinInst.isRouted()` properly to begin with, and then keep it in sync.

PR #799, opened 2023-08-29
[PhysNetlistWriter] BELPin to check logical pin used in netlist

eddieh-xlnx:
Fixes #798.

Issue #798, opened 2023-08-29
FPGAIF physical net includes BEL pin that is not connected to net 


zakn-amd:
For example the physical net  `processor/alu_result_2 processor/lower_reg_banks/DIC0 processor/lower_reg_banks/DID0 processor/lower_zero_lut/I2` in [picoblaze_ooc_X10Y235.phys.txt](https://github.com/Xilinx/RapidWright/files/12467115/picoblaze_ooc_X10Y235.phys.txt) which was generated from the `picoblaze_ooc_X10Y235.dcp` distributed in rapidwright DCPs, looks like this:
```
belPin SLICE_X15Y237 G6LUT O6
{belPin SLICE_X15Y237 G_O G_O
sitePin SLICE_X15Y237 G_O
pip INT_X10Y237 LOGIC_OUTS_E20 INT_NODE_SDQ_29_INT_OUT1 False False
pip INT_X10Y237 INT_NODE_SDQ_29_INT_OUT1 NN2_E_BEG5 False False
pip INT_X10Y239 NN2_E_END5 INT_NODE_SDQ_30_INT_OUT1 False False
{pip INT_X10Y239 INT_NODE_SDQ_30_INT_OUT1 INT_INT_SDQ_34_INT_OUT1 False False
{pip INT_X10Y239 INT_INT_SDQ_34_INT_OUT1 INT_NODE_SDQ_82_INT_OUT0 False False
pip INT_X10Y239 INT_NODE_SDQ_82_INT_OUT0 SS2_W_BEG5 False False
pip INT_X10Y237 SS2_W_END5 INT_NODE_SDQ_80_INT_OUT0 False False
pip INT_X10Y237 INT_NODE_SDQ_80_INT_OUT0 SS2_W_BEG5 False False
pip INT_X10Y235 SS2_W_END5 INT_NODE_SDQ_80_INT_OUT1 False False
pip INT_X10Y235 INT_NODE_SDQ_80_INT_OUT1 WW2_W_BEG6 False False
pip INT_X9Y235 WW2_W_END6 INT_NODE_IMUX_53_INT_OUT1 False False
pip INT_X9Y235 INT_NODE_IMUX_53_INT_OUT1 BYPASS_W12 False False
pip INT_X9Y235 BYPASS_W12 INT_NODE_IMUX_45_INT_OUT1 False False
pip INT_X9Y235 INT_NODE_IMUX_45_INT_OUT1 BYPASS_W6 False False
[...]

PR #810, opened 2023-09-05
[PhysNetlistWriter] Skip cell pin mapping for unconnected pins

eddieh-xlnx:
Built on top of #799.

In #799, we omit BELPin-s that are in the physical routing if there is no logical connectivity. An example situation is seen in #798 where the LUT output is hard-wired to a placed CARRY8's input, but the placed cell doesn't actually care about that pin. 

Go one step further, by omitting pin mappings for such cells (even though Vivado has them). This provides a very obvious indication (even after converting back into the RapidWright data model) that this physical pin should not be considered (e.g. for timing analysis) without needing to access the logical netlist (if it was even available).

Casual experiment with Vivado shows that omitting this physical-to-logical pin mapping is accepted.

### DIFF
--- a/src/com/xilinx/rapidwright/interchange/PhysNetlistWriter.java
+++ b/src/com/xilinx/rapidwright/interchange/PhysNetlistWriter.java
@@ -174,7 +174,21 @@ public class PhysNetlistWriter {
                     j++;
                 }
             }
-            Builder<PinMapping.Builder> pinMap = physCell.initPinMap(cell.getPinMappingsP2L().size()
+
+            EDIFHierCellInst ehci = cell.getEDIFHierCellInst();
+            int numPinMappings;
+            if (ehci == null) {
+                numPinMappings = cell.getPinMappingsP2L().size();
+            } else {
+                numPinMappings = 0;
+                for (Entry<String, String> e : cell.getPinMappingsP2L().entrySet()) {
+                    String logicalPin = e.getValue();
+                    EDIFPortInst epi = ehci.getInst().getPortInst(logicalPin);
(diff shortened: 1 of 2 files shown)


clavin-xlnx (approved):

src/com/xilinx/rapidwright/interchange/PhysNetlistWriter.java
```diff
@@ -174,7 +174,21 @@ public static void writePlacement(PhysNetlist.Builder physNetlist, Design design
                     j++;
                 }
             }
-            Builder<PinMapping.Builder> pinMap = physCell.initPinMap(cell.getPinMappingsP2L().size()
+
+            EDIFHierCellInst ehci = cell.getEDIFHierCellInst();
+            int numPinMappings;
+            if (ehci == null) {
+                numPinMappings = cell.getPinMappingsP2L().size();
```
eddieh-xlnx, line 181:
```suggestion
                // Logical cell unavailable (e.g. logical netlist has been detached)
                // Assume connected and propagate all pin mappings
                numPinMappings = cell.getPinMappingsP2L().size();
```

src/com/xilinx/rapidwright/interchange/PhysNetlistWriter.java
```diff
@@ -174,7 +174,21 @@ public static void writePlacement(PhysNetlist.Builder physNetlist, Design design
...
+            } else {
+                numPinMappings = 0;
```
eddieh-xlnx, line 183:
```suggestion
                // Only write out pin mappings where a net is connected
                numPinMappings = 0;
```

eddieh-xlnx:
No longer necessary.

PR #811, opened 2023-09-07
Adds a tieoff method for module instances with unconnected inputs

clavin-xlnx:


src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -713,6 +720,78 @@ public void connect(String portName, int busIndex0, ModuleInst other, String oth
...
+            if (port.isOutPort())
+                continue;
+
+            for (SitePinInst pin : getCorrespondingPins(port)) {
+                if (pin == null) {
+                    System.err.println("ERROR: Problem encountered with finding corresponding SitePinInst on port "
+                            + getName() + "/" + port.getName() + " skipping...");
+                }
+                if (pin != null && (pin.getNet() == null || pin.getNet().getSource() == null)) {
+                    boolean useVccInverter = pin.getName().contains("RST");
```
eddieh-xlnx, line 739: This seems like an important condition to document the behaviour of.

The name of this boolean is `useVccInverter` but then when it gets passed to `connectToStaticNet()` it becomes its `connectToVcc` argument which does not seem to reconcile, because it doesn't look like this latter method does does any inversion?

clavin-xlnx: I've modified the code to clarify and harmonize the usage.

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -759,6 +764,80 @@ public void connect(String portName, int busIndex0, ModuleInst other, String oth
         }
     }
 
+    /**
+     * Connects unconnected SitePinInst inputs on the module instance to either GND
+     * or VCC (some reset input use VCC as an input and use the inverter to drive
+     * GND).
```
eddieh-xlnx, line 770:
```suggestion
     * or VCC (some reset inputs can accept VCC as an input and use its built-in
     * inverter to drive GND).
```

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -759,6 +764,80 @@ public void connect(String portName, int busIndex0, ModuleInst other, String oth
...
+                }
+                if (pin != null && (pin.getNet() == null || pin.getNet().getSource() == null)) {
+                    // If a port has a RST inverter, drive it with VCC to get GND
+                    boolean connectToVCC = pin.getName().contains("RST");
+                    if (verbose) {
+                        System.out.println(getName() + "/" + port.getName() + " SitePinInst=[" + pin
+                                + "] has no source, connecting to " + (connectToVCC ? "VCC" : "GND"));
+                    }
+
+                    connectToStaticNet(port.getName(), connectToVCC);
```
eddieh-xlnx, line 790: Do we know if simply connecting the SPI to VCC does what we expect? Does any intra-site routing/SitePIPs need to be set?
EDIT: Perhaps not relevant if we stop inverting here.

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -759,6 +764,80 @@ public void connect(String portName, int busIndex0, ModuleInst other, String oth
...
+        // Connect logically
+        EDIFCell parent = getCellInst().getParentCell();
+        EDIFNet logicalStaticNet = EDIFTools.getStaticNet(type, parent, getDesign().getNetlist());
+        EDIFPortInst portInst = getCellInst().getPortInst(portName);
+        if (portInst == null) {
+            logicalStaticNet.createPortInst(portName, getCellInst());
+        } else {
+            portInst.getNet().removePortInst(portInst);
+            logicalStaticNet.addPortInst(portInst);
+        }
```
eddieh-xlnx, line 823: When `connectToVCC` is true, this would also make the port logically connected to VCC (whereas if you're relying on the inverter, shouldn't it be GND?)

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -759,6 +764,80 @@ public void connect(String portName, int busIndex0, ModuleInst other, String oth
...
+                continue;
+
+            for (SitePinInst pin : getCorrespondingPins(port)) {
+                if (pin == null) {
+                    System.err.println("ERROR: Problem encountered with finding corresponding SitePinInst on port "
+                            + getName() + "/" + port.getName() + " skipping...");
+                }
+                if (pin != null && (pin.getNet() == null || pin.getNet().getSource() == null)) {
+                    // If a port has a RST inverter, drive it with VCC to get GND
+                    boolean connectToVCC = pin.getName().contains("RST");
```
eddieh-xlnx, line 784: Actually, I'm beginning to think this is a little risky. You are connecting all module (not primitive) inputs containing `RST` to VCC which means you can't rely that all fanouts of this `RST` port will have inverters. It may even be driving LUTs.

I think it's safer to rely on the `RouterHelper.invertPossibleGndPinsToVccPins()` method (called by default during `RWRoute.preprocess()`) to make this transformation on a primitive level.

PR #821, opened 2023-09-15
[RWRoute] Only add alternative sources to SiteInst if used

clavin-xlnx:
Adds the alternative source to the SiteInst if it is used for routing.


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -856,12 +855,19 @@ protected void postRouteProcess() {
                     // Set the routed state of the used source node
```
eddieh-xlnx, line 855:
```suggestion
                    // Set the routed state of the used source node
                    // and if used and not already present, add it to the SiteInst
```

eddieh-xlnx (approved):

PR #825, opened 2023-09-19
[RWRoute] Properly remove dual-output pins and reset routed flag

clavin-xlnx:
This fixes two issues:
1) Properly resets `SitePinInst.isRouted()` flag in the initial loop
2) Removes the alternate source pin if it was previously routed but is no longer used.

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/RWRoute.java
+++ b/src/com/xilinx/rapidwright/rwroute/RWRoute.java
@@ -836,9 +836,10 @@ public class RWRoute{
                 Net net = e.getKey();
                 SitePinInst source = net.getSource();
                 SitePinInst altSource = net.getAlternateSource();
+                boolean altSourcePreviouslyRouted = altSource != null ? altSource.isRouted() : false;
                 for (SitePinInst spi : Arrays.asList(source, altSource)) {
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -836,9 +836,10 @@ protected void postRouteProcess() {
                 Net net = e.getKey();
                 SitePinInst source = net.getSource();
                 SitePinInst altSource = net.getAlternateSource();
+                boolean altSourcePreviouslyRouted = altSource != null ? altSource.isRouted() : false;
                 for (SitePinInst spi : Arrays.asList(source, altSource)) {
                     if (spi != null) {
-                        source.setRouted(false);
+                        spi.setRouted(false);
```
eddieh-xlnx, line 842: Good catch! Sorry.

eddieh-xlnx (approved):

src/com/xilinx/rapidwright/rwroute/RWRoute.java
clavin-xlnx: I missed in my initial review too.

PR #834, opened 2023-09-28
Enable RWRoute to load Interchange designs from main()

clavin-xlnx:


src/com/xilinx/rapidwright/interchange/Interchange.java
```diff
@@ -50,6 +57,155 @@ public class Interchange {
...
+            }
+            tt.stop().printSummary();
+            return design;
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    /**
+     * Writes out a set of interchange files for the given design. It will write up
```
eddieh-xlnx, line 170:
```suggestion
     * Writes out a set of Interchange files for the given design. It will write up
```

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -1886,23 +1887,32 @@ protected static Design routeDesign(Design design, RWRoute router) {
...
+     * (DCP or FPGA Interchange), and parses the arguments for the
+     * {@link RWRouteConfig} object of the router.
+     * 
+     * @param args An array of strings that are used to create a
+     *             {@link RWRouteConfig} object for the router.
      */
     public static void main(String[] args) {
         if (args.length < 2) {
-            System.out.println("USAGE: <input.dcp> <output.dcp>");
+            System.out.println("USAGE: <input.dcp|input.netlist|input.phys> <output.dcp>");
```
eddieh-xlnx, line 1899: I worry there's a slight implication here that you can specify just a `*.netlist`, without a companion `*.phys` which is not true. A suggestion would be to say "FPGA Interchange Physical Netlist" and have it pick up the companion Logical Netlist if one exists. 

For non-timing driven RWRoute, a logical netlist is optional, but I think there needs to be a bit more work in `PhysNetlistReader` to make it work out of the box.

src/com/xilinx/rapidwright/util/StringTools.java
```diff
@@ -159,6 +159,20 @@ public static String removeLastSeparator(String s) {
...
+    /**
+     * Replaces the file extension of the provided file name. The current extension
+     * includes the final period '.' and all characters following it. If the file
+     * has no extension, it will add it as a suffix to the filename.
+     * 
+     * @param fileName     Name of the file to receive the updated extension.
+     * @param newExtension The new extension (should include starting period '.')
+     * @return The newly updated file name.
+     */
+    public static String replaceExtension(String fileName, String newExtension) {
```
eddieh-xlnx, line 171: Are you aware we have a `FileTools.replaceExtension(Path, String)` method (which admittedly has a different signature to this). Would it make sense to collect them together?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
clavin-xlnx: I've removed the logical netlist implication.

src/com/xilinx/rapidwright/util/StringTools.java
clavin-xlnx: I thought there was one but couldn't find it.  I've abandoned the `StringTools` version.

eddieh-xlnx (approved):

PR #839, opened 2023-10-03
Test that Design.createModuleInst() copies static sources

eddieh-xlnx:


test/src/com/xilinx/rapidwright/design/TestDesign.java
```diff
@@ -384,4 +384,27 @@ public void testMovePinsToNewNetDeleteOldNetIntraSiteOnly() {
...
         Assertions.assertTrue(newNet.getPIPs().isEmpty());
     }
+
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    public void testCreateModuleInstCopiesStaticSource(boolean unrouteStaticNets) {
+        Design design = new Design("testCreateModuleInstCopiesStaticSource", "xcku035");
+
+        Design microblaze = RapidWrightDCP.loadDCP("microblazeAndILA_3pblocks.dcp");
+        Assertions.assertNotNull(microblaze.getSiteInstFromSiteName("SLICE_X60Y116"));
```
clavin-xlnx, line 394: Perhaps we should store `"SLICE_X60Y116"` in a `String` since it is repeated 4 times?

clavin-xlnx (approved):

PR #841, opened 2023-10-03
ModuleInst.place() to check both RAMB36/RAMB18 sites for overlap

eddieh-xlnx:


clavin-xlnx (approved):
I recognize this problem is not elegant to solve, but I don't think there are many instances like this.

PR #844, opened 2023-10-05
Simplify and make DesignTools.updatePinsIsRouted() more robust

eddieh-xlnx:
By remove the a Node's `fanouts` immediately from the map, we are inherently more robust to the presence of loops. This is possible because each node should only be visited once anyway.

This also means we don't need to track and avoid bidirectional PIPs separately, since we can no longer visit nodes more than once.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -3897,10 +3897,11 @@ public class DesignTools {
     }
 
     /**
-     * Update the SitePinInst.isRouted() value of all sink pins on the given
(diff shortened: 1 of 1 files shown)


clavin-xlnx (approved):
IT seems like this method also marks source pins as routed as well.  Should we update the Javadoc accordingly?

eddieh-xlnx:
> It seems like this method also marks source pins as routed as well. Should we update the Javadoc accordingly?

Good suggestion! I've updated the doc to say that a source pin is marked as being routed if it drives at least one PIP. It does not check if that PIP leads to any/all sink pins, so it can't be used to tell if there's full connectivity or if that source pin is just driving an antenna.

PR #851, opened 2023-10-13
Use dummy LUTs instead of PROHIBITs

clavin-xlnx:
Alternative way to keep future placement away from partially used half-SLICEs.  This will create valid dummy LUTs and place them into partially used half SLICEs so that no other part of the LUTs in that half-SLICE can be used for placement.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -79,6 +79,7 @@ import com.xilinx.rapidwright.edif.EDIFPort;
 import com.xilinx.rapidwright.edif.EDIFPortInst;
 import com.xilinx.rapidwright.edif.EDIFPropertyValue;
 import com.xilinx.rapidwright.edif.EDIFTools;
+import com.xilinx.rapidwright.edif.EDIFValueType;
 import com.xilinx.rapidwright.placer.blockplacer.BlockPlacer2Impls;
 import com.xilinx.rapidwright.placer.blockplacer.ImplsInstancePort;
 import com.xilinx.rapidwright.placer.blockplacer.ImplsPath;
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3862,18 +3883,68 @@ public static void prohibitPartialHalfSlices(Design design) {
...
+                                        }
+                                    }
+
+                                    String name = "dummy_lut_" + dummyCount++;
+                                    Cell c = design.createAndPlaceCell(top, name, Unisim.LUT1,
+                                            si.getSiteName() + "/" + bel.getName());
+                                    c.getEDIFCellInst().addProperty("INIT", "1");
+                                    c.getEDIFCellInst().addProperty(EDIFTools.DONT_TOUCH, value);
+                                    c.removePinMapping("A6");
+                                    c.removePinMapping("A5");
```
eddieh-xlnx, line 3916: Is a LUT1 expected to have A6 and A5 pin mappings, or are you removing them both to cover both the `?6LUT` and `x5LUT` case?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3862,18 +3883,68 @@ public static void prohibitPartialHalfSlices(Design design) {
...
+                                    c.setBELFixed(true);
+                                    c.setSiteFixed(true);
+                                    boolean connected = false;
+                                    for (int i = 1; i < 5; i++) {
```
eddieh-xlnx, line 3920: Am I right in understanding that this block of code looks at `A1` to `A4` and attempts finds a pin with a static source, and connects that to `I0` of the newly created `LUT1`.
If no static source is found (i.e. when neither the `?6LUT` nor `?5LUT` BEL is occupied and thus no LUT inputs are used?) then this cell is removed again (and this BEL is not prohibited anymore?)

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx: Just removing both to cover both cases.  I could check for what type of cell it is, but that might be a tad bit slower.

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx:
```suggestion
                                    for (int i = 1; i <= 5; i++) {
```
Yes (although, I think it should be from `A1` to `A5`).  We only want to add a `LUT1` to block the cell if there is an unused pin.  If all the pins are already used, then there is no need to add a `LUT1` as any cell that could be placed there would have to have a strict subset of connections that are already being made and thus wouldn't add any changes to the routing burden of the design (at least on the input cones).

eddieh-xlnx: Okay, I think I understand now. Find the first unused pin or the first pin with a static source, and map it to `I0`. If there are no such pins, this means that all pins are used and the placer would not put something there anyway.

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: Right. It's the creation placing and placing of a LUT that creates such pin mappings by default, right?

clavin-xlnx: Yes, this removes the default mappings.

clavin-xlnx:
Closing for now due to complications in implementation in Vivado.

PR #856, opened 2023-10-20
Replace $(shell ...) with $(wildcard) and $(subst) in Makefile

eddieh-xlnx:
Fix for when the output of `$(shell ...)` being too long to be captured by a variable.


clavin-xlnx (approved):

clavin-xlnx:
Thanks for fixing this!

PR #863, opened 2023-10-26
RelocationTools fixes and more robust testing

### LINKED ISSUE #393: RelocationTools.revertRouting does not really revert
https://github.com/Xilinx/RapidWright/blob/a34c0d71521a758e757056714ff0503de28f0c71/src/com/xilinx/rapidwright/design/tools/RelocationTools.java#L304-L308

is effectively a NOOP because it's modifying the list of references (to the same set of references) but not actually changing the referent.

eddieh-xlnx:
Fixes #393 too, a feature that is no longer relevant as of #808.

### DIFF
--- a/src/com/xilinx/rapidwright/design/tools/RelocationTools.java
+++ b/src/com/xilinx/rapidwright/design/tools/RelocationTools.java
@@ -247,7 +247,7 @@ public class RelocationTools {
             if (src != null && !oldSite.containsKey(src.getSiteInst())) {
                 for (SitePinInst spi : pins) {
                     if (oldSite.containsKey(spi.getSiteInst())) {
-                        // Source is not reloacted, but at least one pin inside site insts to be relocated
+                        // Source is not relocated, but at least one pin inside site insts to be relocated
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/design/tools/RelocationTools.java
```diff
@@ -238,25 +235,27 @@ public static boolean relocate(Design design,
...
 
+            Collection<SitePinInst> pins = n.getPins();
             SitePinInst src = n.getSource();
             if (src != null && !oldSite.containsKey(src.getSiteInst())) {
-                System.out.println("INFO: Unrouting Net '" + n.getName() + "' since output SiteInstPin '" +
-                        src + "' does not belong to SiteInsts to be relocated");
-                n.unroute();
+                for (SitePinInst spi : pins) {
+                    if (oldSite.containsKey(spi.getSiteInst())) {
+                        // Source is not reloacted, but at least one pin inside site insts to be relocated
```
clavin-xlnx, line 250:
```suggestion
                        // Source is not relocated, but at least one pin inside site insts to be relocated
```

clavin-xlnx (approved):

PR #865, opened 2023-10-30
Add RouteThruHelper.isRouteThruPIPAvailable(Design, Node, Node) overload

eddieh-xlnx:
With tests.

### DIFF
--- a/src/com/xilinx/rapidwright/router/RouteThruHelper.java
+++ b/src/com/xilinx/rapidwright/router/RouteThruHelper.java
@@ -153,6 +153,10 @@ public class RouteThruHelper {
         }
     }
 
+    /**
+     * Given a routethru PIP check that it is available for use by checking for net and cell
+     * collisions on the site it is routing through.
+     */
     public static boolean isRouteThruPIPAvailable(Design design, PIP routethru) {
         if (!routethru.isRouteThru()) return false;
         return isRouteThruPIPAvailable(design, routethru.getStartWire(), routethru.getEndWire());
@@ -173,6 +177,13 @@ public class RouteThruHelper {
         return true;
     }
 
+    /**
+     * Given two Wire objects (assumed to make up a routethru PIP) check that this
+     * PIP is available for use by checking for net and cell collisions within the site
+     * it is routing through.
+     * Note that this method is identical to the {@link #isRouteThruPIPAvailable(Design, Node, Node)
+     * overload, kept separate to minimize unnecessary calling Node.getSitePin().}
+     */
     public static boolean isRouteThruPIPAvailable(Design design, Wire start, Wire end) {
         SitePin outPin = end.getSitePin();
         if (!isRouteThruSitePinAvailable(design, outPin)) {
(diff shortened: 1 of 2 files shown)


test/src/com/xilinx/rapidwright/router/TestRouteThruHelper.java
```diff
@@ -0,0 +1,86 @@
...
+            "CLEL_R_X9Y235/CLEL_R.CLE_CLE_L_SITE_0_H6->>CLE_CLE_L_SITE_0_E_O,false",
+            // HMUX
+            "CLEL_R_X9Y235/CLEL_R.CLE_CLE_L_SITE_0_H1->>CLE_CLE_L_SITE_0_EMUX,false",
+            "CLEL_R_X9Y235/CLEL_R.CLE_CLE_L_SITE_0_H2->>CLE_CLE_L_SITE_0_EMUX,false",
+            "CLEL_R_X9Y235/CLEL_R.CLE_CLE_L_SITE_0_H3->>CLE_CLE_L_SITE_0_EMUX,false",
+            "CLEL_R_X9Y235/CLEL_R.CLE_CLE_L_SITE_0_H4->>CLE_CLE_L_SITE_0_EMUX,false",
+            "CLEL_R_X9Y235/CLEL_R.CLE_CLE_L_SITE_0_H5->>CLE_CLE_L_SITE_0_EMUX,false",
+            "CLEL_R_X9Y235/CLEL_R.CLE_CLE_L_SITE_0_H6->>CLE_CLE_L_SITE_0_EMUX,false",
+
+            // Occupied by DLUT6
```
eddieh-xlnx, line 67:
```suggestion
            // Occupied by D6LUT
```

test/src/com/xilinx/rapidwright/router/TestRouteThruHelper.java
```diff
@@ -0,0 +1,86 @@
...
+            "CLEL_R_X9Y235/CLEL_R.CLE_CLE_L_SITE_0_D1->>CLE_CLE_L_SITE_0_D_O,false",
+            "CLEL_R_X9Y235/CLEL_R.CLE_CLE_L_SITE_0_D6->>CLE_CLE_L_SITE_0_DMUX,false",
+
+            // Occupied by DLUT5
```
eddieh-xlnx, line 71:
```suggestion
            // Occupied by D5LUT
```

src/com/xilinx/rapidwright/router/RouteThruHelper.java
```diff
@@ -158,21 +158,57 @@ public static boolean isRouteThruPIPAvailable(Design design, PIP routethru) {
...
+                return false;
+            }
+        }
+        return true;
+    }
 
-        for (BELPin sink : belPin.getSiteConns()) {
-            Cell collision = siteInst.getCell(sink.getBEL());
-            if (collision != null) return false;
+    public static boolean isRouteThruPIPAvailable(Design design, Node start, Node end) {
```
clavin-xlnx, line 196: Would it make sense to unify this method with the one above, have a backend method like `public static boolean isRouteThruPIPAvailable(Design design, SitePin start, SitePin end)`?  The method bodies are very similar.

eddieh-xlnx: The reason I avoided doing this is because computing the second SitePin is expensive, and we'd like to avoid doing that if possible. The other option would be to take a `Tile`+`int` instead of a `Wire` or `Node` since converting either of the latter to the other would require allocation.

clavin-xlnx: Sure, that make sense.  Let's leave it alone for now.  The implementation of `Node.getSitePin()` would likely require another allocation to make that work.  Maybe just drop a comment that the method bodies should be kept in sync since they are doing the same thing.  Perhaps we could introduce an interface that would be compatible for both `Node` and `Wire` objects to avoid this in the future.

clavin-xlnx (approved):

PR #880, opened 2023-11-09
DesignTools.createMissingSitePinInsts(Design) to ignore GLOBAL_USEDNET

eddieh-xlnx:


clavin-xlnx (approved):
I agree, `SitePinInst` objects don't have any significance for `GLOBAL_USEDNET`.

PR #884, opened 2023-11-15
Cap RWRoute.presentCongestionFactor to prevent accuracy loss

eddieh-xlnx:
Specifically, it's been observed that when RWRoute operates on a hard design requiring many routing iterations to resolve congestion, its `presentCongestionFactor` can grow to such large values that overusing any node will inflate the path cost such that all other factors (distance to sink) cannot be captured using single-precision floating point causing the routing algorithm to fly blind. In particular, with `presentCongestionMultiplier` having a default value of 2, within 20 or so iterations, the [units-in-the-last-place](https://en.wikipedia.org/wiki/Unit_in_the_last_place) rises to 1.0 (`Math.ulp((float) Math.pow(2, 23)) == 1.0`).

Cap the `presentCongestionFactor` value using a heuristic that depends on the wirelength weight since RWRoute scales the distance to sink by this value:
https://github.com/Xilinx/RapidWright/blob/7bb791d319d9c863de0ce1666230fa5481d2e63c/src/com/xilinx/rapidwright/rwroute/RWRoute.java#L1634

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/RWRouteConfig.java
+++ b/src/com/xilinx/rapidwright/rwroute/RWRouteConfig.java
@@ -98,7 +98,7 @@ public class RWRouteConfig {
         enlargeBoundingBox = false;
         extensionYIncrement = 2;
         extensionXIncrement = 1;
-        wirelengthWeight = 0.8f;
+        setWirelengthWeight(0.8f);
         timingWeight = 0.35f;
         timingMultiplier = 1f;
         shareExponent = 2;
@@ -107,7 +107,7 @@ public class RWRouteConfig {
         reroutePercentage = (short) 3;
         initialPresentCongestionFactor = 0.5f;
         presentCongestionMultiplier = 2f;
-        setWirelengthWeight(1f);
+        historicalCongestionFactor = 1f;
         timingDriven = true;
         clkRouteTiming = null;
         pessimismA = 1.03f;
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/rwroute/RWRouteConfig.java
```diff
@@ -373,6 +375,16 @@ public void setWirelengthWeight(float wirelengthWeight) {
         if (wirelengthWeight < 0 || wirelengthWeight > 1)
             throw new IllegalArgumentException("ERROR: wirelength-driven weighting factor wirelengthWeight should be within [0, 1].");
         this.wirelengthWeight = wirelengthWeight;
+
+        // Assume that the minimum unit we want to observe is 1/8th of the wirelengthWeight
+        // (since during RWRoute.evaluateCostAndPush(), distanceToSink is scaled by wirelengthWeight)
+        // compute the largest floating-point value that results in this Units-in-the-Last-Place value.
+        final float maxUlp = wirelengthWeight / 8;
```
clavin-xlnx, line 382: How do we get 1/8 as the minimum unit?  Does this still make sense if someone chooses a different initial value for `wirelengthWeight` like `0.5`?

eddieh-xlnx: 1/8 is just a power-of-2 heuristic value. However, if someone chooses a wirelength weight of 0.5 (down from the default of 0.8), then `maxUlp` will be reduced accordingly, and the `maxPresentCongestionCost` will continue to be scaled down until its ULP is below this value.

clavin-xlnx (approved):

PR #887, opened 2023-11-17
Add convenience method for Net code gen, fix redundant code

clavin-xlnx:
Creates a method to generate RapidWright code to reproduce a single `Net`.  Also fixes some redundant generation of `SiteInst` in the `CodeGenerator`.


eddieh-xlnx (approved):
Interesting find! We mustn't have used this generator enough since the previous expected code would have caused an error if run.

PR #888, opened 2023-11-20
[RWRoute] Speedup: exclude PINFEEDs from routing graph and queue

eddieh-xlnx:
An underlying characteristic of the base RWRoute problem is that each connection only has a single exclusive sink node. Such sink nodes will typically have the `Node` intent code `IntentCode.NODE_PINFEED` which translates to the route node type `RouteNodeType.PINFEED_I`. Prior to the actual routing search, all connections will all have created their sink nodes in the routing graph.

Hence, two main optimizations can be applied:
1. In RWRoute, routing search expands the routing graph on-demand, creating routing nodes as they are explored for the first time. During this process, do not create any new routing nodes for `Node`s with `IntentCode.NODE_PINFEED`, since we know that all the pinfeeds that we need to route are already present in the graph. This prevents the router from exploring nodes corresponding to *unused* site pins.
2. During routing expansion, do not enqueue any `RouteNodeType.PINFEED_I` routing nodes unless it is the target route node. This prevents the router from exploring nodes corresponding to site pins *used for other connections*.

Note that both optimizations are only possible for base RWRoute which does not consider:
* LUT routethru-s (i.e. using the thru-site PIP `A[1-6]` to `A_O`)
* LUT pin swapping

**Added complication 1:** On multi-SLR devices, the `IntentCode.NODE_PINFEED` can also lead into a Laguna tile which is necessary to make a SLR crossing. Enumerate all such pinfeeds during routing graph construction such that they are not excluded by optimization 1.

**Added complication 2:** Some sink nodes (e.g. `[A-H](X|_I)`) will have the `IntentCode.NODE_PINBOUNCE` instead of `NODE_PINFEED` since they can be used to bounce onto other nodes. When such nodes are used as a sink node they will be assigned an `RouteNodeType.PINFEED_I`, whereas unused nodes will have the default route node type `RouteNodeType.PINBOUNCE`. As per optimization 2 above, such `RouteNodeType.PINFEED_I` will not be eligible to be enqueued which eliminates any pointless negotiation between two nets for this resource -- clearly the connection whose sink node it is must prevail. However, such nodes can support other connections on the *same* net -- thus, there is a condition in `RWRoute.isAccessiblePinfeedI()` that allows such `PINFEED_I` route nodes to be enqueued if it is already used by another connection on the same net when it has the `IntentCode.NODE_PINBOUNCE`.


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -1217,7 +1229,17 @@ private boolean shouldMergePath(Connection connection) {
      * @param connection The connection to be ripped up.
      */
     private void ripUp(Connection connection) {
-        for (RouteNode rnode : connection.getRnodes()) {
+        List<RouteNode> rnodes = connection.getRnodes();
+        if (rnodes.isEmpty()) {
+            assert(!connection.getSink().isRouted());
+            if (connection.getAltSinkRnode() == null) {
+                // If there is no alternate sink, decrement only this one-and-only sink node
```
eddieh-xlnx, line 1236:
```suggestion
                // If there is no alternate sink, decrement this one-and-only sink node
```

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -1228,7 +1250,8 @@ private void ripUp(Connection connection) {
      * @param connection The routed connection.
      */
     private void updateUsersAndPresentCongestionCost(Connection connection) {
-        for (RouteNode rnode : connection.getRnodes()) {
+        List<RouteNode> rnodes = connection.getRnodes();
+        for (RouteNode rnode : rnodes) {
```
eddieh-xlnx, line 1254:
```suggestion
        for (RouteNode rnode : connection.getRnodes()) {
```

eddieh-xlnx:
Some numbers below, with benchmarks from the [FPGA24 Routing Contest](https://xilinx.github.io/fpga24_routing_contest/):

Before:
```
+--------------------------------+------+------------------+
| Benchmark                      | Pass | Wall Clock (sec) |
+--------------------------------+------+------------------+
| boom_med_pb_rwroute            | True |           524.05 |
| vtr_mcml_rwroute               | True |           457.15 |
| rosetta_fd_rwroute             | True |           309.29 |
| corundum_25g_rwroute           | True |           528.60 |
| vtr_lu64peeng_rwroute          | True |           403.15 |
| corescore_500_rwroute          | True |           276.36 |
| corescore_500_pb_rwroute       | True |           533.32 |
| mlcad_d181_lefttwo3rds_rwroute | True |          4556.23 |
| koios_dla_like_large_rwroute   | True |           686.09 |
| boom_soc_rwroute               | True |          2272.74 |
| ispd16_example2_rwroute        | True |           979.13 |
+--------------------------------+------+------------------+
```
Total: 11526 seconds

After:
```
+--------------------------------+------+------------------+--------------------------+---------+
| Benchmark                      | Pass | Wall Clock (sec) | Critical-Path Wirelength | Score   |
+--------------------------------+------+------------------+--------------------------+---------+
| boom_med_pb_rwroute            | True |           354.01 |                     1014 |  420.01 |
| vtr_mcml_rwroute               | True |           378.93 |                      646 |  405.64 |
| rosetta_fd_rwroute             | True |           300.01 |                      868 |  356.81 |
| corundum_25g_rwroute           | True |           386.90 |                      391 |  387.31 |
| vtr_lu64peeng_rwroute          | True |           319.43 |                     1522 |  439.69 |
| corescore_500_rwroute          | True |           220.24 |                      653 |  263.52 |
| corescore_500_pb_rwroute       | True |           391.60 |                      694 |  421.84 |
| mlcad_d181_lefttwo3rds_rwroute | True |          2994.11 |                      824 | 2777.10 |
| koios_dla_like_large_rwroute   | True |           546.37 |                      566 |  548.33 |
| boom_soc_rwroute               | True |          1841.42 |                     1692 | 1826.48 |
| ispd16_example2_rwroute        | True |           789.49 |                     1113 |  821.84 |
+--------------------------------+------+------------------+--------------------------+---------+
```
Total: 8522 seconds (26% less)

clavin-xlnx (approved):
Awesome work!

PR #889, opened 2023-11-20
Fix RouteThruHelper.isRouteThruPIPAvailable() for SLICEM.H[1-6]

eddieh-xlnx:
The site pin `H[1-6]` on `SLICEM`s are both the read address for the `H[56]LUT` as well as the write address `[A-H][56]LUT` distributed RAMs.

Prior to this PR, querying RouteThruHelper for whether a `H[1-6]` -> `H_O` routethru is possible would fail if there was any other LUT cell (`[A-G][56]LUT`) in the same site.

### DIFF
--- a/src/com/xilinx/rapidwright/router/RouteThruHelper.java
+++ b/src/com/xilinx/rapidwright/router/RouteThruHelper.java
@@ -35,6 +35,7 @@ import com.xilinx.rapidwright.design.Cell;
 import com.xilinx.rapidwright.design.Design;
 import com.xilinx.rapidwright.design.Net;
 import com.xilinx.rapidwright.design.SiteInst;
+import com.xilinx.rapidwright.device.BEL;
 import com.xilinx.rapidwright.device.BELPin;
 import com.xilinx.rapidwright.device.Device;
 import com.xilinx.rapidwright.device.Node;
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/router/RouteThruHelper.java
```diff
@@ -194,11 +195,19 @@ public static boolean isRouteThruPIPAvailable(Design design, Wire start, Wire en
...
-                Cell cellCollision = siteInst.getCell(sink.getBEL());
+                BEL sinkBEL = sink.getBEL();
+                if (sinkBEL.getName().charAt(0) != inPin.getPinName().charAt(0)) {
+                    continue;
+                }
+                Cell cellCollision = siteInst.getCell(sinkBEL);
                 if (cellCollision != null) {
+                    // Ignore BELs that don't share the same LUT letter
+                    // Specifically, this is to prevent H[1-6] inputs on SLICEM sites
+                    // -- which also drive [A-G].WA[1-6] -- from considering [A-G]LUT[56]
```
clavin-xlnx, line 210: Not sure if it is important, but this comment does not appear in the same location in the code as the identical method below.

clavin-xlnx (approved):

src/com/xilinx/rapidwright/router/RouteThruHelper.java
eddieh-xlnx: Ooh yes. This needs moving!!

eddieh-xlnx, line 210:
```suggestion
                    // Ignore BELs that don't share the same LUT letter
                    // Specifically, this is to prevent H[1-6] inputs on SLICEM sites
                    // -- which also drive [A-G].WA[1-6] -- from considering [A-G]LUT[56]
                    continue;
                }
                Cell cellCollision = siteInst.getCell(sinkBEL);
                if (cellCollision != null) {
```

PR #892, opened 2023-11-21
Create a common interface for Node and Wire Objects

clavin-xlnx:
This update will take a few steps, but getting the interface in place first has no dependencies.  

In a future release, `Node` and `Wire` objects would implement this interface.

At a minimum, it would allow us to collapse `RouteThruHelper.isRouteThruPIPAvailable(Design, Node, Node)` and `RouteThruHelper.isRouteThruPIPAvailable(Design, Wire, Wire)`.

### DIFF
--- a/src/com/xilinx/rapidwright/device/WireObject.java
+++ b/src/com/xilinx/rapidwright/device/WireObject.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2023, Advanced Micro Devices, Inc.
+ * All rights reserved.
+ *
+ * Author: Chris Lavin, AMD Research and Advanced Development.
+ *
+ * This file is part of RapidWright.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


clavin-xlnx:
The `RouteNode` object could be another candidate object to implement the interface.

src/com/xilinx/rapidwright/device/WireObject.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ */
+package com.xilinx.rapidwright.device;
+
+/**
+ * An interface to enable methods to operate on the common methods of a Wire
+ * object as well as a Node object. At their heart, they both specify a Tile and
+ * a wire (in the case of a Node it is the base wire).
+ */
+public interface WireObject {
```
eddieh-xlnx, line 29: How about `WireInterface` or `WireBase` or `BaseWire`?

clavin-xlnx: Good names, I'll go with `WireInterface`.

eddieh-xlnx (approved):

PR #899, opened 2023-11-28
Add TestSATRouter.testApplyResult() 

eddieh-xlnx:
which -- in the interests of runtime and predictability -- uses a pre-canned solution just to test the `SATRouter.applyRoutingResult()` method.


test/src/com/xilinx/rapidwright/router/TestSATRouter/reduce_or_routed_7overlaps_solution.txt
```diff
@@ -0,0 +1,3491 @@
+{
```
clavin-xlnx, line 1: Any particular reason to keep this in a the `src` folder?  It seems like `test/RapidWrightDCP` would be more appropriate and has other `.txt` files in that location as well.

clavin-xlnx (approved):

PR #910, opened 2023-12-06
RouterHelper.invertPossibleGndPinsToVccPins() to invert static LUT inputs

eddieh-xlnx:
In the UltraScale/UltraScale+ architectures (if not more) all LUT inputs can be connected to VCC directly without occupying any other PIPs in the switch box. However, the same cannot be said of LUT inputs connected to GND -- these do require more general routing PIPs to service (and thus increases routing pressure). 

However, the nature of LUTs means that inputs tied to GND can be transformed into inputs tied to VCC by absorbing a new inverter by modifying the LUT equation/mask to reflect this.

Of course, it would be even better if LUTs with static inputs can be shrunk into smaller LUTs with no static inputs, but this would not improve routing congestion.

~There is one catch with the current implementation of this PR: it needs a uniquified/unfolded logical netlist -- i.e. where every modified LUT must correspond to a unique `EDIFCellInst` in the logical netlist.~
Enhancement made in #918 allows determination of uniquification state.

Take for example:
```
module top;
sub s0(1'b0);
sub s1(1'b1);
endmodule

module sub(input i);
LUT1 foo(i);
endmodule
```
Modifying the LUT equation on `s0/foo` (the instance `sub.foo`) will have the undesired effect of also modifying `s1/foo`'s equation too. Instead, you would need a uniquified/unfolded design like so:
```
module top;
sub0 s0(1'b0);
sub1 s1(1'b1);
endmodule

module sub0(input i);
LUT1 foo(i);
endmodule

module sub1(i);
LUT1 foo(i);
endmodule
```

Now `EDIFTools.uniqueifyNetlist()` will do this for all cell instances in the design. However, this could be a little too heavy handed if all we want to do is to modify a handful of LUTs (perhaps we need to uniqueify on demand?)

~In addition, there does not seem to be a way to easily check that a netlist/cell-instance is uniquified before we apply this optimization.~ (#918)

This PR enables this optimization by default for `RouterHelper.invertPossibleGndPinsToVccPins()` -- which will only invert a LUT's GND pin into a VCC pin only if the LUT cell is uniquified, By extension this optimization is also enabled by default for RWRoute. RWRoute can use the new `--noInvertGndToVccForLutInputs` option to disable this.


src/com/xilinx/rapidwright/rwroute/RouterHelper.java
```diff
@@ -371,28 +373,56 @@ private static boolean isInvertibleDSPBELPin(BELPin belPin) {
...
+                    // TODO: This may modify an folded EDIFCellInst!!!
+                    LUTTools.configureLUT(cell, newLutEquation);
+                }
+            } else {
+                BELPin[] belPins = currSitePinInst.getSiteInst().getSiteWirePins(currSitePinInst.getName());
+                // DSP or BRAM
+                if (belPins.length == 2) {
+                    for (BELPin belPin : belPins) {
+                        if (belPin.isSitePort()) continue;
+                        if (currSitePinInst.getSite().getName().startsWith("RAM")) {
```
clavin-xlnx, line 411: URAMs also can have invertible inputs:
image

src/com/xilinx/rapidwright/rwroute/RouterHelper.java
```diff
@@ -371,28 +373,56 @@ private static boolean isInvertibleDSPBELPin(BELPin belPin) {
...
+                    String logicalPinName = cell.getLogicalPinMapping(physicalPinName);
+
+                    // Get the LUT equation
+                    String lutEquation = LUTTools.getLUTEquation(cell);
+                    assert(lutEquation.contains(logicalPinName));
+
+                    // Compute a new LUT equation with that logical input inverted
+                    String newLutEquation = lutEquation.replace(logicalPinName, "!" + logicalPinName)
+                            // Cancel out double inversions
+                            .replace("!!", "");
```
clavin-xlnx, line 401: Although not common, Vivado does allow for `~` as an alternative to `!` to specify inversion:
image

src/com/xilinx/rapidwright/rwroute/RouterHelper.java
eddieh-xlnx: You're absolutely right. This PR just indents what we had before here, but I'll add a note that those BELs (plus others) are other things that can be identified in the future.

src/com/xilinx/rapidwright/rwroute/RouterHelper.java
eddieh-xlnx: My understanding is that the full 2^K LUT mask is stored as the `LUT_INIT` property. This `LUTTools.getLUTEquation()` parses this `LUT_INIT` property into a LUT equation, and it only uses `!` instead of `~` so I think what exists should be okay. Added a comment.

clavin-xlnx: You're right, depending on `getLUTEquation()` should prevent any `~` from showing up.

eddieh-xlnx:
> > Now `EDIFTools.uniqueifyNetlist()` will do this for all cell instances in the design. However, this could be a little too heavy handed if all we want to do is to modify a handful of LUTs (perhaps we need to uniqueify on demand?)
> 
> Do we have a sense for the runtime/memory overhead of running this pre-emptively? It would be good to understand the tradeoff of doing a lazy-based uniqueify vs. whole-design uniqueify. I'm not opposed to it, but I think it is some engineering work to do it and we should justify that it will improve memory and/or runtime.

How pre-emptively are we talking? Should we do this on loading every design, or should we require that this uniqueify-all step be done already if this optimization is to be considered?

Besides runtime, there are also memory and file size considerations when uniqueifying all. Once uniqueified, it becomes much more difficult to re-uniqueify again.

> > In addition, there does not seem to be a way to easily check that a netlist/cell-instance is uniquified before we apply this optimization.
> 
> To add this notion, it would be an actively managed flag on instances that would have to be created on load and maintained with all cell creations/removals. I'm sure we could afford an extra boolean memory wise, it would just be the engineering time to implement.

I agree: I think a boolean (or even an `int`) on an extra `EDIFCellInst` is manageable, and it shouldn't incur significant runtime overhead to maintain that.

eddieh-xlnx:
Here's some data. For two huge designs (which are already uniquified), starting with `11`:
```
==============================================================================
==                                                                          ==
==============================================================================
 XML Parse & Device Load:     2.113s    557.151MBs          |   1203.770MBs (peak)
    (Read PhysDB Caches):                      (    6.906s) |  10751.488MBs (peak)
            (EDIF Parse):                      (   24.691s) |  21768.000MBs (peak)
 (Read PhysDB Placement):                      (   97.220s) |  32588.055MBs (peak)
   (Read PhysDB Routing):                      (  104.487s) |  32763.027MBs (peak)
(Read PhysDB Finalizing):                      (   19.987s) |  33726.215MBs (peak)
   Read Netlist & PhysDB:   131.397s  18835.701MBs          |  34018.586MBs (peak)
                 *Total*:   133.509s  19392.852MBs          |  34019.020MBs (peak)
      uniqueifyNetlist():     1.768s      0.001MBs          |  34127.797MBs (peak)
```

And `08`:
```
==============================================================================
==                                                                          ==
==============================================================================
 XML Parse & Device Load:     2.112s    557.151MBs          |   1201.730MBs (peak)
    (Read PhysDB Caches):                      (    4.747s) |   4110.949MBs (peak)
            (EDIF Parse):                      (   53.711s) |  31390.668MBs (peak)
   (Read PhysDB Routing):                      (  107.464s) |  33261.848MBs (peak)
 (Read PhysDB Placement):                      (  129.162s) |  34076.863MBs (peak)
(Read PhysDB Finalizing):                      (   11.799s) |  34326.121MBs (peak)
   Read Netlist & PhysDB:   145.795s  20366.238MBs          |  34669.797MBs (peak)
                 *Total*:   147.907s  20923.390MBs          |  34669.797MBs (peak)
      uniqueifyNetlist():     1.725s      0.001MBs          |  34721.859MBs (peak)
```

Performing the analysis on the netlist (which we know to be unique, and thus no uniquification needs to happen) takes less than two seconds even on our biggest designs which is less than 2% of the time taken to read the design.

Looks like you're right -- this is a storm in a teacup. We still need to address the "don't perform this optimization if we're not sure it is unique" which does mean we have to do this post-analysis (or keep some state up front when reading the netlist). Let's think about this some more.

test/src/com/xilinx/rapidwright/rwroute/TestRouterHelper.java
```diff
@@ -172,4 +177,30 @@ public static Stream<Arguments> testInvertPossibleGndPinsToVccPins() {
...
+
+        // Must have moved onto vccNet, and the LUT mask inverted
+        Net vccNet = design.getVccNet();
+        Assertions.assertEquals("[IN SLICE_X0Y0.A6]", vccNet.getPins().toString());
+        Assertions.assertEquals("O=I0", LUTTools.getLUTEquation(cell));
+    }
+
+    @Test
+    public void testInvertPossibleGndPinsToVccPinsLutInputNotUniquified() {
+        // TODO
```
eddieh-xlnx, line 204: TODO

eddieh-xlnx:
> > To add this notion, it would be an actively managed flag on instances that would have to be created on load and maintained with all cell creations/removals. I'm sure we could afford an extra boolean memory wise, it would just be the engineering time to implement.
> 
> I agree: I think a boolean (or even an `int`) on an extra `EDIFCellInst` is manageable, and it shouldn't incur significant runtime overhead to maintain that.

Implemented in #918, which this PR now depends on.

test/src/com/xilinx/rapidwright/edif/TestEDIFCellInst.java
```diff
@@ -0,0 +1,87 @@
...
+        EDIFCell picoblazeTop = netlist.getCell("picoblaze_top");
+        Assertions.assertFalse(picoblazeTop.isUniquified());
+        for (String name : Arrays.asList(
+                "processor",
+                "your_program"
+        )) {
+            EDIFCellInst eci = picoblazeTop.getCellInst(name);
+            // Only checks that this cell instance is unique (e.g. that there is only
+            // one instantiation) but does not check that any parents on a full
+            // hierarchical path (e.g. "picoblaze_{0,1}_{12,13}/processor") is also
```
eddieh-xlnx, line 82:
```suggestion
            // one instantiation) but does not check that all parents on a full
            // hierarchical path (e.g. "picoblaze_{0,1}_{12,13}/processor") are also
```

eddieh-xlnx:
@clavin-xlnx I merged in your changes from the other branch and made one additional change: https://github.com/Xilinx/RapidWright/pull/910/commits/6a681fb99343ca355b23d2d0845ca4e8d053ad76. Please give it a once over!

clavin-xlnx:
> @clavin-xlnx I merged in your changes from the other branch and made one additional change: [6a681fb](https://github.com/Xilinx/RapidWright/commit/6a681fb99343ca355b23d2d0845ca4e8d053ad76). Please give it a once over!

Looks good to me.

clavin-xlnx (approved):

PR #914, opened 2023-12-08
[RWRoute] GlobalSignalRouting static net router to use [A-H]MUX outputs

eddieh-xlnx:
... on top of `[A-H]_O` to supply static sources.

### DIFF
--- a/src/com/xilinx/rapidwright/interchange/PhysNetlistWriter.java
+++ b/src/com/xilinx/rapidwright/interchange/PhysNetlistWriter.java
@@ -414,7 +414,7 @@ public class PhysNetlistWriter {
                                 continue;
                             }
                         } else {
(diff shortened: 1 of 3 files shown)


clavin-xlnx (approved):
Not sure when, but at some point we'll need to start differentiating based on `Series` as Versal will need to be supported as a target device.  Perhaps we should consider making changes with this in mind?

eddieh-xlnx:
> Not sure when, but at some point we'll need to start differentiating based on `Series` as Versal will need to be supported as a target device. Perhaps we should consider making changes with this in mind?

Fair point! I think we need a big overhaul when we come to supporting Versal, which requires a lot of things to come together: DCPs we can use as tests, our own familiarity with the Versal routing, improvements to the core RapidWright APIs to support more than two output sources, etc.

PR #918, opened 2023-12-11
Add EDIFHierCellInst.isUniquified()

eddieh-xlnx:
Add the ability to determine if an `EDIFHierCellInst` is unique across the netlist, with a view to enabling algorithms (such as #910) to decide whether it is safe to modify.

To answer the `EDIFHierCellInst.isUniquified()` question, we need to know whether all `EDIFCellInst`-s that make up `EDIFHierCellInst` are themselves individually unique (folded, non-hierarchical) instantiations across the netlist.

To answer this `EDIFCellInst.isUniquified()` question, we need to track how many `EDIFCellInst`s exist for each `EDIFCell` which is also implemented in this PR.

Take for example the hierarchical path:
```
top/foo/bar/baz
```
To determine if it's safe to modify the `baz` instantiation (i.e. to change its parameters/LUT masks, or to change its port connections) we must determine that its parent `bar` instantiation is unique, followed by its parent `foo` instantiation being unique (and assume that `top` must be unique).

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFCell.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFCell.java
@@ -168,10 +168,6 @@ public class EDIFCell extends EDIFPropertyObject {
                     getName() + ", trying to add instance " + instance.getName() +
                     " which already exists inside this cell.");
         }
-        EDIFCell cellType = instance.getCellType();
-        if (cellType != null) {
-            cellType.incrementInstanceCount();
-        }
         return instance;
     }
 
@@ -189,10 +185,6 @@ public class EDIFCell extends EDIFPropertyObject {
             instance.setName(instance.getName() + "_" + getLibrary().getNetlist().nameSpaceUniqueCount++);
         }
         instances.put(instance.getName(), instance);
-        EDIFCell cellType = instance.getCellType();
-        if (cellType != null) {
-            cellType.incrementInstanceCount();
-        }
         return instance;
     }
 
--- a/src/com/xilinx/rapidwright/edif/EDIFCell.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFCell.java
@@ -690,6 +690,13 @@ public class EDIFCell extends EDIFPropertyObject {
      */
     public void decrementInstanceCount() {
         instanceCountUpdater.getAndDecrement(this);
+        if (getInstanceCount() == 0) {
+            // Since this cell is no longer instantiated, decrement the count of all the cells
+            // that gets instantiated within
+            for (EDIFCellInst instance : getCellInsts()) {
+                instance.getCellType().decrementInstanceCount();
+            }
+        }
     }
 
     /**
--- a/src/com/xilinx/rapidwright/edif/EDIFCell.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFCell.java
@@ -678,7 +678,7 @@ public class EDIFCell extends EDIFPropertyObject {
     }
 
     /**
-     * Atomically decrement instance count of this cell.
+     * @return The number of times this cell has been instantiated.
      */
     public int getInstanceCount() {
         return instanceCountUpdater.get(this);
--- a/src/com/xilinx/rapidwright/edif/EDIFCell.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFCell.java
@@ -675,16 +675,6 @@ public class EDIFCell extends EDIFPropertyObject {
      */
     public void incrementInstanceCount() {
         instanceCountUpdater.incrementAndGet(this);
-
-        if (getInstanceCount() == 1) {
-            // This is the first instantiation of its cell -- increment the count all the cells
-            // that are instantiated within
-
-            // TODO: How to do this without conflicting with the initial parsers?
-
-            // TODO: How can we determine that the added instance is actually
-            //       connected to the top cell?
-        }
     }
 
     /**
@@ -692,16 +682,6 @@ public class EDIFCell extends EDIFPropertyObject {
      */
     public void decrementInstanceCount() {
         instanceCountUpdater.getAndDecrement(this);
-        if (getInstanceCount() == 0) {
-            // Since this cell is no longer instantiated, decrement the count of all the cells
-            // that gets instantiated within
-            for (EDIFCellInst instance : getCellInsts()) {
-                instance.getCellType().decrementInstanceCount();
-            }
-
-            // TODO: How can we determine that the removed instance was actually
-            // connected to the top cell to begin with?
-        }
     }
 
     /**
--- a/src/com/xilinx/rapidwright/edif/EDIFCell.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFCell.java
@@ -66,10 +66,12 @@ public class EDIFCell extends EDIFPropertyObject {
     /**
      * An atomically updated variable to track the number of `EDIFCellInst`
      * objects (attached to a parent cell) that instantiate this cell.
+     * Note: this does not track the number of `EDIFHierCellInst` objects
+     *       that would exist if the netlist was traversed from the top cell.
      */
-    private volatile int instanceCount = 0;
-    private static final AtomicIntegerFieldUpdater<EDIFCell> instanceCountUpdater =
-            AtomicIntegerFieldUpdater.newUpdater(EDIFCell.class, "instanceCount");
+    private volatile int nonHierInstantiationCount = 0;
+    private static final AtomicIntegerFieldUpdater<EDIFCell> nonHierInstantiationCountUpdater =
+            AtomicIntegerFieldUpdater.newUpdater(EDIFCell.class, "nonHierInstantiationCount");
 
     public EDIFCell(EDIFLibrary lib, String name) {
         super(name);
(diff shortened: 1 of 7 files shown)


src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -656,5 +662,33 @@ public boolean equals(Object o) {
...
+
+    /**
+     * Atomically decrement instance count of this cell.
+     */
+    public void decrementInstanceCount() {
+        instanceCountUpdater.getAndDecrement(this);
+    }
+
+    /**
+     * Atomically decrement instance count of this cell.
```
eddieh-xlnx, line 681:
```suggestion
     * @return The number of times this cell has been instantiated.
```

test/src/com/xilinx/rapidwright/edif/TestEDIFCell.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Test
+    public void testIsUniquifiedFalse() {
+        Design design = RapidWrightDCP.loadDCP("picoblaze4_ooc_X6Y60_X6Y65_X10Y60_X10Y65.dcp");
+        EDIFNetlist netlist = design.getNetlist();
+        Assertions.assertTrue(netlist.getTopCell().isUniquified());
+        EDIFCell picoblazeTop = netlist.getCell("picoblaze_top");
+        Assertions.assertFalse(picoblazeTop.isUniquified());
+        Assertions.assertEquals(4, picoblazeTop.getInstanceCount());
+
```
eddieh-xlnx, line 61:
```suggestion
```

clavin-xlnx:
I think there are still some considerations that need to be reasoned about in order to make this feature a reality that we can depend on.  

1. We will need someway to enforce copy constructors on primitive and macro cells when they are loaded/instantiated in a design so we don't end up changing the values inside a static library of cells.  Obviously, the primitives and macros don't need 'unique-ification' but modifying a cell that doesn't belong to the current design would be bad practice and if two or more designs are loaded at the same time, the instance count would be invalid.  
2. [`public EDIFCellInst removeCellInst(String name)` in `EDIFCell`](https://github.com/Xilinx/RapidWright/blob/f4241d89e65f1ed981f005d175fd41de873a7a6c/src/com/xilinx/rapidwright/edif/EDIFCell.java#L293) would need to be updated accordingly anytime an instance is removed.
3. We may need to think about black boxes and make sure they are handled appropriately without breaking some assumptions of this approach.

eddieh-xlnx:
Sorry, I just cherry-picked some updates I forgot to move back into this branch.

Just to make sure we're on the same page: the purpose of this PR is to allow the determination of whether a `EDIFHierCellInst` is the one-and-only unique instantiation of an `EDIFCell`. The way we do this is to count how many times an `EDIFCellInst`-s instantiates a particular `EDIFCell`, and for `EDIFHierCellInst` to check all instances in its path are unique in this manner.

It does not make sense to call this on an `EDIF{Hier,}CellInst` of a primitive (in fact, we could explicitly block this with an assertion, exception, or false). I cannot see a reason to expect anyone to modify any part of the `EDIFCell` interface or contents of such primitives. You would, however, expect to modify the properties of the `EDIFCellInst` -- and for that, you would want to check that the parent(s) of such a cell instance to be fully uniquified.

Take for example:
```
module foo;
LUT1 #(.INIT(...)) bar();
endmodule
```
It does not make sense to query if `LUT1` is unique, but it does make sense to see if `foo` is (hierarchically) unique in order to determine if the `INIT` parameter can be modified.

In response to your points:
> 1. We will need someway to enforce copy constructors on primitive and macro cells when they are loaded/instantiated in a design so we don't end up changing the values inside a static library of cells. Obviously, the primitives and macros don't need 'unique-ification' but modifying a cell that doesn't belong to the current design would be bad practice and if two or more designs are loaded at the same time, the instance count would be invalid.

Yes, that is a valid point, but one I think should be considered out of the scope of this PR. I believe we have a broader issue with primitives from the static/singleton library not being copied into each design, in which case (and when there are multiple netlists present) then the unique-ness or instance count values are meaningless. Perhaps the best way would be for primitives to error out when asked this question (which is of questionable value to begin with)?

> 2. [public EDIFCellInst removeCellInst(String name) in EDIFCell](https://github.com/Xilinx/RapidWright/blob/f4241d89e65f1ed981f005d175fd41de873a7a6c/src/com/xilinx/rapidwright/edif/EDIFCell.java#L293) would need to be updated accordingly anytime an instance is removed.

This is an excellent point. Currently, I'm hooking into `EDIFCellInst.setCellType()` to increment/decrement the instance count. However, that doesn't capture whether (a) that `EDIFCellInst` is actually attached to an `EDIFCell` (or indeed, if that `EDIFCell` is eventually attached to the top cell of a particular netlist) nor (b) the instance count is not decremented when this cell gets garbage collected, for whatever (other) reason.
This needs more thought, and `EDIFCell.{add,remove}CellInst()` could be a solution here.

> 3. We may need to think about black boxes and make sure they are handled appropriately without breaking some assumptions of this approach.

Can you elaborate on your concerns with blackboxes please? Blackbox-es are `EDIFCell`-s just like white-boxes but have a hollow interior. They can still be instantiated as normal, in which case there can still be count of how many times it gets instantiated, and this state of unique-ness can still be used to determine whether an `EDIFCellInst` can be modified. Nor does it make sense to query the uniqueness `EDIFCellInst` present inside a blackbox, since blackboxes cannot have anything inside.

eddieh-xlnx:
> > 3\. [public EDIFCellInst removeCellInst(String name) in EDIFCell](https://github.com/Xilinx/RapidWright/blob/f4241d89e65f1ed981f005d175fd41de873a7a6c/src/com/xilinx/rapidwright/edif/EDIFCell.java#L293) would need to be updated accordingly anytime an instance is removed.
> 
> This is an excellent point. Currently, I'm hooking into `EDIFCellInst.setCellType()` to increment/decrement the instance count. However, that doesn't capture whether (a) that `EDIFCellInst` is actually attached to an `EDIFCell` (or indeed, if that `EDIFCell` is eventually attached to the top cell of a particular netlist) nor (b) the instance count is not decremented when this cell gets garbage collected, for whatever (other) reason. This needs more thought, and `EDIFCell.{add,remove}CellInst()` could be a solution here.

It looks like handling removals is not so straightforward. In the common case (loading a "minimal" EDIF into memory, where all cells present are instantiated at least once) then each `EDIFCell.addCellInst()` can be assumed to increment the instance count of the `EDIFCell` it instantiates.

However, when we start considering `EDIFCell.removeCellInst()` then things get tricky. Detaching an `EDIFCellInst` from its parent cell would mean that the `EDIFCell` being instantiated should have its instance count decremented. But what about the cell type of *its* instances, and so on, which may also be detached now?

And then how can we elegantly handle `EDIFCell.addCellInst()` for things that were previously detached -- should the whole transitive tree of cells being instantiated also have their instance count bumped? Can double-incrementing be efficiently avoided when the common case uses the same routines?

eddieh-xlnx:
Okay, I think I finally have something I think could work.

`EDIFCellInst`-s will now increment the instance count of the `EDIFCell` it instantiates only if:
1. It has a `parentCell` (the cell it belongs to)
2. It has has a `cellType` (the cell it instantiates)

Attaching to both of those causes the instance count to increment, and detaching either causes it to decrement.

`EDIFCell.removeCellInst()` now sets `EDIFCellInst.parentCell` to be null (indicating detachment) which in turn decrements the instance count of `EDIFCell.cellType`. Unlike an early implementation, we do not recursively traverse into `EDIFCell.cellType` to decrement downstream instances.

`EDIFCell.addCellInst()` does the opposite, and because we check for `parentCell` and `cellType` then I don't believe double incrementing occurs.

src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -62,6 +63,14 @@ public class EDIFCell extends EDIFPropertyObject {
 
     private EDIFName view = DEFAULT_VIEW;
 
+    /**
+     * An atomically updated variable to track the number of `EDIFCellInst`
+     * objects (attached to a parent cell) that instantiate this cell.
+     */
+    private volatile int instanceCount = 0;
```
eddieh-xlnx, line 70: `directInstanceCount`?
`explicitInstanceCount`?
`foldedInstanceCount`?

src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -292,8 +301,11 @@ public EDIFCellInst removeCellInst(EDIFCellInst cellInstance) {
 
     public EDIFCellInst removeCellInst(String name) {
         if (instances == null) return null;
-        trackChange(EDIFChangeType.CELL_INST_REMOVE, name);
-        return instances.remove(name);
+        EDIFCellInst removedInstance = instances.remove(name);
+        if (removedInstance != null) {
+            removedInstance.setParentCell(null);
```
eddieh-xlnx, line 306:
```suggestion
            assert(removedInstance.getParentCell() == this);
            removedInstance.setParentCell(null);
```

src/com/xilinx/rapidwright/edif/EDIFCell.java
eddieh-xlnx: I've gone with `nonHierInstantiationCount` to make it clear that it's most definitely not `Hier`.

PR #917, opened 2023-12-09
[ECOTools] Inline cell insertion

clavin-xlnx:
ECO to insert a cell inline based on a specific input pin.  Method signature:
```
 public static Cell createInlineCellOnInputPin(Design design, EDIFHierPortInst input,
            Unisim reference, Site site, BEL bel, String logInput, String logOutput)
```

For example:

image

The magenta colored LUT1 was inserted along with the cyan colored net to connect it to the existing input pin.  

Also added a few convenience methods.

PR #919, opened 2023-12-12
[DesignTools.makeBlackBox()] Fixes routing issues in makeBlackBox()

clavin-xlnx:
When creating a shell DCP as in the [Reuse Timing-closed Logic As A Shell](https://www.rapidwright.io/docs/ReusingTimingClosedLogicAsAShell.html) tutorial, sometimes certain designs will experience unroutable situations (or antennas) resulting from certain Vivado behaviors and the implementation state of the shell DCP.  This PR fixes at least some of those issues:
1. Previously, when `Net` objects needed to be renamed, some of the `SitePIP` objects were removed inadvertently.  This resolves this issue by using `Net.rename()` rather than `DesignTools.updateNetName()` which does not preserve the site pip routing.
2. There is a rare scenario when a black box is created and one of its outputs drives a flip-flop in a SLICE and the default site input bypass pin `[A-H][X|_I]` is being consumed by a fixed bounce node part of unrelated route.  In the previous context, it was likely driven by a LUT and thus the bypass input pin was not needed.  However when converting this connection input to be used as a shell, it is uncertain that a driver could be placed in the LUT site.  To rectify this issue and still preserve the existing routing in the shell, a `LUT1` is inserted into the netlist to provide a 'user-routethru' (Vivado does not handle a regular routethru inserted into the site).  This allows routing on the bypass pin to be preserved and also allow routing to the input of the flip flop.  This PR uses functionality introduced in #917 to achieve this behavior.
3. After black box creation, some BEL sites, specifically FF sites can be unusable because all their routing exit mux nodes are consumed.  This is somewhat rare, but the Vivado placer does not identify these scenarios and the BEL sites require a `PROHIBIT` in order for the Vivado placer to avoid placing flip flops into those locations.  This update detects those unused flip flop BELs that are unroutable and adds a `PROHIBIT` to them.

This PR also modifies `ECOTools.connectNet()` to allow for a connection to be driven by a black box.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -68,6 +68,7 @@ import com.xilinx.rapidwright.device.SitePin;
 import com.xilinx.rapidwright.device.SiteTypeEnum;
 import com.xilinx.rapidwright.device.Tile;
 import com.xilinx.rapidwright.device.Wire;
+import com.xilinx.rapidwright.eco.ECOTools;
 import com.xilinx.rapidwright.edif.EDIFCell;
 import com.xilinx.rapidwright.edif.EDIFCellInst;
 import com.xilinx.rapidwright.edif.EDIFHierCellInst;
@@ -1817,10 +1818,10 @@ public class DesignTools {
         Set<String> netsToKeep = new HashSet<>();
         for (Entry<Net, String> e : netsToUpdate.entrySet()) {
             EDIFHierNet newSource = d.getNetlist().getHierNetFromName(e.getValue());
-            Net updatedNet = DesignTools.updateNetName(d, e.getKey(), newSource.getNet(), e.getValue());
-            if (updatedNet != null) {
-                netsToKeep.add(updatedNet.getName());
+            if (!e.getKey().rename(e.getValue())) {
+                throw new RuntimeException("ERROR: Failed to rename net '" + e.getKey().getName() + "'");
             }
+            netsToKeep.add(e.getKey().getName());
         }
 
         t.stop().start("cleanup siteinsts");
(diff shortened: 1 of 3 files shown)


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3926,8 +3938,45 @@ public static void prohibitPartialHalfSlices(Design design) {
                         bottomUsed = true;
                     }
                 }
+                if (c.getBEL().isFF()) {
+                    String belName = c.getBELName();
+                    char letter = belName.charAt(0);
+                    if (si.getCell(letter + "6LUT") == null && si.getCell(letter + "5LUT") == null) {
```
eddieh-xlnx, line 3944: I'm guessing there is a very slim (but non zero) chance that this condition might not be true and we miss the opportunity to insert a RT. 

It would be great to get a warning (even under some verbose setting) emitted that a RT won't be considered here just in case we have to shoot around in the dark. It seems like this could be achieved by deferring this check until later -- first you check whether the node is used, and the existence of a sitewire assignment (does `makeBlackbox()` remove site wires assignments?) or SPI, before attempting to insert a RT.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3926,8 +3938,45 @@ public static void prohibitPartialHalfSlices(Design design) {
                         bottomUsed = true;
                     }
                 }
+                if (c.getBEL().isFF()) {
```
eddieh-xlnx, line 3941: It seems like this new code doesn't fit the ethos of the original method name (prohibiting half slices) -- does it make sense to move it out to another standalone method that could be called without prohibiting anything?

At the very least, let's update the javadoc.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3936,12 +3985,51 @@ public static void prohibitPartialHalfSlices(Design design) {
                             bels.add(si.getSiteName() + "/" + bel.getName());
```
eddieh-xlnx, line 3985:
```suggestion
                            bels.add(si.getSiteName() + "/" + bel.getName());
                            continue;
```

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx: Refactored the name and updated Javadoc.

eddieh-xlnx (approved):

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3926,22 +3940,102 @@ public static void prohibitPartialHalfSlices(Design design) {
...
+                                SitePinInst spi = si.getSitePinInst(sitePinName);
+                                if (spi == null || (spi != null && !spi.getNet().equals(net))) {
+                                    BELPin lutInput = si.getBEL(letter + "6LUT").getPin("A6");
+                                    EDIFHierPortInst ffInput = c.getEDIFHierCellInst().getPortInst("D");
+                                    ECOTools.createAndPlaceInlineCellOnInputPin(design, ffInput, Unisim.LUT1,
+                                            si.getSite(), lutInput.getBEL(), "I0", "O");
+                                }
+                            }
+                        } else {
+                            throw new RuntimeException(
```
eddieh-xlnx, line 3976: Am I right in understanding that this branch is reached if `used.contains(n) && (si.getCell("6LUT") != null || si.getCell("5LUT") != null)`?

Is this sufficient? `used.contains(n)` simply means that it is used by any net -- that net could well be entering that flop anyway (and `6LUT`/`5LUT` is used as comb output).

PR #893, opened 2023-11-22
[RWRoute]  Speedup: restrict enqueue of nodes that fanout to PINFEEDs

eddieh-xlnx:
In the same vein as #888, further speedup RWRoute by restricting the exploration of certain nodes that can fan-in to PINFEEDs (site pins):
1. `IntentCode.NODE_PINBOUNCE` nodes -- i.e. wires in the `INT` tile with a name prefix of `BOUNCE_*` or `BYPASS_*`. These nodes are sink nodes exclusive to a single connection (marked as type `RouteNodeType.PINFEED_I`) or can be a non-exclusive sink node/not used as a sink node allowing it to support any connection. Even before this PR, a heuristic exists to restrict such nodes to only be considered if they are in the tile above/below the sink node:
https://github.com/Xilinx/RapidWright/blob/b8ca42976cc784ee25c3185a05d47813c40c6f78/src/com/xilinx/rapidwright/rwroute/RWRoute.java#L1404-L1414
This heuristic remains unchanged (but `usablePINBounce()` renamed/refactored into `isAccessiblePinbounce()` to be more consistent with other similar methods).
2. Some `IntentCode.NODE_LOCAL` nodes -- specifically those with the wire names `INT_NODE_GLOBAL_*`, `INT_NODE_IMUX_*` or `INODE_*`. These nodes fan out to each other (and `BOUNCE_*` and `BYPASS_*`) as well as to `IntentCode.NODE_PINFEED`s that have the wire name prefix `IMUX_*` and `CTRL_*`.
    1. Apply the same heuristic as above for these `NODE_LOCAL`s -- this PR now makes those accessible only in the tiles above/below a sink -- that is, only enqueue such nodes if in the same column as the target INT tile, and only in the row above/below.
    2. However, for `INT` tile columns that service Laguna tiles (for multi-SLR devices), allow all `IntentCode.NODE_LOCAL` along the whole column.

An interesting characteristic of this collection (`{BYPASS,BOUNCE,INT_NODE_{GLOBAL,IMUX},INODE}_*`) of nodes is that they can only reach other nodes in the same column (i.e. same tile X coordinate) -- in fact, any such node can eventually reach all other tiles in the same column. This invariant is verified by two new tests: `TestNode.testNodeReachabilityUltraScale()`.

[...]

PR #895, opened 2023-11-25
[RWRoute] Add "--lutPinSwapping" option

eddieh-xlnx:
Depends on:
- [x] https://github.com/Xilinx/RapidWright/pull/893
- [x] https://github.com/Xilinx/RapidWright/pull/894

When routing connections to LUT pins, this new `--lutPinSwapping` option causes all pins on that LUT to be considered as targets too. When only the `x6LUT` BEL is used, all `A1-6` input pins can be used, but when `x5LUT` BEL is used only `A1-5` can be.

Once routing is complete, `LUTTools.swapLutPins()` is called to commit the swaps into the design. This step can be deferred by setting the system property `rapidwright.rwroute.lutPinSwapping.deferIntraSiteRoutingUpdates` to `true` after which `LUTTools.updateLutPinSwapsFromPIPs()` is needed to bring the design back to a valid state.

Current limitations:
* LUT pins that *actually* connect to other LUT cells (i.e. `H6LUT.A[1-6]` pins on `SLICEM` serve as the `WA` for all LUTs in that slice) are not allowed to be swapped (tracked by #901) -- `bnn.dcp` has an occurrence of this. To be clear -- `H[1-6]` that is not used as a `WA` for any other LUT `[A-G]` can still be swapped.
* `LOCK_PINS` properties are not respected yet.

PR #932, opened 2024-01-04
[RWRoute] Add --lutRoutethru option

eddieh-xlnx:
Add an *option* (disabled by default) that enables unused LUT resources to be routed through.

-----

When LUT routethrus are enabled, optimizations added in #888 and #893 must be fully and partially disabled. Recall that these optimizations prevent input site pins (`PINFEED`s) from being explored unless they're the target, and also prevent some more nodes that fanout to such input site pins when not in the vicinity of the target, respectively. For LUT routethrus to be considered, these optimizations must be fully/partially disabled, respectively.

Currently, this PR disables these optimizations across the entire device, meaning that all possible LUT routethrus in a connection's bounding box will be considered. A future optimization that may be beneficial would be to dynamically disable the #888 and #893 optimizations only in the vicinity of the source/sink tiles where routability is the most constrained.

----

Another intricacy is that the device model presents a number of routethru PIPs for each LUT; e.g. for the `A` LUT, there are 12 PIPs:
* `A[1-6]` -> `A_O`
* `A[1-6]` -> `AMUX`

One PIP for each of the 6 LUT inputs, crossed with one PIP for each SLICE output.
A limitation here appears to be that Vivado only allows one net to be routed through this LUT, meaning that you can't use `A_O` for one net and `AMUX` for another net by fracturing the LUT.

Thus, in this PR, only the `A[1-6]` -> `A_O` routethrus are considered, and all `AMUX` routethrus are discarded in order to guarantee that the LUT does not get fractured. A future optimization would be to emulate a PIP between `A_O` and `AMUX` so that both output nodes can still be considered for the same net.

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/RWRoute.java
+++ b/src/com/xilinx/rapidwright/rwroute/RWRoute.java
@@ -982,7 +982,7 @@ public class RWRoute{
 
                     if (pinName.charAt(1) == '6') {
                         throw new RuntimeException("ERROR: Illegal LUT routethru on " + site + "/" + pinName +
-                                " since 5LUT is being as a static source");
+                                " since the 5LUT is being used as a static source");
                     }
 
                     // Perform intra-site routing back to the LUT5 to not conflict with LUT6 routethru
--- a/src/com/xilinx/rapidwright/rwroute/RWRoute.java
+++ b/src/com/xilinx/rapidwright/rwroute/RWRoute.java
@@ -1,7 +1,7 @@
 /*
  *
  * Copyright (c) 2021 Ghent University.
- * Copyright (c) 2022-2023, Advanced Micro Devices, Inc.
+ * Copyright (c) 2022-2024, Advanced Micro Devices, Inc.
  * All rights reserved.
  *
  * Author: Yun Zhou, Ghent University.
@@ -577,17 +577,23 @@ public class RWRoute{
                         BEL bel = cell.getBEL();
                         assert(bel.isLUT());
                         String belName = bel.getName();
+                        String cellType = cell.getType();
                         if (belName.charAt(0) != lutLetter) {
-                            assert(cell.getType().startsWith("RAM"));
+                            assert(cellType.startsWith("RAM"));
                             // This pin connects to other LUTs! (e.g. SLICEM.H[1-6] also serves
                             // as the WA for A-G LUTs used as distributed RAM) -- do not allow any swapping
                             // TODO: Relax this when https://github.com/Xilinx/RapidWright/issues/901 is fixed
                             numberOfSwappablePins = 0;
                             break;
                         }
-                        if (bel.getName().startsWith("H") && cell.getType().startsWith("RAM")) {
+                        if (bel.getName().startsWith("H") && cellType.startsWith("RAM")) {
                             // Similarly, disallow swapping of any RAMs on the "H" BELs since their
                             // "A" and "WA" inputs are shared and require extra care to keep in sync
+                            numberOfSwappablePins = 0;
+                            break;
+                        }
+                        if (cellType.startsWith("SRL")) {
+                            // SRL* cells cannot support any pin swaps
                             numberOfSwappablePins = 0;
                             break;
                         }
@@ -1143,7 +1149,7 @@ public class RWRoute{
             }
 
             if (sinkRnode == rnodes.get(0)) {
-                List<Node> switchBoxToSink = RouterHelper.findPathBetweenNodes(sinkRnode.getNode(), connection.getSink().getConnectedNode());
+                List<Node> switchBoxToSink = RouterHelper.findPathBetweenNodes(sinkRnode, connection.getSink().getConnectedNode());
                 if (switchBoxToSink.size() >= 2) {
                     for (int i = 0; i < switchBoxToSink.size() - 1; i++) {
                         nodes.add(switchBoxToSink.get(i));
@@ -1155,14 +1161,14 @@ public class RWRoute{
 
                 // Assume that it doesn't need unprojecting back to the sink pin
                 // since the sink node is a site pin
-                assert(rnodes.get(0).getNode().getSitePin() != null);
+                assert(rnodes.get(0).getSitePin() != null);
             }
 
             for (RouteNode rnode : rnodes) {
                 nodes.add(rnode.getNode());
             }
 
-            List<Node> sourceToSwitchBox = RouterHelper.findPathBetweenNodes(connection.getSource().getConnectedNode(), connection.getSourceRnode().getNode());
+            List<Node> sourceToSwitchBox = RouterHelper.findPathBetweenNodes(connection.getSource().getConnectedNode(), connection.getSourceRnode());
             if (sourceToSwitchBox.size() >= 2) {
                 for (int i = 1; i <= sourceToSwitchBox.size() - 1; i++) {
                     nodes.add(sourceToSwitchBox.get(i));
@@ -1374,7 +1380,7 @@ public class RWRoute{
     private void addNodesDelays(NetWrapper net) {
         for (Connection connection:net.getConnections()) {
             for (RouteNode rnode : connection.getRnodes()) {
-                nodesDelays.put(rnode.getNode(), rnode.getDelay());
+                nodesDelays.put(rnode, rnode.getDelay());
             }
         }
     }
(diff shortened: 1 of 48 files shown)


eddieh-xlnx:
In the same style of #895 and #888, here's a comparison on the benchmarks from the [FPGA24 Routing Contest](https://xilinx.github.io/fpga24_routing_contest).

Before:
```
+--------------------------------+------+------------------+
| Benchmark                      | Pass | Wall Clock (sec) |
+--------------------------------+------+------------------+
| boom_med_pb_rwroute            | True |           262.08 |
| vtr_mcml_rwroute               | True |           279.52 |
| rosetta_fd_rwroute             | True |           196.19 |
| corundum_25g_rwroute           | True |           290.14 |
| vtr_lu64peeng_rwroute          | True |           272.99 |
| corescore_500_rwroute          | True |           185.01 |
| corescore_500_pb_rwroute       | True |           320.64 |
| mlcad_d181_lefttwo3rds_rwroute | True |          2042.37 |
| koios_dla_like_large_rwroute   | True |           466.00 |
| boom_soc_rwroute               | True |          1389.25 |
| ispd16_example2_rwroute        | True |           685.60 |
+--------------------------------+------+------------------+
```

Total: 6390 seconds.

After:
```
+--------------------------------+------+------------------+
| Benchmark                      | Pass | Wall Clock (sec) |
+--------------------------------+------+------------------+
| boom_med_pb_rwroute            | True |           294.62 |
| vtr_mcml_rwroute               | True |           364.77 |
| rosetta_fd_rwroute             | True |           222.55 |
| corundum_25g_rwroute           | True |           345.14 |
| vtr_lu64peeng_rwroute          | True |           323.03 |
| corescore_500_rwroute          | True |           215.31 |
| corescore_500_pb_rwroute       | True |           377.79 |
| mlcad_d181_lefttwo3rds_rwroute | True |          1193.92 |
| koios_dla_like_large_rwroute   | True |           549.58 |
| boom_soc_rwroute               | True |          1729.45 |
| ispd16_example2_rwroute        | True |           806.28 |
+--------------------------------+------+------------------+
```

Total: 6422 seconds (0.5% increase).

However, it looks like most benchmarks saw a small increase in runtime, except for `mlcad_d181_lefttwo3rds_rwroute` which had a massive halving.

-----

And some statistics from `TestRWRoute.testNonTimingDrivenFullRoutingWithGlobalNet()`:

Before:
```
Total wirelength:                       434815
Total INT tile nodes:                   837255
Total rnodes created:                  2860250
Average #children per node:                  4
Num iterations:                             14
Connections routed:                     374009
Nodes pushed:                        141263736
Nodes popped:                         65278765
```

After:
```
Total wirelength:                       399364
Total INT tile nodes:                   810400
Total rnodes created:                  3261557
Average #children per node:                  4
Num iterations:                             15
Connections routed:                     357638
Nodes pushed:                        211999922
Nodes popped:                        104715226
```

In terms of these RWRoute statistics, there is a noticeable wirelength saving, but a big increase in the nodes pushed/popped.

eddieh-xlnx:
All in all, it looks like this option only improves runtime for one benchmark (but would be expected to reduce wirelength for all benchmarks) so it's not enabled by default.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -908,96 +911,148 @@ private List<Connection> getCongestedConnections() {
...
+                    if (rnode == null || rnode.getOccupancy() == 0) {
+                        // No LUT6 routethru, nothing to be done
                         continue;
                     }
-                    connection.setSinkRnode(newSinkRnode);
 
-                    SitePin newSitePin = newSinkRnode.getNode().getSitePin();
-                    String existing = pinSwaps.put(oldSinkSpi, newSitePin.getPinName());
-                    assert(existing == null);
+                    // Only expect GND net to use SLICE outputs
```
clavin-xlnx, line 975: Perhaps beyond the scope of this, but there may be some rare cases where VCC is driven from LUTs.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -908,96 +911,148 @@ private List<Connection> getCongestedConnections() {
...
+
+                    // Perform intra-site routing back to the LUT5 to not conflict with LUT6 routethru
+                    BEL outmux = si.getBEL("OUTMUX" + lutLetter);
+                    si.routeIntraSiteNet(staticNet, outmux.getPin("D5"), outmux.getPin("OUT"));
+
+                    if (si.getDesign() == null) {
+                        // Rename SiteInst (away from "STATIC_SOURCE_<siteName>") and
+                        // attach it to the design so that intra-site routing updates take effect
+                        si.setName(site.getName());
+                        design.addSiteInst(si);
```
clavin-xlnx, line 986: Can you demonstrate that this doesn't cause an error when importing the designs back into Vivado?  Unless a placed cell exists (which a route thru may suffice), my understanding is that this could lead to an issue.

src/com/xilinx/rapidwright/rwroute/RouteNodeGraph.java
```diff
@@ -536,4 +573,21 @@ public boolean isAccessible(RouteNode child, Connection connection) {
...
+        }
+
+        if (!RouteThruHelper.isRouteThruPIPAvailable(design, head, tail)) {
+            return false;
+        }
+        assert(PIP.getArbitraryPIP(head, tail).isRouteThru());
+
+        BitSet bs = muxWires.get(tail.getTile().getTileTypeEnum());
+        if (bs != null && bs.get(tail.getWire())) {
+            // Disallow * -> [A-H]MUX routethrus
```
clavin-xlnx, line 588: Can you provide a bit more documentation on why we should disallow `[A-H]MUX` routethrus?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -908,96 +911,148 @@ private List<Connection> getCongestedConnections() {
...
+            }
+            LUTTools.swapMultipleLutPins(pinSwaps);
+        }
+
+        if (lutRoutethru) {
+            // When LUT routethru-s are considered, examine both static nets to find
+            // any cases where the *MUX output pin is used as a static source alongside
+            // the *_O output being used as a routethru. In such cases, configure the
+            // OUTMUX* site PIP to source from the 5LUT rather than the default 6LUT
+            // so that no conflict occurs
```
clavin-xlnx, line 947: Can you document the motivation for why we want to normalize on 6LUTs for routethrus and 5LUT for static nets?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
eddieh-xlnx: Removed this assertion as it wasn't enforcing any functionality.

src/com/xilinx/rapidwright/rwroute/RouteNodeGraph.java
eddieh-xlnx: Expanded the comment.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
eddieh-xlnx: Comment expanded.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
eddieh-xlnx: I'm seeing a breakpoint hit this line in the newly added test`testNonTimingDrivenFullRoutingWithLutRoutethru` with `bnn.dcp` for `SLICE_X79Y201`, and for `optical-flow.dcp` for `SLICE_X66Y252` and 10 others.

For `bnn.dcp`, even after RWRoute is done, that `SiteInst` has no placed cells, yet Vivado 2022.1 seems to accept it just fine and reports a clean routing result.

image
(highlighted net is `GLOBAL_LOGIC0`; rest of slice is completely empty)

Vivado 2022.1 also accepts this `optical-flow.dcp` routed result.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -978,8 +980,10 @@ protected void postRouteProcess() {
                         continue;
                     }
 
-                    // Only expect GND net to use SLICE outputs
-                    assert(staticNet.getType() == NetType.GND);
+                    if (pinName.charAt(1) == '6') {
+                        throw new RuntimeException("ERROR: Illegal LUT routethru on " + site + "/" + pinName +
+                                " since 5LUT is being as a static source");
```
clavin-xlnx, line 985:
```suggestion
                                " since 5LUT is being used as a static source");
```

src/com/xilinx/rapidwright/rwroute/RWRoute.java
clavin-xlnx: We probably need to upgrade our terminology as there are "routethrus" that terminate at a cell in the site and 
other "routethrus" that don't (being specified by a `PIP`).

clavin-xlnx (approved):

PR #944, opened 2024-01-17
[EDIFTools] Adds support for lower hierarchy connection

clavin-xlnx:
Adds support for the case where the connection to be made is lower in the hierarchy (descendant) rather than a parent.


eddieh-xlnx (commented):
Would you be kind enough to demonstrate this with a test?

There's a pretty basic test at `TestEDIFTools.testConnectPortInstsThruHier()` that you duplicate or parameterize?

eddieh-xlnx:
@clavin-xlnx Is this superseded by https://github.com/Xilinx/RapidWright/pull/1180?

PR #945, opened 2024-01-19
[RWRoute] Replace main src with altsrc if main is unused

clavin-xlnx:


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -1013,6 +1013,7 @@ protected void postRouteProcess() {
             SitePinInst altSource = net.getAlternateSource();
             SiteInst si = source.getSiteInst();
             boolean altSourcePreviouslyRouted = altSource != null ? altSource.isRouted() : false;
+            boolean sourcePreviouslyRouted = source != null ? source.isRouted() : false;
```
eddieh-xlnx, line 1016: It's an unwritten invariant that `source != null` -- perhaps we should have an `assert()` here just in case.
(It comes up later too)

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -1066,6 +1067,13 @@ protected void postRouteProcess() {
                     source.getSiteInst().routeIntraSiteNet(net, altSource.getBELPin(), altSource.getBELPin());
                 }
             }
+            // If the original source was routed and is now not routed but altSource is,
+            // let's remove it and make it the main source
+            if (source != null && sourcePreviouslyRouted && !source.isRouted() && altSource != null
+                    && altSource.isRouted()) {
+                net.removePin(source, true);
+            }
+
```
eddieh-xlnx, line 1076:
```suggestion
            // If the original source was routed and is now no longer used, it must
            // mean that the alternate source is used instead. Let's remove the main
            // source and promote the alternate to main.
            if (sourcePreviouslyRouted && !source.isRouted()) {
                assert(altSource != null && altSource.isRouted());
                net.removePin(source, true);
                assert(net.getSource() == altSource());
                assert(net.getAlternateSource() == null);
            }
```

eddieh-xlnx (commented):
A suggested change.

Perhaps rename title to: "[RWRoute] Replace main src with altsrc if main is unused" ?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
clavin-xlnx, line 1016:
```suggestion
            assert(source != null);
            boolean sourcePreviouslyRouted = source.isRouted();
```

eddieh-xlnx (approved):

PR #949, opened 2024-01-26
[RWRoute] When removing unused source SPI restore intra-site routing

eddieh-xlnx:
Removal from https://github.com/Xilinx/RapidWright/pull/945

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/RWRoute.java
+++ b/src/com/xilinx/rapidwright/rwroute/RWRoute.java
@@ -1064,7 +1064,6 @@ public class RWRoute{
                 assert(source.isRouted());
                 altSource.getSiteInst().removePin(altSource);
                 net.removePin(altSource); // TODO: Check why this doesn't have preserveOtherRoutes=true
-                if (true) throw new RuntimeException();
                 assert(source.isRouted());
                 if (source.getName().endsWith("MUX")) {
                     assert(altSource.getName().endsWith("_O"));
--- a/src/com/xilinx/rapidwright/rwroute/RWRoute.java
+++ b/src/com/xilinx/rapidwright/rwroute/RWRoute.java
@@ -1057,15 +1057,18 @@ public class RWRoute{
                     break;
                 }
             }
+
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -1076,6 +1076,8 @@ protected void postRouteProcess() {
                 net.removePin(source, true);
                 assert(net.getSource() == altSource);
                 assert(net.getAlternateSource() == null);
+                // Restore the intra-site routing on the *_O site wire removed by removePin()
+                altSource.getSiteInst().routeIntraSiteNet(net, source.getBELPin(), source.getBELPin());
```
clavin-xlnx, line 1080: I think this makes sense for the case where `source==*_O` and `altSource==*MUX`, but what about the reverse? 

If `source==*MUX` and `altSource==*_O`, then aren't we are re-annotating the MUX output with the net?

clavin-xlnx: After the call to `removePin()`, if we insert a `source = net.getSource()`, before this line, then I think it fixes the issue where its a problem in the first case and no-op in the second.

eddieh-xlnx: Are you suggesting that we replace the diff with:
```
                source = altSource;
                si.routeIntraSiteNet(net, source.getBELPin(), source.getBELPin());
```
In this case, we'd be just routing the `*MUX` wire. The original intention was to route the `*_O` wire only. Alternatively, we'd have to compute the correct source BEL pin to reroute the entire path.

You do make a good point about the inverse case though. The simplest way would be to gate it with an `if` statement like the `if` block above.

clavin-xlnx (approved):

src/com/xilinx/rapidwright/rwroute/RWRoute.java
clavin-xlnx: Sure, gating with `if` is the way to go.

PR #953, opened 2024-02-06
[GlobalSignalRouter] No intra site routing for new static source pins

eddieh-xlnx:
To be conservative, do not allow GlobalSignalRouter to update any intra-site routing, including those sitewires corresponding to site pins.

This means that we can no longer fracture existing user LUTs (placed at `[A-H]6LUT`) so that the `[A-H]5LUT` can serve as a static source, losing some of the routability benefits provided by https://github.com/Xilinx/RapidWright/pull/914, but in return being more robust.

Quantifying this loss in `TestGlobalSignalRouting.testRouteStaticNet()` operating on `optical-flow.dcp`:
* Number of static ground sources before: 1952, now: 1858 (-4.8%)
* Number of static ground PIPs before: 39777, now: 39827 (+0.13%)

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
+++ b/src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
@@ -442,7 +442,9 @@ public class GlobalSignalRouting {
                             "net '" + spi.getNet().getName() + "'");
                 }
             }
-            SitePinInst spi = currNet.createPin(sitePin.getPinName(), si);
+            SitePinInst spi = new SitePinInst(sitePin.getPinName(), si);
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -529,13 +532,22 @@ private static boolean isNodeUsableStaticSource(Node node, NetType type, Design
 
             String sitePinName;
             if (wireName.endsWith("_O")) {
+                // Fall through
```
eddieh-xlnx, line 535:
```suggestion
```

clavin-xlnx (approved):

PR #954, opened 2024-02-07
[RWRoute] Preserve [A-H]_O node when [A-H]MUX used as static src

eddieh-xlnx:
This prevents `*_O` and `*MUX` from being used for two different static sources, and also prevents *`_O` from being later used as a LUT routethru (support for which was added in #932).

Along the lines of https://github.com/Xilinx/RapidWright/pull/953 which unwound modification of intra-site routing for static nets.

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/RWRoute.java
+++ b/src/com/xilinx/rapidwright/rwroute/RWRoute.java
@@ -506,9 +506,8 @@ public class RWRoute{
 
             preserveNet(staticNet, false);
 
-            // When a [B-H]MUX pin is used as a static source, also preserve the [B-H]_O pin
+            // When a [A-H]MUX pin is used as a static source, also preserve the [A-H]_O pin
             // so that it can't be used by other static nets, nor as a LUT routethru
-            // Assume that AMUX can always supply GND
             for (SitePinInst spi : staticNet.getPins()) {
                 if (!spi.isOutPin()) {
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -505,6 +505,28 @@ protected void routeStaticNets() {
...
+
+            // When a [A-H]MUX pin is used as a static source, also preserve the [A-H]_O pin
+            // so that it can't be used by other static nets, nor as a LUT routethru
+            for (SitePinInst spi : staticNet.getPins()) {
+                if (!spi.isOutPin()) {
+                    continue;
+                }
+
+                SiteInst si = spi.getSiteInst();
+                if (!Utils.isSLICE((si))) {
```
clavin-xlnx, line 517:
```suggestion
                if (!Utils.isSLICE(si)) {
```
extra parens?

clavin-xlnx (approved):
I'm fine with this, but you could also consider this (more rare) scenario:

image

Seems like for `AMUX`, you could have `GND` output and `VCC` on `A_O` without an issue.

eddieh-xlnx:
Unfortunately, it looks like Vivado requires an intra-site routing update even for `AMUX` when `A_O` is used, otherwise we get an  "Invalid Programming for Site" error.

image

PR #961, opened 2024-02-21
[EDIFTokenizer] Handle tokens that wraparound the buffer correctly

eddieh-xlnx:


eddieh-xlnx:
Got the completely wrong end of the stick here. Let me open a new PR.

PR #962, opened 2024-02-22
[EDIFTokenizer] Account for byte size of UTF-8 characters correctly

eddieh-xlnx:
Correct fix for https://github.com/Xilinx/RapidWright/pull/961, where the previous inability to account for the size of Unicode characters caused the wrong offsets to be calculated and the wrong part of the buffer to be read/written to.


clavin-xlnx (approved):
I can't detect anything wrong by looking at this.

PR #972, opened 2024-03-25
[SLRCrosserGenerator] Adds North/South parameterizable bus widths; some error checking

clavin-xlnx:
This updates the `SLRCrosserGenerator` tool to be able to independently update bus widths of signals traveling North and South.  Previously, the tool required the bus widths to be equal in width in both directions.  The relevant options are now:

`-w` : Sets the bus width for both North and South
`-j`: Sets the bus width for the North bound signals
`-k`: Sets the bus width for the South bound signals

`-w` cannot be used at the same time as `-j` or `-k`.  Also, using `-j` or `-k` requires the other so that both directions are unambiguously specified.  Error checking has been to ensure these requirements are met.  It also checks to ensure that the bus widths do not exceed a total of 1440 bits for each `LAGUNA` column.

### DIFF
--- a/src/com/xilinx/rapidwright/examples/SLRCrosserGenerator.java
+++ b/src/com/xilinx/rapidwright/examples/SLRCrosserGenerator.java
@@ -653,6 +653,10 @@ public class SLRCrosserGenerator {
                     + "limitations: total bus width = " + busWidth
                     + ", each LAGUNA column can only support a width of " + lagunaColumnCrossingWidth);
         }
+        if (northBusWidth < 0 || southBusWidth < 0) {
+            throw new RuntimeException("ERROR: Cannot have a negative bus width: North=" 
+                    + northBusWidth + ", South=" + southBusWidth); 
+        }
 
         Double clkPeriodConstraint = null;
         if (opts.hasArgument(CLK_CONSTRAINT_OPT)) {
--- a/src/com/xilinx/rapidwright/examples/SLRCrosserGenerator.java
+++ b/src/com/xilinx/rapidwright/examples/SLRCrosserGenerator.java
@@ -261,33 +261,34 @@ public class SLRCrosserGenerator {
 
     /**
      * Places and routes an SLR crossing given a north and south bus of size width.
-     * @param d The current design
-     * @param northStart The starting Laguna site to start placement (placement moves north)
-     *     for the north traveling bus
-     * @param northBusName Full hierarchical net name of the bus to cross SLR in north direction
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/examples/SLRCrosserGenerator.java
```diff
@@ -432,15 +433,18 @@ public static void createBUFGCE(Design d, String clkName, String clkInName, Stri
...
-     * @param d Current design
-     * @param busWidth Width of the buses to create
-     * @param busPrefixes Prefixes to use for bus names
-     * @param clkName Name of the clock net
-     * @param clkInName Name of the clock in port
-     * @param clkOutName Name of the clock out port
+     * 
+     * @param d              Current design
+     * @param northBusWidth  Width of the North buses to create
+     * @param sorthBusWidth  Width of the South buses to create
```
eddieh-xlnx, line 439:
```suggestion
     * @param southBusWidth  Width of the South buses to create
```

eddieh-xlnx: Actually, there's a lot of the misspelling: `sorth` used elsewhere in this file

eddieh-xlnx (commented):
I tried `-j 0` and it failed (`-k 0` was fine though, as was `-j 1`):
```
$ bin/rapidwright SLRCrosserGenerator -j 0 -k 256
==============================================================================
==                           SLRCrosserGenerator                            ==
==============================================================================
                    Init:     2.319s
          Create Netlist:     0.016s
Exception in thread "main" java.lang.NullPointerException
	at com.xilinx.rapidwright.examples.SLRCrosserGenerator.placeAndRouteLagunaFlopPair(SLRCrosserGenerator.java:168)
	at com.xilinx.rapidwright.examples.SLRCrosserGenerator.placeAndRouteSLRCrossing(SLRCrosserGenerator.java:289)
	at com.xilinx.rapidwright.examples.SLRCrosserGenerator.main(SLRCrosserGenerator.java:708)
	at com.xilinx.rapidwright.MainEntrypoint.main(MainEntrypoint.java:223)
```
Generated circuit seems okay fine.

Also looks like the default `clkPeriodConstraint` member (line 550) is never used, and is shadowed by the local in `main()` (line 658) which is potentially confusing.

eddieh-xlnx (approved):
Looks like my previous issues were resolved!

A final suggestion I have is to error out when `w <= 0` or `j < 0 || k < 0 || j + k == 0` to catch human typos or any bugs caused by a calling script. (I was able to generate a DCP with `-w -1` which turned out to be empty)

PR #973, opened 2024-03-26
[RWRoute,PhysNetlistReader] Set logical driver on PIPs

eddieh-xlnx:
For nets with multiple source pins (e.g. `A_O` and `AMUX`) set the logical driver flag on the first PIP leaving the primary source.

### DIFF
--- a/src/com/xilinx/rapidwright/interchange/PhysNetlistReader.java
+++ b/src/com/xilinx/rapidwright/interchange/PhysNetlistReader.java
@@ -26,6 +26,7 @@ package com.xilinx.rapidwright.interchange;
 import com.xilinx.rapidwright.design.AltPinMapping;
 import com.xilinx.rapidwright.design.Cell;
 import com.xilinx.rapidwright.design.Design;
+import com.xilinx.rapidwright.design.DesignTools;
 import com.xilinx.rapidwright.design.Net;
 import com.xilinx.rapidwright.design.NetType;
 import com.xilinx.rapidwright.design.SiteInst;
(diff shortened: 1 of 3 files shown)


src/com/xilinx/rapidwright/interchange/PhysNetlistReader.java
```diff
@@ -521,6 +523,32 @@ protected void readRouting(StructList.Reader<PhysNet.Reader> nets, Consumer<Net>
                 net.addPIP(pip);
             }
             stubWires.clear();
+
+            // Nets with more than one routed source (e.g. A_O and AMUX) should have
+            // the first primary source PIP marked as a logical driver
+            SitePinInst altSource = net.getAlternateSource();
```
clavin-xlnx, line 529: Should we avoid running this code for the nets `GLOBAL_LOGIC[01]`?  At the least, it might help runtime.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -1423,10 +1423,25 @@ protected void setPIPsOfNets() {
         for (Entry<Net,NetWrapper> e : nets.entrySet()) {
             NetWrapper netWrapper = e.getValue();
             Net net = netWrapper.getNet();
+
```
clavin-xlnx, line 1426: Same question as above, do we want to exclude GND/VCC nets?

src/com/xilinx/rapidwright/interchange/PhysNetlistReader.java
eddieh-xlnx: Good spot! f6b92f9 makes it only operate on wire nets (not either static, nor the used net) and assert that it is not a clock net either.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
eddieh-xlnx: Static/global nets are dealt with by `GlobalSignalRouter` and should not be processed here. Added an assert just to be clear: 8317f6d

clavin-xlnx (approved):

PR #979, opened 2024-04-02
[RWRoute] Fix logical driver flag setting for DCP write

clavin-xlnx:
The DCP writer depends on nets with a logical driver flag (used only for nets that use multiple site output pins for the same logical routing tree) to be the first source `PIP` in the list of pips on the `Net`.  These changes modify the addition of this flag in RWRoute such that it adheres to this convention.


eddieh-xlnx:
Fixes issue introduced by #973

eddieh-xlnx (approved):

PR #987, opened 2024-04-25
BlockPlacer2: Fix off by one error in selecting module instance to move

jakobwenzel:
I have identified an off by one error in the `BlockPlacer2` when we try to select a module instance to move: The upper bound for `Random.nextInt` is exclusive, but we subtract 1 anyway. Therefore we never select the very last module instance in the list of module instances. It will just stay at its initial placement position forever (unless it gets pushed away by overlapping with another instance).

This issue seems to have been there since the very beginning of RapidWright: https://github.com/Xilinx/RapidWright/blob/17af8990400fd86b79bf5e3284d3ba746aa643bc/com/xilinx/rapidwright/placer/blockplacer/BlockPlacer2.java#L401

@clavin-xlnx: This might explain some bad placement results we've seen in the past.


clavin-xlnx (approved):
Thanks @jakobwenzel!  Indeed this is a bug, thank you for finding and fixing it!

Issue #986, opened 2024-04-25
Alveo U250 DSP48E2_{X28|X29} cell: <LOCKED>(BEL: (unplaced))


RipperJ:
Hi. When using `DesignTools.calculateUtilization(...)` on a design implemented on Alveo U250, the two columns of DSPs on the right side (beside the Vitis default static region) are locked and hence triggers exceptions in `public static Map<UtilizationType, Integer> calculateUtilization(Collection<SiteInst> siteInsts)`, when `c.getBELName()` is found null.

clavin-xlnx:
Thank you for reporting this issue, please give #988 a try, it should resolve the NPE.

RipperJ:
Yes #988 resolves this issue. Thanks.

PR #988, opened 2024-04-25
[DesignTools] Fix NPE in calculateUtilization()

clavin-xlnx:
This is an attempt to fix the issue reported in #986.


RipperJ:
This update resolves the issue #986 .

PR #990, opened 2024-04-29
Add FileTools.getAutoBufferedInputStream() with zstd auto-detect

eddieh-xlnx:


clavin-xlnx (approved):
Thanks for creating this!  Do you think a test is valuable in this case?

clavin-xlnx (approved):

test/src/com/xilinx/rapidwright/util/TestFileTools.java
```diff
@@ -106,4 +112,51 @@ public void testIsGzippedFile(@TempDir Path tmpPath) {
...
+        PipedInputStream pis = new PipedInputStream(1024);
+        PipedOutputStream pos = new PipedOutputStream();
+        pos.connect(pis);
+
+        byte[] cbuf;
+        if (mode == 0) {
+            // No compression
+            cbuf = ubuf;
+        }
+        else if (mode == 1) {
```
eddieh-xlnx, line 133:
```suggestion
        } else if (mode == 1) {
```

PR #997, opened 2024-06-10
2024.1 DCP Write Test

clavin-xlnx:
Tests RapidWright's ability to write a 2024.1 (physDB) DCP.  This checks that the design can be exported using 2024.1 physDB related components of the DCP.


eddieh-xlnx (approved):
Looks good. I've merged https://github.com/eddieh-xlnx/RapidWrightDCP/pull/29 so please update to latest master before merging. Note that the GHA submodules check doesn't seem to be working, possibly because the target branch is `2024.1.0` rather than `master`?

PR #1002, opened 2024-06-12
Node.getAllWiresInNode() returns empty list when Vivado doesn't

eddieh-xlnx:
```
link_design -part xcau10p-ubva368-2-i; llength [get_wires -of [get_nodes INT_IBRK_FSR2IO_X0Y140/IO_TO_CTR_FT0_3]]
```
returns 22 wires.

However, the testcase in this PR returns no wires.

### DIFF
--- a/test/src/com/xilinx/rapidwright/device/TestNode.java
+++ b/test/src/com/xilinx/rapidwright/device/TestNode.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2023, Advanced Micro Devices, Inc.
+ * Copyright (c) 2022-2024, Advanced Micro Devices, Inc.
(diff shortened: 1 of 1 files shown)


clavin-xlnx:
After some investigation, it appears that the node information is not present in our source information and even scouring across the device to get the alternate tile patterns, the information is simply not available.  This will likely require a new spin of device files and will probably take some time to fix.

eddieh-xlnx:
Interestingly, it seems that:
```
get_pips -of [get_nodes INT_IBRK_FSR2IO_X0Y140/IO_TO_CTR_FT0_3]
```
also returns nothing, meaning that this isn't a terribly important node to begin with.

PR #1005, opened 2024-06-13
Test for wire/node mismatch reported in #983

### LINKED ISSUE #983: xcvh1782 wire/node mismatch
When iterating through all wires and nodes in a device, this code below should never output anything. It does for xcvh1782. It would be good to add to the tests.

```
for (Wire wire: node.getAllWiresInNode()) {
    if(!wire.getNode().equals(node)) {
        System.out.println(node.toString() + " != " + wire.toString() + ".getNode() " + wire.getNode().toString());
    }
}
```

eddieh-xlnx:
Fixes #983

### DIFF
--- a/test/src/com/xilinx/rapidwright/device/TestNode.java
+++ b/test/src/com/xilinx/rapidwright/device/TestNode.java
@@ -152,5 +152,22 @@ public class TestNode {
         }
         System.out.println("visited.size() = " + visited.size());
     }
+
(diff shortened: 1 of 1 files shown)


clavin-xlnx (approved):
> Fixes #983

This is a test that specifically checks for the deficiency in #983 has been fixed, which now is included in 2024.1.0.

PR #1006, opened 2024-06-13
[EDIFTools] Adding method to create a flat netlist from a hierarchical one

clavin-xlnx:
Adding a method to `EDIFTools` that creates a new flat netlist from an existing hierarchical one.


src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -488,6 +488,18 @@ public boolean isPrimitive() {
         return getLibrary().getName().equals(EDIFTools.EDIF_LIBRARY_HDI_PRIMITIVES_NAME) && isLeafCellOrBlackBox();
     }
 
+    public boolean isStaticSource() {
+        return isPrimitive() && isVCCSource() || isGNDSource();
```
eddieh-xlnx, line 492:
```suggestion
        return isPrimitive() && (isVCCSource() || isGNDSource());
```

eddieh-xlnx (approved):

PR #1016, opened 2024-06-24
flatten netlist with shapes

zhilix:


src/com/xilinx/rapidwright/placer/dreamplacefpga/DREAMPlaceFPGA.java
```diff
@@ -204,23 +214,24 @@ public static Design placeDesign(Design design, Path workDir, boolean makeOutOfC
     }
 
     public static void main(String[] args) {
-        // Usage: <input DCP> <output DCP> [--out_of_context] [work directory]
-        if (args.length < 2 || args.length > 4) {
-            System.out.println("USAGE: <input DCP> <output DCP> [" 
-                            + PhysicalNetlistToDcp.MAKE_DCP_OUT_OF_CONTEXT + "] [work directory]");
+        // Usage: <input DCP> <output DCP> <--out_of_context/design.pl> [work directory]
```
clavin-xlnx, line 217:
```suggestion
        // Usage: <input DCP> <output DCP> <--out_of_context/design.pl> [work directory]
```
This makes the usage a bit confusing.  I think what you are trying to accomplish is that the user provide an IO placement file or the design should get marked as "out of context."  I think providing an `[IO placement file (.pl)]` should be fine.  If it is not there, then you can simply assume out of context and remove the command line option.  You could discriminate between the IO placement file and a work directory by checking if the third command line option is a directory or not.

src/com/xilinx/rapidwright/placer/dreamplacefpga/DREAMPlaceFPGA.java
```diff
@@ -151,12 +154,19 @@ public static void writeJSONForDREAMPlaceFPGA(Path jsonPath, Map<String, Object>
 
     }
 
-    public static Design placeDesign(Design design, Path workDir, boolean makeOutOfContext) {
+    public static Design placeDesign(Design design, String io_placemenet_file, boolean makeOutOfContext, Path workDir) {
```
clavin-xlnx, line 157: In Java, we tend to use camel case as a naming convention (as opposed to underscores).  Do you mind if we refactor `io_placement_file` to `ioPlacementFile`?

clavin-xlnx (approved):

PR #1018, opened 2024-06-26
Shapes v2, adds tags, fixes issues

clavin-xlnx:
This PR does the following
- Adds tags to the shapes (updates the schema to include them) 
- Maintains the order of the cells in the shapes relative to the shape file
- Also fixes a few issues: (1) the site types were not populated, now they are, (2) previously, if a run had generated the shapes file from Vivado there was an error, this now fixes that scenario.


zhilix:
Hi Chris, thank you for fixing the bugs, I checked out and it works fine now

PR #1019, opened 2024-06-30
[PblockRange]: fix null pointer corner case for BUFG_GT range

Licheng-Guo:
PBlock.getAllTiles() will NPE when the pblock contains BUFG_GT, e.g., BUFG_GT_X1Y96:BUFG_GT_X1Y119


clavin-xlnx:
This is actually a bug in the method `Site.getIntTile()`, this works fine as an intermediate fix, but we'll include the fix in an upcoming release.

PR #1020, opened 2024-06-30
[DesignTools.deletePblock()]: utility to delete a pblock

Licheng-Guo:
Utility to delete a pblock.

Only tested on Vivado-generated pblocks, i.e., those existing in an out-of-box Vivado DCP

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -4197,4 +4197,56 @@ public class DesignTools {
             updatePinsIsRouted(net);
         }
     }
+
+    /**
+     * (Untested) Remove the speicifed pblock from the XDC constraints of the design.
+     * 
+     * This utility will iterate through the lines of the XDC constraints and
+     * remove any lines that contain the specified pblock name and any pblock-related
+     * commands. The commands that are considered pblock-related are:
+     * - add_cells_to_pblock
+     * - create_pblock
+     * - get_pblocks
+     * - resize_pblock
+     * 
+     * This utility is only tested on Vivado-maintained XDC constraints.
+     * 
+     * @param design The design to remove the pblock from.
+     * @param pblockName The name of the pblock to remove.
+     */
+    public static void deletePblock(Design design, String pblockName) {
+        List<String> pblockCommands = Arrays.asList(
+            "add_cells_to_pblock",
+            "create_pblock",
+            "get_pblocks",
+            "resize_pblock"
+        );
+
+        boolean foundPblock = false; // Flag to track if any matching pblock is found
+
+        for (ConstraintGroup cg : ConstraintGroup.values()) {
+            // Get the current constraints for the constraint group
+            List<String> updatedConstraints = new ArrayList<>();
+
+            // Check if the line contains any pblock-related command and the pblockName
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -4197,4 +4197,56 @@ public static void updatePinsIsRouted(Design design) {
...
+     * - create_pblock
+     * - get_pblocks
+     * - resize_pblock
+     * 
+     * This utility is only tested on Vivado-maintained XDC constraints.
+     * 
+     * @param design The design to remove the pblock from.
+     * @param pblockName The name of the pblock to remove.
+     */
+    public static void deletePblock(Design design, String pblockName) {
```
eddieh-xlnx, line 4217: I would suggest calling it something more clear:
```suggestion
    public static void removeConstraintsReferencingPBlock(Design design, String pblockName) {
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -4197,4 +4197,56 @@ public static void updatePinsIsRouted(Design design) {
...
+        boolean foundPblock = false; // Flag to track if any matching pblock is found
+
+        for (ConstraintGroup cg : ConstraintGroup.values()) {
+            // Get the current constraints for the constraint group
+            List<String> updatedConstraints = new ArrayList<>();
+
+            // Check if the line contains any pblock-related command and the pblockName
+            for (String line : design.getXDCConstraints(cg)) {
+                boolean containsPblockCommand = pblockCommands.stream().anyMatch(line::contains);
+                boolean containsPblockName = line.contains(pblockName);
```
eddieh-xlnx, line 4234: Using `contains` seems quite brittle. Given `pblockName` is `foo`, the following would match:
- `create_pblock foobar`
- `... [get_pblocks notfoo]`
- `add_cells_to_pblock ... [get_cells foo]`

Now the robustness of this is certainly limited by RapidWright not having a XDC parser, but I would still prefer to see some stricter matching, perhaps using regex.

eddieh-xlnx (changes requested):
Thanks for the contribution @Licheng-Guo! Can you give a motivating use-case for when this method is useful? I think I'd like to see a testcase as part of this PR so it's no longer "untested".

@clavin-xlnx Since `DesignTools` is already so big, and since there could be more manipulation of the XDCs, perhaps we should create a new `ConstraintTools` class?

Licheng-Guo:
> Can you give a motivating use-case for when this method is useful? 

The current design merger does not handle XDCs, thus the final output has obsolete constraints. One simple hack for now is to delete the ones no longer needed.

> I think I'd like to see a testcase as part of this PR so it's no longer "untested".

I can put a test DCP into RapidWrightDCP repo and add a test here

> @clavin-xlnx Since DesignTools is already so big, and since there could be more manipulation of the XDCs, perhaps we should create a new ConstraintTools class?

I agree, we might need more XDC manipulation when stitching complex designs.

clavin-xlnx:
> @clavin-xlnx Since `DesignTools` is already so big, and since there could be more manipulation of the XDCs, perhaps we should create a new `ConstraintTools` class?

Sure, I think a `ConstraintsTools` class would be good, make sense to me.

PR #1009, opened 2024-06-19
[DesignTools.makeBlackBox()] Fixes an issue of removing CARRY blocks fed by routethrus

clavin-xlnx:
Also removes missed cells from encrypted modules inside the black box module to be removed.

PR #1023, opened 2024-07-02
[DesignTools.makeBlackBox()] Remove flawed loop intended for encrypted cell removal

clavin-xlnx:
In #1009, a bad loop was added in an attempt to remove cells that appeared to be in encrypted cells that were part of the black box.  This loop has a flaw in it in that any cells with similar names, but not in the black box can be removed accidentally.  This causes a failure on an internal test and so it is best removed for now.


eddieh-xlnx (approved):
Passing on that internal test now.

PR #1029, opened 2024-07-12
[EDIFTools] writeTclLoadScriptForPartialEncryptedDesigns abspath

eddieh-xlnx:
Write `*_load.tcl` into same dir as DCP rather than into current working dir


clavin-xlnx (approved):
Thanks!

PR #1030, opened 2024-07-12
[RWRoute] Adjust wire base costs + cross SLR estimates

eddieh-xlnx:
1. Adjust wire base costs such that:
  i. A `NODE_SINGLE`/`NODE_DOUBLE` can (typically) no longer lead to a lower total path cost than the `NODE_LOCAL` that feeds it, even if it gets us closer to the sink. The reasoning is that we want all the equivalent-cost `NODE_LOCAL`-s to be explored before allowing single/double wires to be taken, in case there is a chance that a longer wire (double/quad) that can get us there even more efficiently.

Take for example the following, where the sink is solely to its north:
```
LAG_LAG_X102Y359/RXD1 -> INT_X103Y359/INT_NODE_SDQ_74_INT_OUT0 [C0] -> INT_X103Y359/EE4_W_BEG4
                                                                    -> INT_X103Y359/NN1_W_BEG4  [C1]
                                                                    -> INT_X103Y359/NN2_W_BEG4  [C2]
                                                                    -> INT_X103Y359/WW4_W_BEG4
                      -> INT_X103Y359/INT_NODE_SDQ_74_INT_OUT1 [C0] -> INT_X103Y359/EE2_W_BEG5
                                                                    -> INT_X103Y359/NN1_W_BEG5
                                                                    -> INT_X103Y359/NN4_W_BEG4  [C3]
                                                                    -> INT_X103Y359/WW2_W_BEG5
                      -> INT_X103Y359/INT_NODE_SDQ_76_INT_OUT0 [C0] -> ...
                      -> INT_X103Y359/INT_NODE_SDQ_76_INT_OUT0 [C0] -> ...
```

Now, expansion of `LAG_LAG_X102Y359/RXD1` would place all `INT_NODE_SDQ_7[46]_INT_OUT[01]` nodes onto the priority queue, all with the same cost of `C0`.

If `NN1` or `NN2`'s cost (`C1`/`C2`) was smaller than `C0`, then RWRoute would explore those nodes ahead of the other `INT_NODE_SDQ`s and miss out on the `NN4` (which potentially leads to a `NN12`)

Ideally, we want: `C1 (NN1) > C2 (NN2) > C0 > C3 (NN4)` assuming that the sink is at least 4 tiles away such that single and double wires are not explored before all the `SDQ` nodes are explored, but quads and longs that close the distance to the sink will take priority.

2. Reduce the base cost of local/default (i.e. zero length) nodes to encourage more exploration
3. Fix distance estimate to sink for cross-SLR connections
4. Do not explore `RouteNodeGraph.accessibleWireOnlyIfAboveBelowTarget` nodes (e.g. `INT_NODE_IMUX`/`BYPASS`/`BOUNCE`/etc.) nodes if no SLR crossing needed to reach sink


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -1911,6 +1911,7 @@ private float getNodeCost(RouteNode rnode, Connection connection, int countSameS
         // CRoute paper states that the bias factor cannot be more than half of the wire cost
         // (it may exceed this here because we may not be using the minimum-sized bounding box)
         float biasCost = rnode.getBaseCost() * Math.min(distToCenter * net.getBiasFactor(), 0.5f);
+        biasCost = 0;
```
eddieh-xlnx, line 1914:
```suggestion
```

PR #1031, opened 2024-07-13
[Interchange] Prototype multi-message support for logical and physical netlists

clavin-xlnx:
Splits up the logical and physical netlist Interchange writers/readers to use multiple messages.  This is to avoid out of memory issues and also to avoid hitting fundamental limits with Cap'n Proto for very large designs.  This builds on #1017.

The logical netlist that is reused inside the DeviceResources message is preserved as not multi-message.


clavin-xlnx:
Closing as not much interest for now.

PR #1039, opened 2024-07-27
[Interchange] Reorders tile types and tiles to follow their Vivado index

clavin-xlnx:
This PR changes the export order of tiles ([`tileList`](https://github.com/chipsalliance/fpga-interchange-schema/blob/c985b4648e66414b250261c1ba4cbe45a2971b1c/interchange/DeviceResources.capnp#L102)) and tile types ([`tileTypeList`](https://github.com/chipsalliance/fpga-interchange-schema/blob/c985b4648e66414b250261c1ba4cbe45a2971b1c/interchange/DeviceResources.capnp#L101)) in Interchange `.device` files such that the order reflects the original Vivado-given index.  

**Tiles:** Each tile in a device is given a unique address or `INDEX` that is often computed by `TOTAL_DEVICE_COLUMNS*TILE_ROW + TILE_COLUMN`.  This PR now orders the tiles in the device resources file according to this index.

**Tile Types:** The Interchange has a Tile Type object, but Vivado annotates tile type information on the tile.  Vivado uses a `TILE_TYPE_IDX` to denote the order of the tile type.  This is mostly alphabetical order except the 0-indexed entry is the `NULL` type.

The device verifier is also modified to verify that the tile orders match once the file is parsed.  Since tile type indices are not declared in the interchange schema, the ordering is not verified.

### DIFF
--- a/src/com/xilinx/rapidwright/interchange/DeviceResourcesVerifier.java
+++ b/src/com/xilinx/rapidwright/interchange/DeviceResourcesVerifier.java
@@ -339,6 +339,7 @@ public class DeviceResourcesVerifier {
             expect(tile.getTileTypeEnum().name(), tileTypeName);
             expect(tile.getRow(), tileReader.getRow());
             expect(tile.getColumn(), tileReader.getCol());
+            // Note: Tile.getUniqueAddress() is equivalent to the INDEX property on a Vivado Tile object
             expect(tile.getUniqueAddress(), i);
 
             // Verify Tile Types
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/interchange/DeviceResourcesWriter.java
```diff
@@ -694,13 +694,20 @@ public static Map<TileTypeEnum, Integer> writeAllTileTypesToBuilder(Design desig
 
         Map<TileTypeEnum, Integer> tileTypeIndicies = new HashMap<TileTypeEnum, Integer>();
 
-        int i=0;
+        // Order tile types by their TILE_TYPE_IDX
```
eddieh-xlnx, line 697:
```suggestion
        // Order tile types by their TILE_TYPE_IDX (may not be contiguous)
```

src/com/xilinx/rapidwright/interchange/DeviceResourcesVerifier.java
```diff
@@ -338,6 +339,7 @@ public static boolean verifyDeviceResources(String devResFileName, String device
             expect(tile.getTileTypeEnum().name(), tileTypeName);
             expect(tile.getRow(), tileReader.getRow());
             expect(tile.getColumn(), tileReader.getCol());
+            expect(tile.getUniqueAddress(), i);
```
eddieh-xlnx, line 342:
```suggestion
            // Note: Tile.getUniqueAddress() is equivalent to the INDEX property on a Vivado Tile object
            expect(tile.getUniqueAddress(), i);
```

eddieh-xlnx (approved):

PR #1040, opened 2024-07-31
[Interchange] Fixes to support Versal designs via Interchange

clavin-xlnx:
- Adds necessary Versal IO types
- Manages some intra-routing issues to properly store the BELPins necessary to communicate site routing within Versal sites
- Turns off obsolete checks


eddieh-xlnx (approved):
Looks like it passes all internal/external tests.

I'm approving this PR, but bear in mind that I did change things up quite a bit so you'll want to give that a once over.

PR #1046, opened 2024-08-09
[RWRoute] Refactoring/cleanup/preparation for multi-threading

eddieh-xlnx:
- Fold `RouteNodeImpl` into `RouteNode` which is now a non-abstract *static* class -- static should remove one reference (to the outer `RouteNodeGraph`) from all `RouteNode` objects -- achieved by making necessary methods require a `RouteNodeGraph` reference instead of relying on that outer reference.
- Convert a number of counters (`RWRoute.{connectionsRouted{,Iteration}`, `RouteNodeGraph.nodesMapSize`, etc.) to `AtomicInteger` or `AtomicLong` to facilitate multi-threaded use
- Create `RWRoute.getQueue()` for getting the `PriorityQueue` to allow downstream (multi-threaded) classes to override
- Create a `ConnectionState` class to encapsulate all state necessary only when routing a particular connection
- General tidying up

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/RWRoute.java
+++ b/src/com/xilinx/rapidwright/rwroute/RWRoute.java
@@ -774,7 +774,7 @@ public class RWRoute {
         RuntimeTracker routeWireNets = routerTimer.createRuntimeTracker("route wire nets", "Routing");
         routeWireNets.start();
         preRoutingEstimation();
-        routeIndirectConnections();
+        routeIndirectConnectionsIteratively();
         // NOTE: route direct connections after indirect connection.
         // The reason is that there maybe additional direct connections in the soft preserve mode for partial routing,
         // and those direct connections should be included to be routed
@@ -856,10 +856,18 @@ public class RWRoute {
         }
     }
 
+    protected void routeIndirectConnections(Collection<Connection> connections) {
+        for (Connection connection : connections) {
+            if (shouldRoute(connection)) {
+                routeIndirectConnection(connection);
+            }
+        }
+    }
+
     /**
      * Routes indirect connections iteratively.
      */
-    public void routeIndirectConnections() {
+    public void routeIndirectConnectionsIteratively() {
         sortConnections();
         initializeRouting();
         long lastIterationRnodeCount = 0;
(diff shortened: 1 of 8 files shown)


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -1484,20 +1495,59 @@ protected void checkPIPsUsage() {
...
+            this.connection = connection;
+            this.sequence = sequence;
+            this.rnodeCostWeight = rnodeCostWeight;
+            this.shareWeight = shareWeight;
+            this.rnodeWLWeight = rnodeWLWeight;
+            this.estWlWeight = estWlWeight;
+            this.dlyWeight = dlyWeight;
+            this.estDlyWeight = estDlyWeight;
+            this.queue = queue;
+            targets = new ArrayList<>();
```
clavin-xlnx, line 1525: Seems like adding a new object for each connection in a design could impact memory--even if it is just accumulating a bunch of parameters that were floating around previously.  It does look like the `targets` list is new.  Would it make sense to add them directly to the `Connection` object itself to avoid some of the overhead?

eddieh-xlnx: This is a new object that is only live when this `Connection` is being routed, which for the single-threaded case is at most one object (mostly removed from the function stack). Putting it into the `Connection` object means paying this cost for all connections regardless of whether they are currently being routed.

For the single-threaded case again, the `targets` list is moved from `RouteNodeGraph` into this new short-lived `ConnectionState` object.

I agree that creating and destroying `ConnectionState` now does create a more garbage though.

clavin-xlnx: Ok, makes sense so no real impact to memory footprint.  

You could consider a `static` `ThreadLocal` instance if you are worried about garbage collector thrashing, but I wonder if the latest JVMs compensate for this pattern.

clavin-xlnx (approved):

src/com/xilinx/rapidwright/rwroute/RWRoute.java
eddieh-xlnx: My first approach was to use `ThreadLocal`, but I'm assuming that a `ThreadLocal.get()` incurs a non-trivial overhead such that it doesn't make sense to call it inside the inner loop (e.g. the `push()` method called for each node that is put on the queue) if you can pass it as a function parameter instead.

PR #1053, opened 2024-08-27
[DesignTools] Conform to Vivado *RST* pin inversion site routing configuration 

clavin-xlnx:
as part of the `DesignTools.createCeSrRstPinsToVCC()` method which gets called by `RWRoute.preprocess()`

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -3302,15 +3302,15 @@ public class DesignTools {
     }
     
     private static void maybeCreateVccPinAndPossibleInversion(SiteInst si, String sitePinName, Net vcc, Net gndInvertibleToVcc) {
-        if (si.getSitePinInst(sitePinName) == null) {
-            SitePinInst sitePin = vcc.createPin(sitePinName, si);
-            if (gndInvertibleToVcc != null && sitePinName.contains("RST")) {
-                assert sitePin.getSiteWireBELPins().size() == 1;
-                BELPin pin = sitePin.getSiteWireBELPins().get(0);
-                assert pin.getBEL().getPins().length == 2;
-                si.routeIntraSiteNet(gndInvertibleToVcc, pin, pin);
-            }            
+        SitePinInst sitePin = si.getSitePinInst(sitePinName);
+        if (sitePin == null) {
+            sitePin = vcc.createPin(sitePinName, si);
         }
+        // For the RST inversion to be interpreted properly by Vivado, there must be no
+        // site routing
+        // on the path.
+        BELPin belPin = sitePin.getBELPin();
+        si.unrouteIntraSiteNet(belPin, belPin);
     }
 
     public static void createMissingStaticSitePins(BELPin belPin, SiteInst si, Cell cell) {
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -3306,11 +3306,12 @@ public class DesignTools {
         if (sitePin == null) {
             sitePin = vcc.createPin(sitePinName, si);
         }
-        // For the RST inversion to be interpreted properly by Vivado, there must be no
-        // site routing
-        // on the path.
-        BELPin belPin = sitePin.getBELPin();
-        si.unrouteIntraSiteNet(belPin, belPin);
+        if (gndInvertibleToVcc != null) {
+            // For the RST inversion to be interpreted properly by Vivado, there must be no
+            // site routing on the path around the inverter BEL
+            BELPin belPin = sitePin.getBELPin();
+            si.unrouteIntraSiteNet(belPin, belPin);
+        }
     }
 
     public static void createMissingStaticSitePins(BELPin belPin, SiteInst si, Cell cell) {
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3299,13 +3295,23 @@ public static void createCeSrRstPinsToVCC(Design design) {
...
     }
+    
+    private static void maybeCreateVccPinAndPossibleInversion(SiteInst si, String sitePinName, Net vcc, Net gndInvertibleToVcc) {
+        if (si.getSitePinInst(sitePinName) == null) {
+            SitePinInst sitePin = vcc.createPin(sitePinName, si);
+            if (gndInvertibleToVcc != null && sitePinName.contains("RST")) {
+                assert sitePin.getSiteWireBELPins().size() == 1;
+                BELPin pin = sitePin.getSiteWireBELPins().get(0);
+                assert pin.getBEL().getPins().length == 2;
+                si.routeIntraSiteNet(gndInvertibleToVcc, pin, pin);
```
eddieh-xlnx, line 3311: Looks like this only does intra-site routing for when GND is inverted to VCC. Do you want to do intra-site routing for when VCC is actually desired (e.g. for `CE`)?

eddieh-xlnx: Oh wait, my mistake. `CKEN`-s do not have an inverter.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3299,13 +3295,24 @@ public static void createCeSrRstPinsToVCC(Design design) {
...
+        SitePinInst sitePin = si.getSitePinInst(sitePinName);
+        if (sitePin == null) {
+            sitePin = vcc.createPin(sitePinName, si);
+        }
+        if (gndInvertibleToVcc != null) {
+            // For the RST inversion to be interpreted properly by Vivado, there must be no
+            // site routing on the path around the inverter BEL
+            BELPin belPin = sitePin.getBELPin();
+            si.unrouteIntraSiteNet(belPin, belPin);
+        }
```
eddieh-xlnx, line 3314: Would this be safer inside the `(sitePin == null)` condition so that we're only modifying the site routing if we were the one to insert it?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3299,13 +3295,24 @@ public static void createCeSrRstPinsToVCC(Design design) {
...
-                    if (si.getSitePinInst(pinName) == null) {
-                        vcc.createPin(pinName, si);
-                    }
+                    maybeCreateVccPinAndPossibleInversion(si, pinName, vcc, gndInvertibleToVcc);
                 }
             }
         }
     }
+    
+    private static void maybeCreateVccPinAndPossibleInversion(SiteInst si, String sitePinName, Net vcc, Net gndInvertibleToVcc) {
```
eddieh-xlnx, line 3304: Is this still the most descriptive method name?

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx: The intent of this method performs two tasks:
1. Ensure that the site pin exists (create it if it doesn't)
2. If architecturally relevant, modify the site routing (completely remove it) to ensure that the inversion is communicated.

If we guard that code, it seems like #2 would not be ensured.

eddieh-xlnx: Since the parent method is `DesignTools.createCeSrRstPinsToVCC()` I would argue that the main purpose is 1, and only when 1 is undertaken (which will paint the associated sitewire), to ensure 2. Right now, if the pin already exists, the sitewire is always unpainted which might be necessary if we're assuming that the design was faulty to begin with.

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: My slight irk was that this method doesn't do any inversion by itself anymore, since it doesn't set any site PIPs nor do any site routing; rather, it erases some site routing to allow inversion to occur. Let's leave it then.

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx: My concern is that RapidWright creates all site pins with the site wire attached painted (for better or worse).  If we encounter a site pin that was created in another method, the site wire would never be reset (the requirement for inversion).  I suppose we could change the site pin creation behavior, but I think that would touch and break a lot of things.

eddieh-xlnx: That's a very valid concern, in which case I would prefer that we fix them at/close to where those site pins were created (or come up with a more holistic solution). Note that `DesignTools.createCeSrRstPinsToVCC()` is only called by `DesignTools.createPossiblePinsToStaticNets()` which is always called for `RWRoute.preprocess()` but not always called by `PartialRouter.preprocess()`.

eddieh-xlnx: Another thought is -- do we need to set that sitewire at all for VCC/GND nets? If so, perhaps not ever setting it for these two special nets could be a better way?

clavin-xlnx: I think the cleanest solution is to simply stop setting the sitewire on site pin addition or creation.  Unfortunately, this will probably break a lot of downstream use cases.

eddieh-xlnx (approved):
An improvement on what we had previously.

PR #1056, opened 2024-09-10
[RWRoute] Always clear prev pointer of unpreserved RouteNode-s

eddieh-xlnx:
Very rare behaviour described below.

Context: Net1 is a fully routed net utilizing a number of nodes (Node1, Node2, Node3, ...)

1. During clock routing, it's decided that Node1 is needed to complete the clock net, and so the entirety of Net1 is unpreserved, and the result of the clock route is then preserved.
2. During static net routing, Node2 is determined to be needed to complete a static net. Since Node2 is no longer being preserved, the static net gets to preserve it. 

Note that because Node2 was unpreserved previously, it already exists in the routing graph as a `RouteNode`. 

3. Later, during signal routing, this node breaks the assumption that preserved nodes can only be used by the same net pointed to by its `prev` pointer.

This PR allows clears this `prev` pointer so that this assumption is maintained.

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/PartialRouter.java
+++ b/src/com/xilinx/rapidwright/rwroute/PartialRouter.java
@@ -354,7 +354,7 @@ public class PartialRouter extends RWRoute {
                             continue;
                         }
 
-                        assert (rnode.getPrev() != null);
+                        assert(rnode.getPrev() != null);
                     }
 
                     // RouteNode must only exist if its net has been unpreserved
@@ -366,12 +366,13 @@ public class PartialRouter extends RWRoute {
                     // from the beginning. However, since other objects (e.g. Connection-s) may hold a reference
                     // to this, that would not be wise.
                     // Instead, increment the use of this RouteNode with a null net (since global nets have
-                    // no corresponding NetWrapper) in order to flag it as being irreversibly used thus forcing
-                    // the non-global to find another path
+                    // no corresponding NetWrapper) in order to indicate to Connection.isCongested() that it needs
+                    // to be rerouted.
                     rnode.incrementUser(null);
                 }
             }
         }
+
         return unpreserveNets;
     }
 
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/rwroute/PartialRouter.java
```diff
@@ -321,40 +321,54 @@ private List<Net> unpreserveCongestedNets(Collection<Net> globalNets) {
...
-                        unpreserveNets.add(preservedNet);
-                    }
+                        rnode = routingGraph.getNode(node);
+                        if (rnode == null) {
+                            continue;
+                        }
+
+                        // This could be a leftover node from a previously unpreserved net
+                        // (with its prev pointer set to allow its routing to be recovered)
+                        // Fall through so that this can be cleared
```
eddieh-xlnx, line 334:
```suggestion
                        // Fall through so that this can be cleared and blocked
```

src/com/xilinx/rapidwright/rwroute/PartialRouter.java
```diff
@@ -224,7 +224,7 @@ protected NodeStatus getGlobalRoutingNodeStatus(Net net, Node node) {
             return NodeStatus.INUSE;
         }
         if (preservedNet != null) {
-            if (!softPreserve) {
+            if (!softPreserve || net.isStaticNet()) {
```
eddieh-xlnx, line 227:
```suggestion
            if (!softPreserve) {
```

src/com/xilinx/rapidwright/rwroute/PartialRouter.java
```diff
@@ -321,43 +321,58 @@ private List<Net> unpreserveCongestedNets(Collection<Net> globalNets) {
...
+                    // from the beginning. However, since other objects (e.g. Connection-s) may hold a reference
+                    // to this, that would not be wise.
+                    // Instead, increment the use of this RouteNode with a null net (since global nets have
+                    // no corresponding NetWrapper) in order to indicate to Connection.isCongested() that it needs
+                    // to be rerouted.
+                    rnode.incrementUser(null);
                 }
             }
         }
+
```
eddieh-xlnx, line 375:
```suggestion
```

clavin-xlnx (approved):

PR #1061, opened 2024-09-16
Test for Design.retargetPart()

clavin-xlnx:
This is a test that exercises the API on `Design` called `retargetPart(Part targetPart, int tileXOffset, int tileYOffset)`.

### DIFF
--- a/test/src/com/xilinx/rapidwright/design/TestDesign.java
+++ b/test/src/com/xilinx/rapidwright/design/TestDesign.java
@@ -39,6 +39,9 @@ import org.junit.jupiter.params.provider.ValueSource;
 
 import com.xilinx.rapidwright.device.BEL;
 import com.xilinx.rapidwright.device.Device;
+import com.xilinx.rapidwright.device.PIP;
+import com.xilinx.rapidwright.device.Part;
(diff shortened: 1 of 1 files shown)


test/src/com/xilinx/rapidwright/design/TestDesign.java
```diff
@@ -440,4 +443,49 @@ public void testPlaceIOB(String partName, String pkgPin) {
...
+
+            Assertions.assertEquals(targetPartName, d.getPartName());
+            ensureDesignInSLR(d, slr);
+
+            d.writeCheckpoint(output);
+
+            Design d2 = Design.readCheckpoint(output);
+
+            Assertions.assertEquals(targetPartName, d2.getPartName());
+            ensureDesignInSLR(d2, slr);
```
eddieh-xlnx, line 487: Should we check that Vivado can accept the retargeted/relocated result?

clavin-xlnx: Yes, I'll add that.  I'll also try adding a test to see if we can invert the process and arrive back at what we started with as this should be a reversible process.

eddieh-xlnx (approved):

PR #1065, opened 2024-09-19
[EDIF] Fixes rare bus renaming collision

clavin-xlnx:
There are two competing issues relevant to this fix.

1. There are unfortunately situations where an `EDIFCell` can have two `EDIFPort` objects, a single bit port and a bussed port with the same name, for example `my_signal` and `my_signal[1:0]`.  We resolve this naming collision by storing the bussed version with a `[` suffix, `my_signal[`.  
2. In some designs, it is possible to have two `EDIFPort` objects belonging to the same `EDIFCell` such that when one bussed name is EDIF legalized, it collides with another valid `EDIFPort` name.  For example, `phy_ctrl[1:0]` and `phy_ctrl_[13:0]`.  Due to the possible naming collision issues from issue 1, all bussed ports are given a `[` suffix to the name and thus the legal names of busses are all extended by one character, `phy_ctrl`  (original) -> `phy_ctrl[` (stored) -> `phy_ctrl_` (legalized EDIF name).  This leads to ambiguity for the other bused port on the cell and leads to reading problems of the netlist (see test added in the PR to illustrate).

This PR attempts to resolve both of these issue by removing the trailing `[` before legalization and also checking to see if there is indeed a colliding single-bit port name that collides with the bussed port name, if there is, it is given an additional suffix during EDIF naming legalization and stored in a separate map, otherwise, the renaming proceeds as normal.  This has a small impact to runtime (~5% EDIF write time).  This does not impact EDIF read.

This also fixes a connectivity issue when generating multipliers (`MultGenerator`) where it was blindly connecting all signals as busses.  This now detects if it is a bus or single bit port and connects accordingly.

### DIFF
--- a/test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
+++ b/test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
@@ -461,7 +461,7 @@ class TestEDIFNetlist {
 
         String portName = "unfortunate_name";
 
-        // Create two ports, one single-bit and another bussed with the same root name
+        // Create two bussed ports with a common root name, one having a trailing '_' underscore
         EDIFPort port0 = top.createPort(portName + "[2:0]", EDIFDirection.INOUT, 3);
         EDIFPort port1 = top.createPort(portName + "_[1:0]", EDIFDirection.INOUT, 2);
 
--- a/test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
+++ b/test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
@@ -464,6 +464,8 @@ class TestEDIFNetlist {
         // Create two bussed ports with a common root name, one having a trailing '_' underscore
         EDIFPort port0 = top.createPort(portName + "[2:0]", EDIFDirection.INOUT, 3);
         EDIFPort port1 = top.createPort(portName + "_[1:0]", EDIFDirection.INOUT, 2);
+        // And for good measure, throw in a single-bit port too with the same root
+        EDIFPort port2 = top.createPort(portName, EDIFDirection.INOUT, 1);
 
         EDIFNet net0 = top.createNet("net0");
         net0.createPortInst(port0, 0);
@@ -472,6 +474,10 @@ class TestEDIFNetlist {
         EDIFNet net1 = top.createNet("net1");
         net1.createPortInst(port1, 1);
         net1.createPortInst("R", ff);
+        
+        DIFNet net2 = top.createNet("net2");
+        net2.createPortInst(port2);
+        net2.createPortInst("Q", ff);
 
         Path tempFile = path.resolve("test.edf");
         origNetlist.exportEDIF(tempFile);
--- a/test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
+++ b/test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
@@ -475,7 +475,7 @@ class TestEDIFNetlist {
         net1.createPortInst(port1, 1);
         net1.createPortInst("R", ff);
         
-        DIFNet net2 = top.createNet("net2");
+        EDIFNet net2 = top.createNet("net2");
         net2.createPortInst(port2);
         net2.createPortInst("Q", ff);
 
--- a/test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
+++ b/test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
@@ -451,6 +451,38 @@ class TestEDIFNetlist {
     }
 
     @Test
+    public void testBussedPortReNamingCollision(@TempDir Path path) {
+        final EDIFNetlist origNetlist = EDIFTools.createNewNetlist("test");
+
+        EDIFCell top = origNetlist.getTopCell();
+
+        EDIFCellInst ff = top.createChildCellInst("ff", Design.getPrimitivesLibrary().getCell("FDRE"));
+        origNetlist.getHDIPrimitivesLibrary().addCell(ff.getCellType());
+
+        String portName = "unfortunate_name";
+
+        // Create two ports, one single-bit and another bussed with the same root name
+        EDIFPort port0 = top.createPort(portName + "[2:0]", EDIFDirection.INOUT, 3);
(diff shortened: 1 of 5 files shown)


test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
```diff
@@ -450,6 +450,38 @@ public void testBussedPortNamingCollision(@TempDir Path path) {
...
+        final EDIFNetlist origNetlist = EDIFTools.createNewNetlist("test");
+
+        EDIFCell top = origNetlist.getTopCell();
+
+        EDIFCellInst ff = top.createChildCellInst("ff", Design.getPrimitivesLibrary().getCell("FDRE"));
+        origNetlist.getHDIPrimitivesLibrary().addCell(ff.getCellType());
+
+        String portName = "unfortunate_name";
+
+        // Create two ports, one single-bit and another bussed with the same root name
```
eddieh-xlnx, line 464: Is this comment correct? Looks like two multi-bit ports?

eddieh-xlnx (commented):
So it looks like the common (non-collision) path now has an extra:
1. `EDIFPort.getBusName(/*keepOpenBracket*/ false)` (instead of `true`) which involves constructing a new sub-`String` that strips any `[`: https://github.com/Xilinx/RapidWright/blob/8797324c61865bbf4ad3f35df474c32e88d95901/src/com/xilinx/rapidwright/edif/EDIFPort.java#L157-L159
2. An extra `EDIFCell.getPort()` hash lookup (or two; see below)

I can see that `EDIFCell.getPort()` also has some logic with `[`: https://github.com/Xilinx/RapidWright/blob/8797324c61865bbf4ad3f35df474c32e88d95901/src/com/xilinx/rapidwright/edif/EDIFCell.java#L280-L288
Am I right in interpreting that 1. will strip the `[` suffix, and then 2. will try and lookup with this (`unfortunate_name` in this test, which will fail) and then it will do a second lookup (with a new string) back on `unfortunate_name[` and find a hit)?

test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
clavin-xlnx: Thanks, updated.

test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
```diff
@@ -450,6 +450,38 @@ public void testBussedPortNamingCollision(@TempDir Path path) {
...
+        EDIFPort port0 = top.createPort(portName + "[2:0]", EDIFDirection.INOUT, 3);
+        EDIFPort port1 = top.createPort(portName + "_[1:0]", EDIFDirection.INOUT, 2);
+
+        EDIFNet net0 = top.createNet("net0");
+        net0.createPortInst(port0, 0);
+        net0.createPortInst("D", ff);
+
+        EDIFNet net1 = top.createNet("net1");
+        net1.createPortInst(port1, 1);
+        net1.createPortInst("R", ff);
```
eddieh-xlnx, line 474: Just to be extra pedantic (it seems to work!)

```suggestion
        // Create two bussed ports with a common root name, one having a trailing '_' underscore
        EDIFPort port0 = top.createPort(portName + "[2:0]", EDIFDirection.INOUT, 3);
        EDIFPort port1 = top.createPort(portName + "_[1:0]", EDIFDirection.INOUT, 2);
        // And for good measure, throw in a single-bit port too with the same root
        EDIFPort port2 = top.createPort(portName, EDIFDirection.INOUT, 1);

        EDIFNet net0 = top.createNet("net0");
        net0.createPortInst(port0, 0);
        net0.createPortInst("D", ff);

        EDIFNet net1 = top.createNet("net1");
        net1.createPortInst(port1, 1);
        net1.createPortInst("R", ff);
        
        DIFNet net2 = top.createNet("net2");
        net2.createPortInst(port2);
        net2.createPortInst("Q", ff);
```

This is the relevant part of the resulting EDIF:
```
  (Library work
    (edifLevel 0)
    (technology (numberDefinition ))
   (cell test (celltype GENERIC)
     (view netlist (viewtype NETLIST)
       (interface  
        (port (array (rename unfortunate_name_BUS_ "unfortunate_name[2:0]") 3) (direction INOUT))
        (port unfortunate_name (direction INOUT))
        (port (array (rename unfortunate_name_ "unfortunate_name_[1:0]") 2) (direction INOUT))
       )
       (contents
         (instance ff (viewref netlist (cellref FDRE (libraryref hdi_primitives))))
         (net net0 (joined
          (portref D (instanceref ff))
          (portref (member unfortunate_name_BUS_ 0))
          )
         )
         (net net2 (joined
          (portref Q (instanceref ff))
          (portref unfortunate_name)
          )
         )
         (net net1 (joined
          (portref R (instanceref ff))
          (portref (member unfortunate_name_ 1))
          )
         )
       )
     )
   )
  )
```

I think that looks okay -- it even gets accepted by Vivado:
image

clavin-xlnx:
> Am I right in interpreting that 1. will strip the `[` suffix, and then 2. will try and lookup with this (`unfortunate_name` in this test, which will fail) and then it will do a second lookup (with a new string) back on `unfortunate_name[` and find a hit)?

That is correct.  This is the mechanics used to avoid the root bus name collision problem with single bit ports of the same root name.  

To be clear, this PR only modifies how the legalization of the EDIF name on export is handled.  The internal mechanics/storage/retrieval remain the same.

test/src/com/xilinx/rapidwright/edif/TestEDIFNetlist.java
eddieh-xlnx: And just to satisfy my curiosity, here's what Vivado itself outputs:
```
       (interface
        (port unfortunate_name (direction INOUT))
        (port (array (rename unfortunate_name_HDI_0 "unfortunate_name[2:0]") 3) (direction INOUT))
        (port (array (rename unfortunate_name_ "unfortunate_name_[1:0]") 2) (direction INOUT))
       )
```

eddieh-xlnx:
> That is correct. This is the mechanics used to avoid the root bus name collision problem with single bit ports of the same root name.
> 
> To be clear, this PR only modifies how the legalization of the EDIF name on export is handled. The internal mechanics/storage/retrieval remain the same.

My analysis here was to see if we can explain where the 5% performance impact comes from. Could this be reduced now with your latest changes?

eddieh-xlnx (approved):

PR #1070, opened 2024-09-25
[RWRoute] Further cleanup

eddieh-xlnx:
- `RWRoute` to not check for sink routability anymore
- `PartialRouter` to only check sink routability after all nodes have been preserved
- `RWRoute.handleUnroutableConnection()` to abandon connections that are guaranteed to be unroutable since the bounding box is never enlarged (nor any nets be unpreserved, in the case of `PartialRouter`)
- `RWRoute` to not pre-emptively set up the second source; do so only when needed in `RWRoute.swapOutputPin()` or when recovering partially routed nets in `PartialRouter`
- Various cleanups/optimizations

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/Connection.java
+++ b/src/com/xilinx/rapidwright/rwroute/Connection.java
@@ -272,7 +272,6 @@ public class Connection implements Comparable<Connection>{
 
     public void resetRoute() {
         getRnodes().clear();
-        sink.setRouted(false);
     }
 
     public RouteNode getSourceRnode() {
--- a/src/com/xilinx/rapidwright/rwroute/Connection.java
+++ b/src/com/xilinx/rapidwright/rwroute/Connection.java
@@ -34,6 +34,7 @@ import com.xilinx.rapidwright.design.SitePinInst;
 import com.xilinx.rapidwright.device.Node;
 import com.xilinx.rapidwright.timing.TimingEdge;
 import com.xilinx.rapidwright.timing.delayestimator.DelayEstimatorBase;
+import com.xilinx.rapidwright.util.Pair;
 
 /**
  * A Connection instance represents a pair of source-sink {@link SitePinInst} instances of a {@link Net} instance.
@@ -488,5 +489,26 @@ public class Connection implements Comparable<Connection>{
                 }
             }
         }
+    }
+
+    protected Pair<SitePinInst,RouteNode> getOrCreateAlternateSource(RouteNodeGraph routingGraph) {
+        SitePinInst altSource = netWrapper.getOrCreateAlternateSource(routingGraph);
+        if (altSource == null) {
+            return null;
+        }
+
+        Net net = netWrapper.getNet();
+        RouteNode altSourceRnode;
+        if (source.equals(net.getSource())) {
+            altSourceRnode = netWrapper.getAltSourceRnode();
+        } else {
+            assert(source.equals(net.getAlternateSource()));
+            altSource = net.getSource();
+            assert(altSource != null);
+            altSourceRnode = netWrapper.getSourceRnode();
+        }
+
+        assert(altSourceRnode != null);
+        return new Pair<>(altSource, altSourceRnode);
     }
 }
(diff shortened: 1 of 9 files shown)


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -549,38 +549,44 @@ protected NetWrapper createNetWrapperAndConnections(Net net) {
...
-        }
-
         RouteNode sourceINTRnode = null;
-        RouteNode altSourceINTRnode = null;
         int indirect = 0;
         for (SitePinInst sink : sinkPins) {
             Connection connection = new Connection(numConnectionsToRoute++, source, sink, netWrapper);
             List<Node> nodes = RouterHelper.projectInputPinToINTNode(sink);
+            if (sourceINTNode == null && !nodes.isEmpty()) {
+                // Sink can be projected to an INT tile, but source cannot be; try alternate source
```
eddieh-xlnx, line 558:
```suggestion
                // Sink can be projected to an INT tile, but primary source (e.g. COUT)
                // cannot be; try alternate source
```

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -549,38 +549,44 @@ protected NetWrapper createNetWrapperAndConnections(Net net) {
...
+                    assert(sourceINTNode != null);
+                    if (sourceINTRnode == null) {
+                        sourceINTRnode = routingGraph.getOrCreate(sourceINTNode, RouteNodeType.PINFEED_O);
+                        // Where only a single primary source exists, always preserve
+                        // its projected-to-INT source node, since it could
+                        // be a projection from LAGUNA/RXQ* -> RXD* (node for INT/LOGIC_OUTS_*)
+                        routingGraph.preserve(sourceINTNode, net);
+                        netWrapper.setSourceRnode(sourceINTRnode);
+                    }
+                    assert(sourceINTRnode != null);
```
eddieh-xlnx, line 582:
```suggestion
                        assert(sourceINTRnode != null);
                        routingGraph.preserve(sourceINTNode, net);
                        netWrapper.setSourceRnode(sourceINTRnode);
                    }
```

src/com/xilinx/rapidwright/rwroute/PartialRouter.java
```diff
@@ -240,83 +240,26 @@ protected void routeGlobalClkNets() {
...
-            // Use the prev pointers to attempt to recover routing for all indirect connections
-            for (Connection connection : netWrapper.getConnections()) {
-                if (connection.isDirect()) {
-                    continue;
-                }
-                RouteNode sourceRnode = connection.getSourceRnode();
-                RouteNode sinkRnode = connection.getSinkRnode();
-                assert(sourceRnode.getType() == RouteNodeType.PINFEED_O);
+        // With all routingGraph.preserveAsync() calls having completed,
+        // now check sink routability
```
clavin-xlnx, line 244: What does it mean to "check sink routability"?  Are we making sure there are no obvious blocking nodes to prevent node expansion out of the gate?

clavin-xlnx (approved):

src/com/xilinx/rapidwright/rwroute/Connection.java
```diff
@@ -498,4 +490,25 @@ public void setAllTargets(RWRoute.ConnectionState state) {
             }
         }
     }
+
+    protected Pair<SitePinInst,RouteNode> getOrCreateAlternateSource(RouteNodeGraph routingGraph) {
```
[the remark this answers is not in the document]
eddieh-xlnx: Let's clean up UltraScale/UltraScale+ before we consider Versal. I think there's some other RapidWright infrastructure upgrades that needs to happen before Versal's alternate sources can really be captured.

src/com/xilinx/rapidwright/rwroute/PartialRouter.java
eddieh-xlnx, line 244:
```suggestion
        // now check that no sinks are preserved by another net
        // (e.g. a pin was moved from one net to the other, but
        // its old routing was not ripped up and got preserved)
        // if so, unpreserve that blocking net
```

PR #1081, opened 2024-10-09
[RouterHelper] projectOutputPinToINTNode() to breadth-first-search

### LINKED ISSUE #558: RWRoute RouteHelper fails to find the INT tile entry node from a BITSLICE_CONTROL site pin

eddieh-xlnx:
Much like `projectInputPinToINTNode()` as this seems necessary for Versal I/Os (and even fixed #558!)

Without this, in an upcoming `PartialRouter` test with `picoblaze_2022.2.dcp`, the projected result is not the same as Vivado's solution thereby leading to an unroutable situation.

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/RouterHelper.java
+++ b/src/com/xilinx/rapidwright/rwroute/RouterHelper.java
@@ -31,6 +31,7 @@ import java.io.IOException;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
--- a/src/com/xilinx/rapidwright/rwroute/RouterHelper.java
+++ b/src/com/xilinx/rapidwright/rwroute/RouterHelper.java
@@ -144,27 +144,18 @@ public class RouterHelper {
      * @return A node that connects to an INT tile from an output pin.
      */
     public static Node projectOutputPinToINTNode(SitePinInst output) {
-        int watchdog = 5;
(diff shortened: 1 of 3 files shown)


src/com/xilinx/rapidwright/rwroute/RouterHelper.java
```diff
@@ -143,27 +144,18 @@ public static boolean isExternalConnectionToCout(SitePinInst source, SitePinInst
...
-            List<Node> downhillNodes = node.getAllDownhillNodes();
-            if (downhillNodes.isEmpty()) {
-                continue;
-            }
-
-            queue.clear();
-            for (Node downhill : downhillNodes) {
+            for (Node downhill : node.getAllDownhillNodes()) {
                 if (Utils.isInterConnect(downhill.getTile().getTileTypeEnum())) {
+                    // Return node that has at laest one downhill in the INT tile
```
eddieh-xlnx, line 158:
```suggestion
                    // Return node that has at least one downhill in the INT tile
```

PR #1084, opened 2024-10-16
[TestSite] Add xcvp1902 SLR test

eddieh-xlnx:


eddieh-xlnx:
@clavin-xlnx Looks like we can go from `Tile.getClockRegion()` (now) works but from `ClockRegion.getSLR()` does not for SLR1-3.

PR #1086, opened 2024-10-22
[YosysTools] Add synthXilinx() wrapper for Yosys

eddieh-xlnx:
Add `YosysTools` convenience class to use [Yosys](https://github.com/YosysHQ/Yosys) for synthesizing Verilog files into RapidWright `EDIFNetlist` objects. Requires that the `yosys` binary be available on `PATH`.

Currently, only a number of `synthXilinx` overloads are available, corresponding to Yosys' `synth_xilinx` command, with the following flags:
- `-family xcup` (target UltraScale+ devices)
- `-edif <file>` (write the result into an EDIF file)
- `-flatten` (flatten the design into one module to benefit from cross-module optimizations)
- `-noclkbuf -noiopad` (disable BUFG and IBUF/OBUF insertion, simulating Vivado's out-of-context mode)

### DIFF
--- a/src/com/xilinx/rapidwright/edif/YosysTools.java
+++ b/src/com/xilinx/rapidwright/edif/YosysTools.java
@@ -124,4 +124,12 @@ public class YosysTools {
         FileTools.deleteFolder(workDir.toString());
         return netlist;
     }
+
+    /**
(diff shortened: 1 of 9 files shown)


eddieh-xlnx:
`TestYosysTools` passes with vanilla Yosys-0.46.

clavin-xlnx (approved):

PR #1089, opened 2024-10-30
[NetTools] Add getNodeTrees() method and NodeTree class

eddieh-xlnx:
```
     /** Compute the node routing tree of the given Net by examining its PIPs.
       * Note that this method: (a) assumes that no loops are present, (b) only discovers subtrees that start at an
       * output SitePinInst or a node tied to VCC/GND (i.e. gaps and islands will be ignored).
```

The `NodeTree` class extends the existing `Node` class by adding a new `fanouts` field.
`NodeTree.toString()` emulates Vivado's `report_route_status -of [get_nets ...]` output.

`DesignTools.updatePinsIsRouted()` now leverages this method.

### DIFF
--- a/src/com/xilinx/rapidwright/design/NetTools.java
+++ b/src/com/xilinx/rapidwright/design/NetTools.java
@@ -22,9 +22,16 @@
 
 package com.xilinx.rapidwright.design;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
+import com.xilinx.rapidwright.device.Node;
+import com.xilinx.rapidwright.device.PIP;
 import com.xilinx.rapidwright.device.SiteTypeEnum;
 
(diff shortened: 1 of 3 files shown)


src/com/xilinx/rapidwright/design/NetTools.java
```diff
@@ -46,4 +53,95 @@ public static boolean isGlobalClock(Net net) {
         
         return clkSrcSiteTypeEnums.contains(srcSpi.getSiteTypeEnum());
     }
+
+    public static class NodeTree extends Node {
```
clavin-xlnx, line 57: I think I can be persuaded to allow inner classes, but it seems like `NodeTree` has no relationship to `NetTools`, other than we are trying to hide it and `DesignTools` is too large.  Any reservations to making a regular class?

clavin-xlnx: Nevermind, I see there is a method at the bottom, I missed the indent.

clavin-xlnx (approved):

PR #1096, opened 2024-11-06
Tile.getWireConnections() not consistent with Vivado

eddieh-xlnx:
Javadoc for `Tile.getWireConnections()` states that it should be equivalent to:
```
get_nodes -downhill -of_objects [get_nodes -of [get_wires -of $tile/$wireName]]]
```

Applying it in the following way on a `xcvp1002` in Vivado:
```
get_nodes -downhill -of [get_nodes -of [get_wires CLK_REBUF_VERT_VNOC_BAO_TILE_X30Y471/IF_WRAP_CLK_V_BOT_CLK_VDISTR21]]
WARNING: [Vivado 12-2683] No nodes matched 'get_nodes -downhill -of [get_nodes -of [get_wires CLK_REBUF_VERT_VNOC_BAO_TILE_X30Y471/IF_WRAP_CLK_V_BOT_CLK_VDISTR21]]'
```
returns nothing.

But this test demonstrates that RapidWright returns:
```
Expected :[]
Actual   :[CLK_REBUF_VERT_VNOC_CBO_TILE_X30Y375/IF_WRAP_CLK_V_TOP_CLK_VDISTR21, INVALID1_X30Y431/null]
```

1. `CLK_REBUF_VERT_VNOC_CBO_TILE_X30Y375/IF_WRAP_CLK_V_TOP_CLK_VDISTR21` seems to be a wire on the original input node:
  ```
  get_nodes -of [get_wires CLK_REBUF_VERT_VNOC_CBO_TILE_X30Y375/IF_WRAP_CLK_V_TOP_CLK_VDISTR21]
  CLK_REBUF_VERT_VNOC_BAO_TILE_X30Y471/IF_WRAP_CLK_V_BOT_CLK_VDISTR21
  ```
2. Obviously returning `INVALID1_X30Y431/null` is not desirable.

PR #1102, opened 2024-11-13
[RWRoute] Clock router for Versal architecture

WenhaoLin-AMD:
The clock router for Versal architecture is based on the symmetric clock routing flow for UltraScale/UltraScale+. The main differences between them are listed below:

UltraScale/UltraScale+ | Versal
-- | --
First compute the `centroid` clock region (CR) of all target CRs, then route the clock buffer to an HROUTE node in `centroid`. | Vivado routes the signal from the clock buffer to a VROUTE node in a nearby CR, and we follow this behavior.
Use two VROUTE nodes: `vrouteUp` (for the target CRs above the `centroid` ) and `vrouteDown` (for the target CRs below the `centroid`) | Only use one VROUTE node. It uses `VDISTR_LVL{1,2}` nodes to go up and down.
First map the sink pins to leaf clock buffers (LCBs) with a simple heuristic, then route LCBs to sinks | Directly route the sinks to LCBs.
Operate on `Wire` objects, using `RouteNode.getWireConnections()` to search downhill | Operate on `Node` objects, using `Node.getAllDownhillNodes()` (since `RouteNode.getWireConnections()` relies on `tile.getWireConnections()` which doesn't work correctly on Versal devices, see https://github.com/Xilinx/RapidWright/pull/1096). Specifically, we replace `RouteNode` with `RouterHelper.NodeWithPrev`.


src/com/xilinx/rapidwright/router/VersalClockRouting.java
```diff
@@ -0,0 +1,782 @@
...
+                if (!allowedIntentCodes.contains(downhill.getIntentCode())) {
+                    continue;
+                }
+                if (!findCentroidHroute && downhill.getIntentCode() == IntentCode.NODE_GLOBAL_HROUTE_HSR) {
+                    continue;
+                }
+                if (visited.contains(downhill)) continue;
+                RouteNode rn = new RouteNode(downhill.getTile(), downhill.getWireIndex(), curr, curr.getLevel()+1);
+                
+                // The clockRegion.getApproximateCenter() may return an INVALID_* tile with huge coordinates.
```
eddieh-xlnx, line 151: Can you provide some more information (better yet, create a new PR with a test)

src/com/xilinx/rapidwright/router/VersalClockRouting.java
```diff
@@ -0,0 +1,613 @@
...
+     * @param clk The clock net to be routed.
+     * @param vroute The node to start the route.
+     * @param clockRegions Target clock regions.
+     * @param down To indicate if it is routing to the group of top clock regions.
+     * @return A list of RouteNodes indicating the reached horizontal distribution lines.
+     */
+    public static Map<ClockRegion, RouteNode> routeToHorizontalDistributionLines(Net clk,
+                                                                     RouteNode vroute,
+                                                                     Collection<ClockRegion> clockRegions,
+                                                                     boolean down,
```
eddieh-xlnx, line 514: This `down` argument is never used?

src/com/xilinx/rapidwright/router/VersalClockRouting.java
```diff
@@ -84,10 +84,9 @@ public static RouteNode routeBUFGToNearestRoutingTrack(Net clk) {
      * @param clk The current clock net to contribute routing.
      * @param startingRouteNode The intermediate start point of the clock route.
      * @param clockRegion The center clock region or the clock region that is one row above or below the center.
-     * @param adjusted A flag to guard the default functionality when routing to centroid clock region.
      * @param findCentroidHroute The flag to indicate the returned RouteNode should be HROUTE in the center or VROUTE going up or down.
      */
-    public static RouteNode routeToCentroid(Net clk, RouteNode startingRouteNode, ClockRegion clockRegion, boolean adjusted, boolean findCentroidHroute) {
+    public static RouteNode routeToCentroid(Net clk, RouteNode startingRouteNode, ClockRegion clockRegion, boolean findCentroidHroute) {
```
eddieh-xlnx, line 89: `adjusted` parameter is always true -- I've constant-propagated it. Can you see a problem with that?

src/com/xilinx/rapidwright/router/VersalClockRouting.java
WenhaoLin-AMD: Yes. It's a parameter in UltraScale clock routing method.
In UltraScale symmetric clock routing, we use `vrouteUp` and `vrouteDown` to go up and down, and `down` is to distinguish these two nodes and compute the offset of Y-coordinate of `currCR` when we are using `vrouteDown`. 

However, I found that there is only one VROUTE node in vivado solutions for Versal designs, so there is no need to use `down` here.

WenhaoLin-AMD: In Versal clock routing, we use `VDISTR_LVL{1, 2}` to go up and down.

src/com/xilinx/rapidwright/router/VersalClockRouting.java
WenhaoLin-AMD: Yes! I just copied it from `UltraScaleClockRouting` class when I started out, and I never set it to `false` in any calls in Versal clock routing.

Actually, you can see that `UltraScaleClockRouting` does not use a false value in any calls in symmetric routing too. It provides an overload method here https://github.com/Xilinx/RapidWright/blob/c43c9fe7e5882d1d7b49a5fb1d477cd50799af2a/src/com/xilinx/rapidwright/router/UltraScaleClockRouting.java#L91 which sets `adjusted` to false, and it is only called here: https://github.com/Xilinx/RapidWright/blob/c43c9fe7e5882d1d7b49a5fb1d477cd50799af2a/src/com/xilinx/rapidwright/examples/SLRCrosserGenerator.java#L371 for cross SLR clock routing.

src/com/xilinx/rapidwright/router/VersalClockRouting.java
WenhaoLin-AMD: Sorry that I wrongly compared their coordinates with the number of rows and columns of xcvu3p. I made a test for  clockRegion.getApproximateCenter() and the coordinates of returned tiles works correctly.

eddieh-xlnx: Then is this comment outdated and can be removed?

src/com/xilinx/rapidwright/router/VersalClockRouting.java
eddieh-xlnx: > However, I found that there is only one VROUTE node in vivado solutions for Versal designs, so there is no need to use down here.

Can you elaborate on what you mean by only one `VROUTE` node?

> In Versal clock routing, we use VDISTR_LVL{1, 2} nodes to go up and down.

Have you found (on the `xcvp1002`) that you only ever need to use `VDISTR` to go up?

src/com/xilinx/rapidwright/router/VersalClockRouting.java
eddieh-xlnx: > Actually, you can see that UltraScaleClockRouting does not use a false value in any calls in symmetric routing too.

That statement is not quite accurate; I would say: `SLRCrosserGenerator` calls the `routeToCentroid(Net, RouteNode, ClockRegion)` overload, which in turn calls `routeToCentroid(Net, RouteNode, ClockRegion, boolean, boolean)` where the first boolean (`adjusted`) is set to false. So yes, in `UltraScaleClockRouting`, that `adjusted` parameter is important.

src/com/xilinx/rapidwright/router/VersalClockRouting.java
WenhaoLin-AMD: > Then is this comment outdated and can be removed?

Yes

src/com/xilinx/rapidwright/router/VersalClockRouting.java
WenhaoLin-AMD: > Can you elaborate on what you mean by only one VROUTE node?

https://github.com/Xilinx/RapidWright/blob/c43c9fe7e5882d1d7b49a5fb1d477cd50799af2a/src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java#L281

Here we only have one VROUTE node for all sink pins. However, on US/US+ devices we have https://github.com/Xilinx/RapidWright/blob/c43c9fe7e5882d1d7b49a5fb1d477cd50799af2a/src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java#L208-L224 here we use two VROUTE nodes `vrouteUp` and `vrouteDown`.

src/com/xilinx/rapidwright/router/VersalClockRouting.java
eddieh-xlnx: Should we go back to using `getApproximateCenter()` then?

eddieh-xlnx: (I went ahead and changed it, and it seems to continue to work)

src/com/xilinx/rapidwright/router/VersalClockRouting.java
```diff
@@ -0,0 +1,494 @@
...
+        clk.getPIPs().addAll(allPIPs);
+        return crToVdist;
+    }
+
+    /**
+     * Routes from a vertical distribution centroid to destination horizontal distribution lines
+     * in the clock regions provided.
+     * @param clk The current clock net
+     * @param crMap A map of target clock regions and their respective vertical distribution lines
+     * @return The List of nodes from the centroid to the horizontal distribution line.
```
eddieh-xlnx, line 244: @WenhaoLin-AMD Can you update this javadoc line please?

src/com/xilinx/rapidwright/router/VersalClockRouting.java
```diff
@@ -0,0 +1,494 @@
...
+        clk.getPIPs().addAll(allPIPs);
+    }
+
+    /**
+     * Routes from a GLOBAL_VERTICAL_ROUTE to horizontal distribution lines.
+     * @param clk The clock net to be routed.
+     * @param vroute The node to start the route.
+     * @param clockRegions Target clock regions.
+     * @param down To indicate if it is routing to the group of top clock regions.
+     * @return A list of RouteNodes indicating the reached horizontal distribution lines.
```
eddieh-xlnx, line 380: @WenhaoLin-AMD Can you update this javadoc line please?

src/com/xilinx/rapidwright/router/VersalClockRouting.java
WenhaoLin-AMD: > Have you found (on the xcvp1002) that you only ever need to use VDISTR to go up?

BUFG_FABRIC sources are in the middle of the xcvp1002 so we may still need VDISTR to go down.

src/com/xilinx/rapidwright/router/VersalClockRouting.java
WenhaoLin-AMD: Have updated this part in https://github.com/Xilinx/RapidWright/pull/1102/commits/823f291b38eb59d5c650690939746d436cc1f290.

src/com/xilinx/rapidwright/router/VersalClockRouting.java
WenhaoLin-AMD: Have updated this part in https://github.com/Xilinx/RapidWright/pull/1102/commits/823f291b38eb59d5c650690939746d436cc1f290.

I don't know whether we should remove `down`. If we need to merge this method with the one in `UltraScaleClockRouting`, it seems better to keep this variable. Otherwise, we should also remove related javadoc.

eddieh-xlnx: Thanks. I no longer to plan merging `VersalClockRouting` back into `UltraScaleClockRouting` since they are sufficiently different in being `Node` and `Wire` based respectively.

src/com/xilinx/rapidwright/router/VersalClockRouting.java
```diff
@@ -0,0 +1,495 @@
...
+                        continue;
+                    }
+                    if (!visited.add(downhill)) {
+                        continue;
+                    }
+                    int cost = downhill.getTile().getManhattanDistance(lcbTile);
+                    q.add(new NodeWithPrevAndCost(downhill, curr, cost));
+                }
+            }
+            throw new RuntimeException("ERROR: Couldn't route to distribution line in clock region " + lcb);
```
clavin-xlnx, line 370: `lcb` is a node, so maybe reword this message?

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -190,39 +193,93 @@ private static Map<String, List<Node>> getListOfNodesFromRoutes(Device device, M
...
-
-        RouteNode vrouteUp = null;
-        RouteNode vrouteDown;
-        // Two VROUTEs going up and down
-        ClockRegion aboveCentroid = upClockRegions.isEmpty() ? null : centroid.getNeighborClockRegion(1, 0);
-        if (aboveCentroid != null) {
-            vrouteUp = UltraScaleClockRouting.routeToCentroid(clk, centroidHRouteNode, aboveCentroid, true, false);
-        }
-        vrouteDown = UltraScaleClockRouting.routeToCentroid(clk, centroidHRouteNode, centroid.getNeighborClockRegion(0, 0), true, false);
+        if (device.getSeries() == Series.UltraScale || device.getSeries() == Series.UltraScalePlus) {
```
clavin-xlnx, line 196: Would it make sense to refactor the bodies of the if/else if into their own method inside `UltraScaleClockRouting.java` and `VersalClockRouting.java`?

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -190,39 +193,93 @@ private static Map<String, List<Node>> getListOfNodesFromRoutes(Device device, M
...
+
+            UltraScaleClockRouting.routeLCBsToSinks(clk, lcbMappings, getNodeStatus);
+        } else if (device.getSeries() == Series.Versal) {
+            // Clock routing on Versal devices
+
+            List<ClockRegion> clockRegions = getClockRegionsOfNet(clk);
+            SitePinInst source = clk.getSource();
+            SiteTypeEnum sourceTypeEnum = source.getSiteTypeEnum();
+            // In US/US+ clock routing, we use two VROUTE nodes to reach the clock regions above and below the centroid.
+            // However, we can see that Vivado only use one VROUTE node in the centroid clock region for Versal clock routing,
```
clavin-xlnx, line 237:
```suggestion
            // However, we can see that Vivado only uses one VROUTE node in the centroid clock region for Versal clock routing,
```

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -190,39 +193,93 @@ private static Map<String, List<Node>> getListOfNodesFromRoutes(Device device, M
...
+            Node centroidHRouteNode;
+
+            if (sourceTypeEnum == SiteTypeEnum.BUFG_FABRIC) {
+                // These source sites are located in the middle of the device. The path from the output pin to VROUTE matches the following pattern:
+                // NODE_GLOBAL_BUFG (the output node with a suffix "_O") -> 
+                // NODE_GLOBAL_BUFG (has a suffix "_O_PIN") -> 
+                // NODE_GLOBAL_GCLK ->
+                // NODE_GLOBAL_VROUTE (located in the same clock region of the source site)
+
+                // Notice that Vivado always use the above VROUTE node, there is no need to find a centroid clock region to route to.
```
clavin-xlnx, line 250:
```suggestion
                // Notice that Vivado always uses the above VROUTE node, there is no need to find a centroid clock region to route to.
```

clavin-xlnx (approved):

PR #1059, opened 2024-09-16
[RWRoute] Cleanup static router and RouterHelper

eddieh-xlnx:
* Improve static router performance
  * Check for used/VCC/GND wires on push, and check for static sources on pop
  * Improve re-use inside static router (sort sinks, consider routing bounce/bypass site pins, etc.)
  * Net result is reduced number of static sources used, smaller decrease in PIPs
* Add `RouterHelper.NodeWithPrev` inner class, which extends the `Node` class
* Refactor and cleanup code

PR #1105, opened 2024-11-14
[GlobalSignalRouting] routeStaticNets() to take a list of static pins

eddieh-xlnx:
Checking that all such pins are available (and removing those that are not available, such as those occurring in encrypted designs where it's not possible to infer all pins on signal nets)

To achieve this, `addStaticNetRoutingTargets()` now preserves static nets' PIPs and *output* pins only; preservation of input pins is deferred to `routeStaticNets()` (at which point all signal nets' routing will have been preserved) --- which can move or remove input pins between/from static nets --- before doing so.

Also remove a GlobalSignalRouting optimization added in https://github.com/Xilinx/RapidWright/pull/1059 for static nets that encourages reuse -- that optimization didn't kick in anyway, since static sinks were not always preserved, to ensure tests continue to pass.

### DIFF
--- a/src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
+++ b/src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
@@ -337,11 +337,9 @@ public class GlobalSignalRouting {
      * @param design The {@link Design} instance to use.
      * @param routeThruHelper The {@link RouteThruHelper} instance to use.
      */
-    public static void routeStaticNet(Net currNet,
+    public static void routeStaticNet(List<SitePinInst> pins,
                                       Function<Node,NodeStatus> getNodeState,
                                       Design design, RouteThruHelper routeThruHelper) {
-        NetType netType = currNet.getType();
-        Set<PIP> netPIPs = new HashSet<>(currNet.getPIPs());
         Queue<Node> q = new ArrayDeque<>();
         Set<Node> usedRoutingNodes = new HashSet<>();
         Map<Node, Node> prevNode = new HashMap<>();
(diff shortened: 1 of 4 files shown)


src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -525,10 +525,16 @@ protected void routeStaticNets() {
             Net staticNet = e.getKey();
             List<SitePinInst> pins = e.getValue();
 
+            // For some encrypted designs, it's possible that RapidWright cannot infer all SitePinInst-s leading to
+            // some site pins (e.g. CKEN) to default to a static net. Detect such cases -- when signal nets are
```
eddieh-xlnx, line 529:
```suggestion
            // some site pins (e.g. CKEN) defaulting those to static nets. Detect such cases -- when signal nets are
```

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -332,21 +332,19 @@ private static ClockRegion findCentroid(Net clk, Device device) {
 
     /**
      * Routes a static net (GND or VCC).
-     * @param currNet The current static net to be routed.
+     * @param pins A list of static pins to be routed.
```
clavin-xlnx, line 335: Are we assuming all the pins are on the same static net (all VCC or all GND)?  If so, perhaps indicate it in the Javadoc.

clavin-xlnx (approved):

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -533,28 +530,57 @@ protected void addStaticNetRoutingTargets(Net staticNet) {
...
+                });
+                // Remove from map if empty
+                return pins.isEmpty() ? null : pins;
+            });
+            preserveNet(staticNet, false);
+        }
+
+        if (noStaticRouting) {
+            return;
+        }
```
eddieh-xlnx, line 574:
```suggestion
            // Preserve this static net regardless of any routing necessary
            preserveNet(staticNet, false);
        }

        if (noStaticRouting) {
            // Now that all static nets have been preserved, return if no work to be done
            return;
        }
```

PR #1110, opened 2024-11-20
Changes to support new array-based sitewire-net changes 

clavin-xlnx:
This update tests to reflect bug fixes and minor behavior changes associated with a new storage method in `SiteInst` objects where instead of using maps, the `Net` to site wire now uses an array.  This led to a change in iteration order and also a bug was discovered when `SiteInst` are renamed, they were not getting updated in the `Set<SiteInst>` class members of `Net` objects.  This led to some failures to fully remove the references when prompted in some of the tests.

### DIFF
--- a/test/src/com/xilinx/rapidwright/design/compare/TestDesignComparator.java
+++ b/test/src/com/xilinx/rapidwright/design/compare/TestDesignComparator.java
@@ -149,7 +149,7 @@ public class TestDesignComparator {
         Assertions.assertTrue(net.hasPIPs());
         test2.removeNet(net);
 
-        compareDesign(22, 1, DesignDiffType.NET_MISSING, dc, gold, test2);
+        compareDesign(28, 1, DesignDiffType.NET_MISSING, dc, gold, test2);
 
         Net extra = new Net("extraNet");
         for (PIP p : net.getPIPs()) {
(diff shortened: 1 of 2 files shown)


test/src/com/xilinx/rapidwright/design/compare/TestDesignComparator.java
```diff
@@ -149,25 +149,25 @@ public void testDesignComparator() {
         Assertions.assertTrue(net.hasPIPs());
         test2.removeNet(net);
 
-        compareDesign(22, 1, DesignDiffType.NET_MISSING, dc, gold, test2);
+        compareDesign(28, 1, DesignDiffType.NET_MISSING, dc, gold, test2);
```
eddieh-xlnx, line 152: I don't think there's any `SiteInst` renaming happening in this test, so are the increased number of diffs due to iteration order (which shouldn't matter?)

clavin-xlnx: This changed because the `SiteInst` renaming was taking placed at the end of the DCP load when `ModuleInst` objects were being formed (RapidWright has added special metadata to reconstruct them after a DCP has traditionally been fully loaded).  Because of the rename taking place, one of the sites was not being considered previously.  The new code now handles the `SiteInst` that was being left out before the change.

eddieh-xlnx (approved):

PR #1093, opened 2024-11-04
[RWRoute] Versal optimizations

WenhaoLin-AMD:
This PR:
- Applies https://github.com/Xilinx/RapidWright/pull/1095 to Versal nodes based on their intent codes -- ensuring that NON_LOCAL -> LOCAL node transition only happens in the vicinity of the sink tile
  - With special handling for `NODE_(CLE|INTF)_[BC]NODE` nodes to prevent them from being used to travel between `INT` tiles, but still used to "bounce" (back) into the original (sink) tile
- Further applies https://github.com/Xilinx/RapidWright/pull/1098 to Versal nodes based on their wire or tile name to ensure that only the relevant east/west side of the INT tile is explored
  - Note that `NODE_CLE_[BC]NODE` nodes can reach `NODE_INODE`s on the opposite side of the former's wire name
- Add a `RouteNodeType.LOCAL_RESERVED` as part of a Versal-only *heuristic* to reserve `[BC]NODE`s uphill of a `NODE_(CLE|INTF)_CTRL` sink, since `[BC]NODE`s (especially CNODEs) are the only way to reach CTRL sinks, while also providing extra connectivity to `NODE_INODE`s in the `INT` tile.
  - e.g. On 1c2b471, for `finn_radioml` using `LOCAL_RESERVED` requires 119M pops over 10 iterations to route, but without needed 176M pops over 17 iterations.
  - The downside to this heuristic is that all CNODEs are reserved for even one CTRL sink, meaning that just one sink will prevent any other nets from using those CNODEs

Results:
```
			Before (a3af7dc)		After (1c2b471)	
			Nodes popped	Runtime (s)	Nodes popped	Runtime (s)
boom_med_pb		  308,212,121	 201		 92,043,436	 79
boom_soc		3,320,844,555	2184		831,951,144	698
corescore_500		  381,326,217	 249		141,432,563	133
corescore_500_pb	  393,285,727	 260		147,020,615	137
finn_radioml		  401,124,667	 253		118,592,148	101
koios_dla_like_large	1,329,175,620	 923		270,148,933	263
logicnets_jscl		   85,632,115	  58		 42,305,715	 33
vtr_lu64peeng		  429,006,483	 295		194,564,833	175
vtr_mcml		  185,015,595	 127		 79,474,273	 71
```

Overall, 2.7x improvement in wall clock RWRoute runtime, 3.6x the number of nodes popped.

PR #1112, opened 2024-11-21
Various preprocessing fixes for Versal routing

eddieh-xlnx:
A number of changes necessary to infer missing `SitePinInst`s on fully unrouted Versal designs so that they may be routed.

Also move the Versal-only CTRL sink RWRoute heuristic introduced in #1093's `determineRoutingTargets()` to `initializeRouting()` (after static routing has completed) in order to not prevent that router from completing.

Built on top of https://github.com/Xilinx/RapidWright/pull/1102


eddieh-xlnx:
Superseded by #1115

PR #1115, opened 2024-11-22
Various preprocessing fixes for Versal routing

eddieh-xlnx:
A number of changes necessary to infer missing `SitePinInst`s on fully unrouted Versal designs so that they may be routed.

Also move the Versal-only CTRL sink RWRoute heuristic introduced in #1093's `determineRoutingTargets()` to `initializeRouting()` (after static routing has completed) in order to not prevent that router from completing.


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2296,11 +2296,19 @@ public static List<String> getAllRoutedSitePinsFromPhysicalPin(Cell cell, Net ne
...
-                            String routeThru = possibleRouteThru.getPinMappingsP2L().keySet().iterator().next();
-                            queue.add(bel.getPin(routeThru));
+                        if (possibleRouteThru == null) {
+                            BELPin clkBelPin = bel.isSliceFFClkMod() ? bel.getPin("CLK") : null;
+                            if (clkBelPin != null && inst.getNetFromSiteWire(clkBelPin.getSiteWireName()) == net) {
+                                queue.add(clkBelPin);
+                            }
+                        } else {
+                            if (possibleRouteThru.isRoutethru()) {
+                                String routeThru = possibleRouteThru.getPinMappingsP2L().keySet().iterator().next();
```
clavin-xlnx, line 2309: FYI - In 2024.2 there will be a convenient API to replace this.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3244,61 +3272,74 @@ public static void createA1A6ToStaticNets(Design design) {
...
-                    } else {
-                        throw new RuntimeException("Assumption that only 5LUTs can use O5 failed here.");
-                    }
+                    assert(fiveOrSix == '5');
                 } else {
                     if (fiveOrSix != '6') {
                         // Assume that O6 is only driven by 6LUT, even though possible for 5LUT, unless
                         // it's a routethru
-                        assert (cell.isRoutethru());
+                        assert(cell.isRoutethru());
```
clavin-xlnx, line 3340: Our coding guidelines does not stipulate a space or absence thereof.  Since `assert` is a keyword rather than a method, it seems like my IDE follows the same as `while` and `for`.  Probably not worth legislating, but I think its one of those inconsistencies that has leaked through.

clavin-xlnx (approved):

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: Mine likes to add a space too, but I personally prefer not one which is the `C` way. Let's restore it to prevent a diff.

eddieh-xlnx, line 3340:
```suggestion
                        assert (cell.isRoutethru());
```

eddieh-xlnx:
Changes here will be tested in a future PR which depends on https://github.com/Xilinx/RapidWright/pull/1102 and 2024.2.0.

PR #1117, opened 2024-12-02
[NetTools] Add check for net driving clock sinks

clavin-xlnx:
This avoid looking for gap routing as a proxy check to see if a net is a clock.

### DIFF
--- a/src/com/xilinx/rapidwright/design/NetTools.java
+++ b/src/com/xilinx/rapidwright/design/NetTools.java
@@ -144,4 +144,18 @@ public class NetTools {
 
         return subtrees;
     }
+    
+    /**
+     * Checks if the provided net drives a clock site pin input.
+     * 
+     * @param net The net to examine.
+     * @return True if the net has a site pin clock input, false otherwise.
+     */
+    public static boolean hasClockSinks(Net net) {
+        for (SitePinInst sink : net.getPins()) { 
+            if (sink.isOutPin()) continue;
+            if (sink.getName().contains("CLK")) return true;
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/design/NetTools.java
```diff
@@ -144,4 +144,18 @@ public static List<NodeTree> getNodeTrees(Net net) {
...
+    /**
+     * Checks if the provided net drives a clock site pin input.
+     * 
+     * @param net The net to examine.
+     * @return True if the net has a site pin clock input, false otherwise.
+     */
+    public static boolean hasClockSinks(Net net) {
+        for (SitePinInst sink : net.getPins()) { 
+            if (sink.isOutPin()) continue;
+            if (sink.getName().contains("CLK")) return true;
```
eddieh-xlnx, line 157: Would this work?
```suggestion
            if (sink.getName().startsWith("CLK")) return true;
```

eddieh-xlnx (approved):

src/com/xilinx/rapidwright/design/NetTools.java
clavin-xlnx: Here are all the site pin inputs across all sitetypes/devices/series that contain "CLK":
[all_site_pins_contain_clk.txt](https://github.com/user-attachments/files/17984849/all_site_pins_contain_clk.txt)

I think we should probably stick with `.contains()`.

eddieh-xlnx (approved):

PR #1119, opened 2024-12-04
RWRoute preprocessing fixes

eddieh-xlnx:
1. Tie off `MMCM.CLKFBIN` and `MMCM.CLKIN2` site pins to VCC if unused.
2. `RouterHelper.invertPossibleGndPinsToVccPins()`: Unroute intra-site net when inverting (and do not reroute)
  When inverting `SitePinInst`s from being on the GND net to the VCC net, unpaint the sitewire if there's an inverter.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -3324,16 +3324,13 @@ public class DesignTools {
                     if ("SRLC32E".equals(cell.getPropertyValueString("XILINX_LEGACY_PRIM"))) {
                         expectGndNet = true;
                         staticNet = gndNet;
-                        // Expect sitewire to be VCC or GND
-                        if (!a6Net.isStaticNet()) {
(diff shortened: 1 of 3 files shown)


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3385,74 +3385,77 @@ public static void createCeSrRstPinsToVCC(Design design) {
...
+                                spi = new SitePinInst(false, sitePinName, si);
+                            }
+                            boolean updateSiteRouting = false;
+                            vccNet.addPin(spi, updateSiteRouting);
                         }
-                        boolean updateSiteRouting = false;
-                        vccNet.addPin(spi, updateSiteRouting);
+                    }
+                } else if (si.getSiteTypeEnum() == SiteTypeEnum.MMCM) {
+                    for (String sitePinName : new String[]{"CLKFBIN", "CLKIN2"}) {
```
clavin-xlnx, line 3457: The MMCM in both Versal and UltraScale+ have the same set of site pins.  Is there something we should also be doing for UltraScale+?

clavin-xlnx (approved):

Issue #1124, opened 2024-12-15
NullPointerException during stitching


ElaHobby:
Hi, 

I have an issue in the BlockStitcher and would appreciate your help. I am running rapid_compile_ipi, I generated all routed.dcp files based on the PBlock and the flow stops in the BlockStitcher as described below. 

The line which throws error (in rapidwright.ipi.BlockStitcher):  
```
    ModuleInst mi = stitched.createModuleInst(modInstName, modImpls.get(implementationIndex));
```
It starts going through modules (loop  _for (Entry<String,String> e : ipNames.entrySet())_ ) but stops at the first one and throws: 
```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "com.xilinx.rapidwright.edif.EDIFNetlist.getLibrary(String)" because the return value of "com.xilinx.rapidwright.design.Design.getNetlist()" is null
        at com.xilinx.rapidwright.design.Design.createModuleInst(Unknown Source)
        at com.xilinx.rapidwright.design.Design.createModuleInst(Unknown Source)
        at com.xilinx.rapidwright.ipi.BlockStitcher.main(BlockStitcher.java:541)
```        
The problem is: Device class is closed source, so I can not debug it to figure out what is wrong. Seems like netlist is empty? But isn't this expected due to code line before :" stitched.setNetlist(null); "? If I comment that line, it goes through a few modules without complaining (compared to above where it stops at the first one), but then it throws another error at another module later on...
````
Exception in thread "main" java.lang.RuntimeException: ERROR: Destination netlist already contains EDIFCell named 'axis_broadcaster_v1_1_32_core' in library 'work'
        at com.xilinx.rapidwright.edif.EDIFNetlist.copyCellAndSubCellsWorker(EDIFNetlist.java:665)
        at com.xilinx.rapidwright.edif.EDIFNetlist.copyCellAndSubCellsWorker(EDIFNetlist.java:656)
        at com.xilinx.rapidwright.edif.EDIFNetlist.copyCellAndSubCellsWorker(EDIFNetlist.java:656)
        at com.xilinx.rapidwright.edif.EDIFNetlist.copyCellAndSubCells(EDIFNetlist.java:626)
        at com.xilinx.rapidwright.design.Design.a(Unknown Source)
        at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1228)
        at com.xilinx.rapidwright.design.Design.addModule(Unknown Source)
[...]

PR #1134, opened 2025-01-03
[BlockStitcher] Work arounds to address stitching issues in latest Vivado 2024.2

clavin-xlnx:
This is a few fixes/work arounds to enable the `BlockStitcher` to run on 2024.2 designs from Vivado.  

1. Consolidate the final netlist set of cells to the `work` library and update all references of the work library to the canonical instance in the netlist.
2. Change the name of the design to match the incoming netlist top cell name.
3. Works around a reference issue for pass-thru port connections where the `EDIFPortInst` has a null reference to its `EDIFCellInst`.  It works around this issue by referencing the `EDIFCellInst` in the `EDIFHierCellInst` instead.
4. The BlockStitcher relies on operating on a null netlist, however, modern changes do not allow this.  A work around is added that puts a temporary 'dummy' netlist in place and then removes it.

See also #1124.

### DIFF
--- a/src/com/xilinx/rapidwright/ipi/BlockStitcher.java
+++ b/src/com/xilinx/rapidwright/ipi/BlockStitcher.java
@@ -107,16 +107,17 @@ public class BlockStitcher {
     private ArrayList<EDIFHierPortInst> getPassThruPortInsts(Port port, EDIFHierPortInst curr) {
         ArrayList<EDIFHierPortInst> list = new ArrayList<>();
         for (String name : port.getPassThruPortNames()) {
-            EDIFPortInst passThruInst = curr.getPortInst().getCellInst().getPortInst(name);
+            EDIFHierCellInst hierInst = curr.getHierarchicalInst();
+            EDIFPortInst passThruInst = hierInst.getInst().getPortInst(name);
             if (passThruInst == null) {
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/ipi/BlockStitcher.java
```diff
@@ -534,7 +535,7 @@ public static void main(String[] args) {
             }
             //System.out.println(modInstName + " " + implementationIndex);
             EDIFNetlist tmp = stitched.getNetlist();
-            stitched.setNetlist(null);
+            stitched.setNetlist(EDIFTools.createNewNetlist("dummy"));
             ModuleInst mi = stitched.createModuleInst(modInstName, modImpls.get(implementationIndex));
             stitched.setNetlist(tmp);
```
eddieh-xlnx, line 540: Are you sure we don't leak any references here? It seems that `createModuleInst` will modify (and reference) the "dummy" netlist; the design then switches back out to "tmp" immediately after creation.

clavin-xlnx: This is a work-around for this specific use case in order to not make changes to `Design.createModuleInst()`.  SInce this is a rare use case, we will let it go for now.

PR #1145, opened 2025-02-11
Routing Heat Map Example

clavin-xlnx:
A simple example application that reads in a DCP and outputs a CSV of the routing used in each INT tile of the design.  This can be imported into Excel and using the conditional coloring option by selecting all of the non-empty cells:

image

This can then generate a heat map of routing congestion:

image

To run this example from scratch:

```
git clone -b routing_heat_map --recursive https://github.com/Xilinx/RapidWright.git
cd RapidWright
./gradlew compileJava
bin/rapidwright RoutingHeatMap test/RapidWrightDCP/picoblaze_2022.2.dcp picoblaze_2022.2.csv
```
Then you can open the csv in Excel or a similar tool.


eddieh-xlnx (approved):
Neat!

PR #1147, opened 2025-02-12
[EDIF] Fix Export of Single Bit Busses

clavin-xlnx:
If a single-bit bus was declared, it was not being exported properly by the EDIF writer.  This attempts to fix this issue.

### DIFF
--- a/src/com/xilinx/rapidwright/design/RTLStubGenerator.java
+++ b/src/com/xilinx/rapidwright/design/RTLStubGenerator.java
@@ -79,7 +79,7 @@ public class RTLStubGenerator {
             for (int i=0; i < ports.length; i++) {
                 EDIFPort p = ports[i];
                 String dir = p.getDirection().name().toLowerCase();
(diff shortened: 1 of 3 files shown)


clavin-xlnx:
Unfortunately, this change has somehow triggered a sporadic, bizarre failure of the JVM to run out of memory:
```
OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00000004a0800000, 2835349504, 0) failed; error='Not enough space' (errno=12)
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 2835349504 bytes for committing reserved memory.
# An error report file with more information is saved as:
# .../hs_err_pid3700608.log
```
I obviously tried increasing the memory available to the JVM, but it fails in exactly the same way.  Somehow there is a race somewhere as the error goes away in debug mode, so running it in the debugger is not useful.

eddieh-xlnx (approved):

PR #1148, opened 2025-02-14
[EDIFNetlist] exportEDIF() to use getLibrariesInExportOrder()

eddieh-xlnx:
`EDIFNetlist.getLibrariesInExportOrder()` is called by `BinaryEDIFWriter.writeBinaryEDIF()` already.

Without this, after generating an array of particular instances the output DCP contains an incorrectly ordered EDIF that does not get accepted by Vivado.


eddieh-xlnx:
@clavin-xlnx has expressed a concern over the performance impact of adding this call (which iteratively analyzes every cell in the design) to the critical-path of exporting EDIFs. Let's collect some data.

PR #1154, opened 2025-03-19
[DesignTools] Ensure cell pin has a logical connection before attempting to unroute it

clavin-xlnx:
This is a less common case that can occur around a CARRY's inputs.  Previously, the method was mistaking the fact that a LUT output was simply driving an input pin on the CARRY that led to it being unrouted, but in reality, they are not connected.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -1556,6 +1556,13 @@ public class DesignTools {
             Net net = siteInst.getNetFromSiteWire(belPin.getSiteWireName());
             if (net == null)
                 return Collections.emptyList();
+            else {
+                // Check if net is connected to logical cell
+                EDIFPortInst portInst = cell.getEDIFCellInst().getPortInst(logicalPinName);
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -1556,6 +1556,13 @@ public static List<SitePinInst> unrouteCellPinSiteRouting(Cell cell, String logi
             Net net = siteInst.getNetFromSiteWire(belPin.getSiteWireName());
             if (net == null)
                 return Collections.emptyList();
+            else {
+                // Check if net is connected to logical cell
+                EDIFPortInst portInst = cell.getEDIFCellInst().getPortInst(logicalPinName);
```
eddieh-xlnx, line 1561: How should `cell.getEDIFCellInst() == null` be handled, as for encrypted designs? (outside of not NPE-ing)

clavin-xlnx: Thanks, I added a check for that.

eddieh-xlnx (approved):

PR #1157, opened 2025-03-25
[DesignTools] Fix routethru corner case in fullyUnplaceCellHelper()

clavin-xlnx:
Discovered an issue when removing A6LUT cell (LUT1) before CARRY8 cell caused the site routing to be left in undesirable state to fully remove the routethru in H6LUT.  
image

This change checks to make sure when traversing other potential sinks that it only continues if the routethru cell encountered is servicing the currently explored cell's pin.

### DIFF
--- a/test/src/com/xilinx/rapidwright/eco/TestECOTools.java
+++ b/test/src/com/xilinx/rapidwright/eco/TestECOTools.java
@@ -591,4 +591,38 @@ public class TestECOTools {
         Assertions.assertNull(si.getUsedSitePIP("CLKINV"));
 
     }
+
+    @Test
+    public void testRemoveCellNeighborRouteThru() {
+        // This is a corner case that involves a CARRY8 in US+ and a LUT removal from
+        // A6LUT that would prevent the routethru removal later when the CARRY8 was
+        // removed.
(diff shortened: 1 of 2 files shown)


test/src/com/xilinx/rapidwright/eco/TestECOTools.java
```diff
@@ -591,4 +591,38 @@ public void testRemoveCellVersal() {
...
+        routethruNet.connect(ff, "Q");
+        routethruNet.connect(carry, "S[7]");
+        si.routeIntraSiteNet(routethruNet, ff.getBEL().getPin("Q"), si.getBELPin("HQ", "HQ"));
+        si.routeIntraSiteNet(routethruNet, si.getBELPin("H6", "H6"), carry.getBEL().getPin("S7"));
+
+        List<EDIFHierCellInst> remove = new ArrayList<>();
+        remove.add(lut1.getEDIFHierCellInst());
+        ECOTools.removeCell(d, remove, null);
+
+        // Ensure the net stays routed after LUT1 is remove from A6LUT
```
eddieh-xlnx, line 625:
```suggestion
        // Ensure the net stays routed after LUT1 is removed from A6LUT
```

eddieh-xlnx (approved):

PR #1159, opened 2025-03-28
[ECOTools] Refactor cell from one hier to another

clavin-xlnx:
Adds functionality to move a cell from one level of hierarchy in a design to another.  If the cell is placed and/or routed, the implementation aspects will remain the same.

For example, given a cell `base_mb_i/microblaze_0/U0/MicroBlaze_Core_I/Performance.Core/Decode_I/PreFetch_Buffer_I1/Instruction_Prefetch_Mux[9].Gen_Instr_DFF/EX_Op3[2]_i_2` marked in Magenta below:
image

Refactoring the cell to a new parent hierarchy (`dbg_hub/inst`) would yield:
image

The placement and routing remain identical, only the logical netlist changes.

### DIFF
--- a/test/src/com/xilinx/rapidwright/eco/TestECOTools.java
+++ b/test/src/com/xilinx/rapidwright/eco/TestECOTools.java
@@ -671,6 +671,9 @@ public class TestECOTools {
 
     }
 
+    /**
+     * Generated with CodeGenerator.genCodeForTestSite()
+     */
     public Design genTestDesign() {
         Design design = new Design("test", "xcku060-ffva1517-2-i");
         Device device = design.getDevice();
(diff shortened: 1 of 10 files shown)


test/src/com/xilinx/rapidwright/eco/TestECOTools.java
```diff
@@ -625,4 +633,324 @@ public void testRemoveCellNeighborRouteThru() {
...
+                    continue;
+                }
+                Assertions.assertTrue(leafPins.remove(leafPin));
+            }
+            Assertions.assertEquals(0, leafPins.size());
+        }
+
+    }
+
+    public Design genTestDesign() {
```
clavin-xlnx, line 674:
```suggestion
    /**
     * Generated with CodeGenerator.genCodeForTestSite()
     */
    public Design genTestDesign() {
```

eddieh-xlnx (approved):

PR #1160, opened 2025-04-03
[CodeGenerator] Creates Java Code to Construct a Test Design With a Single Example SiteInst 

clavin-xlnx:
This PR augments the `CodeGenerator` with a new method that can take an existing `SiteInst` and re-generate a new `Design` object with the same `SiteInst` configuration with the following modifications:

- The names of cells and nets are changed such that they are simple enumerations (`net0`, `net1`, `cell2`, ...).  This allows users to generate a test of a single `SiteInst` without sharing any information related to the existing design
- All site routing, routethrus, sitePIPs, and cells in the site are preserved and reproduced.  Any nets that connect to the `SiteInst` are re-created, but are only ported up to the top level.  Multiple site instances are currently not supported.

For example, the provided test will generate the following configured `SiteInst`:
image
image

### DIFF
--- a/src/com/xilinx/rapidwright/util/CodeGenerator.java
+++ b/src/com/xilinx/rapidwright/util/CodeGenerator.java
@@ -44,7 +44,6 @@ import com.xilinx.rapidwright.device.SitePIP;
 import com.xilinx.rapidwright.device.SiteTypeEnum;
 import com.xilinx.rapidwright.edif.EDIFHierCellInst;
 import com.xilinx.rapidwright.edif.EDIFHierPortInst;
-import com.xilinx.rapidwright.edif.EDIFNetlist;
 import com.xilinx.rapidwright.edif.EDIFPortInst;
 import com.xilinx.rapidwright.edif.EDIFPropertyValue;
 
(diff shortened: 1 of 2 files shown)


test/src/com/xilinx/rapidwright/util/TestCodeGenerator.java
```diff
@@ -112,4 +131,65 @@ public void testTestNetGenerator() {
...
+    @Test
+    public void testGenCodeForTestSite(@TempDir Path tempDir) {
+        Design d = RapidWrightDCP.loadDCP("microblazeAndILA_3pblocks.dcp");
+        SiteInst si = d.getSiteInstFromSiteName("SLICE_X36Y97");
+
+        Path javaFile = tempDir.resolve(CodeGenerator.TEST_SITE_INST_CLASS_NAME + ".java");
+        try (PrintStream ps = new PrintStream(javaFile.toString())) {
+            CodeGenerator.genCodeForTestSite(si, ps, false);
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
```
eddieh-xlnx, line 144:
```suggestion
            throw new RuntimeException(e);
```

Or just let it propagate out and get caught by Junit.

src/com/xilinx/rapidwright/util/CodeGenerator.java
```diff
@@ -174,4 +187,249 @@ public static void addPIPs(Net net, String[] pips) {
...
+     * @param bel         The name of the BEL where the cell should be placed
+     * @param pinMaps     A variable number of pin mappings for the cell where each
+     *                    string is of the format <physical pin name>:<logical pin
+     *                    name>
+     * @return The created cell.
+     */
+    public static Cell genCell(SiteInst si, String name, boolean isRoutethru, String type, String bel,
+            String... pinMaps) {
+        Cell c = null;
+        if (isRoutethru || type.startsWith("<")) {
```
eddieh-xlnx, line 237: Did you find `type.startsWith("<")` to be necessary, which may imply that we aren't setting `Cell.isRoutethru()` correctly?

clavin-xlnx: We have several "special" cells that aren't routethrus but don't have logical equivalence either.  They all have a special type that begins with "<", those are ones we want to detect here.

eddieh-xlnx (approved):

PR #1162, opened 2025-04-09
Get Partial Routed Net Examples

coherent17:
This simple tool discussed in [#1152] & [#1161] to help user get the partial routed net (site pin conflict) to further debug with the route fail case from Vivado.

The # of nets reported in RapidWright is the same as the tcl command ```report_route_status``` in Vivado.

1. Fully placed & route design:
image
image
The number match in the Vivado report_route_status # of routable nets

2. Route fail design:
image
image
The code snippet in GetRouteNets.java will get the number of site pin conflict as report in # of nets with unrouted pins in Vivado report_route_status command and also show the detail info about the net and the pins.

Thanks

### DIFF
--- a/src/com/xilinx/rapidwright/MainEntrypoint.java
+++ b/src/com/xilinx/rapidwright/MainEntrypoint.java
@@ -50,6 +50,7 @@ import com.xilinx.rapidwright.edif.EDIFPropertyValue;
 import com.xilinx.rapidwright.edif.EDIFTools;
 import com.xilinx.rapidwright.examples.AddSubGenerator;
 import com.xilinx.rapidwright.examples.CopyMMCMCell;
+import com.xilinx.rapidwright.examples.CountRoutedNets;
 import com.xilinx.rapidwright.examples.CustomRouting;
 import com.xilinx.rapidwright.examples.DecomposeLUT;
 import com.xilinx.rapidwright.examples.ExampleNetlistCreation;
@@ -131,6 +132,7 @@ public class MainEntrypoint {
         addFunction("CheckAccuracyUsingGnlDesigns", CheckAccuracyUsingGnlDesigns::main);
         addFunction("CompareRouteStatusReports", CompareRouteStatusReports::main);
         addFunction("CopyMMCMCell", CopyMMCMCell::main);
+        addFunction("CountRoutedNets", CountRoutedNets::main);
         addFunction("CUFR", CUFR::main);
         addFunction("CustomRouting", CustomRouting::main);
         addFunction("DcpToInterchange", DcpToInterchange::main);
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/examples/GetRouteFailNets.java
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright (c) 2025, Advanced Micro Devices, Inc.
+ * All rights reserved.
+ *
+ * Author: Coherent Ho, Advanced Micro Devices, Inc.
```
clavin-xlnx, line 5: Could you update your affiliation?  I'm assuming you don't work at AMD.

src/com/xilinx/rapidwright/examples/GetRouteFailNets.java
```diff
@@ -0,0 +1,81 @@
...
+
+
+/**
+ * Simple tool for get the partially routed nets in a design.
+ */
+public class GetRouteFailNets {
+
+    public static void main(String[] args) {
+        if (args.length != 2) {
+            System.out.println("Usage: <input DCP> <input EDIF>");
```
clavin-xlnx, line 41: Most of the time, we name the `.edf` file with the same root file name as the `.dcp`.  In these cases, RapidWright will automatically look for the corresponding `.edf` file from the `.dcp` name.  In most applications, we make supplying the `.edf` file an option rather than a requirement (see https://github.com/Xilinx/RapidWright/blob/5bdc51c8c6e95bedb46d0d9f902b1d1d72e74ba3/src/com/xilinx/rapidwright/interchange/PhysicalNetlistExample.java#L35 for an example).  It might be best to do that here as well.

src/com/xilinx/rapidwright/examples/GetRouteFailNets.java
```diff
@@ -0,0 +1,81 @@
...
+import com.xilinx.rapidwright.design.Design;
+import com.xilinx.rapidwright.design.DesignTools;
+import com.xilinx.rapidwright.design.Net;
+import com.xilinx.rapidwright.design.SitePinInst;
+
+
+/**
+ * Simple tool for get the partially routed nets in a design.
+ */
+public class GetRouteFailNets {
```
clavin-xlnx, line 37: Perhaps we could refactor `GetRouteFailNets` to `CountRoutedNets`?  That might better match the functionality.

clavin-xlnx (commented):
Thanks for the PR!

src/com/xilinx/rapidwright/examples/GetRouteFailNets.java
coherent17: Solved, thanks

src/com/xilinx/rapidwright/examples/GetRouteFailNets.java
coherent17: Solved, leave it only input dcp path is needed for the program argv!

src/com/xilinx/rapidwright/examples/GetRouteFailNets.java
coherent17: Update the class name and also the addFunction method in ```src/com/xilinx/rapidwright/MainEntrypoint.java```, thanks!

clavin-xlnx (approved):

PR #1173, opened 2025-04-29
[DesignTools] getAllRoutedSitePinsFromPhysicalPin() fix

eddieh-xlnx:
To accept VCC sitewires when searching for GND sitewires, since the latter will invert to the former


clavin-xlnx (approved):
Part of me is worried about this generating false positives, but I can't seem to think of any scenario where that might happen.

eddieh-xlnx:
> Part of me is worried about this generating false positives, but I can't seem to think of any scenario where that might happen.

That's fair. It does pass our internal suite though, so I'm going ahead with merging it.

PR #1182, opened 2025-05-03
[ModuleInst] Connect port to GND or VCC

clavin-xlnx:


src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -760,6 +762,52 @@ public void connect(String portName, int busIndex0, ModuleInst other, String oth
...
+                    + " is an output, cannot connect it to " + type);
+        }
+        Net physNet = type == NetType.GND ? getDesign().getGndNet() : getDesign().getVccNet();
+        EDIFNet logNet = EDIFTools.getStaticNet(type, getCellInst().getParentCell(), getDesign().getNetlist());
+        logNet.createPortInst(input.getName(), getCellInst());
+
+        for (SitePinInst pin : input.getSitePinInsts()) {
+            SitePinInst instPin = getCorrespondingPin(pin);
+            if (instPin.getNet() != null) {
+                instPin.getNet().removePin(pin);
```
eddieh-xlnx, line 805: Should we have a warning here that we're overriding a pin that's existing driven?

src/com/xilinx/rapidwright/design/ModuleInst.java
```diff
@@ -760,6 +762,52 @@ public void connect(String portName, int busIndex0, ModuleInst other, String oth
...
+                    "ERROR: Couldn't find port " + portName + ", idx=" + busIndex + " on ModuleInst "
+                            + getName());
+        }
+        if (input.isOutPort()) {
+            throw new RuntimeException("ERROR: Port " + input.getName() + " on ModuleInst " + getName()
+                    + " is an output, cannot connect it to " + type);
+        }
+        Net physNet = type == NetType.GND ? getDesign().getGndNet() : getDesign().getVccNet();
+        EDIFNet logNet = EDIFTools.getStaticNet(type, getCellInst().getParentCell(), getDesign().getNetlist());
+        logNet.createPortInst(input.getName(), getCellInst());
```
eddieh-xlnx, line 800: Do you need to handle the case where the port inst is already driven by a different logical net? (As you currently do below for the physical net)

test/src/com/xilinx/rapidwright/design/TestModuleInst.java
```diff
@@ -221,6 +223,61 @@ public void testConnectMovesPins() {
         Assertions.assertTrue(net2.getPins().isEmpty());
     }
 
+    @ParameterizedTest
+    @ValueSource(strings = { "GND", "VCC" })
+    public void testConnectGNDVCC(String staticTypeName) {
```
eddieh-xlnx, line 228: Does this test check both empty port insts and those that another net is already driving?

eddieh-xlnx:
To be clear, I think refusing to connect a port that's already connected to something else is perfectly acceptable.

src/com/xilinx/rapidwright/design/ModuleInst.java
clavin-xlnx: All ports of a `ModuleInst` are always connected to an empty `Net` when they are instantiated in a design.  The other `ModuleInst.connect()` methods also will remove this empty `Net` and connect it to the desired net.

src/com/xilinx/rapidwright/design/ModuleInst.java
clavin-xlnx: I've added code to handle this and updated the test accordingly.

src/com/xilinx/rapidwright/design/ModuleInst.java
eddieh-xlnx: I think your changes should address the case where a user calls `ModuleInst.connect()` once (as per your "... always connected to an empty `Net`" reply) as well as more than once (accidentally or intentionally) now.

eddieh-xlnx (approved):

Issue #1183, opened 2025-05-04
Infinite loop for int DesignTools::updatePinsIsRouted(Net net)


coherent17:
Hi RapidWright Developer,

I'm now read_checkpoint with partially routed design in the below code snippet:
```
        design = design.readCheckpoint(dcpFile, edifFile);
        DesignTools.makePhysNetNamesConsistent(design);
        DesignTools.createMissingSitePinInsts(design);
        DesignTools.updatePinsIsRouted(design);
```

Inside design/DesignTools.java

```java
    /**
     * Update the SitePinInst.isRouted() value of all sink pins in the given
     * Design. See {@link #updatePinsIsRouted(Net)}.
     * @param design Design in which pins are to be updated.
     * @return Number of unrouted sink pins (not driven by hierarchical ports) across design.
     */
    public static int updatePinsIsRouted(Design design) {
        int totalUnroutedSinkPins = 0;
        for (Net net : design.getNets()) {
            System.out.println("Updating pins for net " + net.getName());
            int numUnroutedSinkPins = updatePinsIsRouted(net);
            if (!DesignTools.isNetDrivenByHierPort(net)) {
                totalUnroutedSinkPins += numUnroutedSinkPins;
            }
        }
        return totalUnroutedSinkPins;
    }
```

There's a net with route status [Partially routed (route conflicts)] has a really large runtime in the function ```updatePinsIsRouted(net)``` been called in ```updatePinsIsRouted(design)``` method. The queue inside the function ```updatePinsIsRouted(net)``` is unable to become empty.

```java
    public static int updatePinsIsRouted(Net net) {
        ...
        while (!queue.isEmpty()) {
            NetTools.NodeTree node = queue.poll();
            SitePinInst spi = node2spi.remove(node);
            if (spi != null) {
                spi.setRouted(true);
                if (!spi.isOutPin()) {
                    assert(numUnroutedSinkPins > 0);
                    numUnroutedSinkPins--;
                }
            }
            queue.addAll(node.fanouts);
        }
        return numUnroutedSinkPins;
    }
```

Can you suggest how to resolve this issue?

Thanks,
Coherent

eddieh-xlnx:
Hi @coherent17. Thanks for the report!

[...]

PR #1184, opened 2025-05-06
[DesignTools] fix infinite loop in updatePinsIsRouted

coherent17:
This is my trial fix to solve the infinite loop as mentioned in #1183, simply add a hashset to avoid the traversed nodes been pushed into queue again.


coherent17:
Duplicate with PR #1185, closed.

PR #1201, opened 2025-05-28
Regroup Instances -- command line tool

clavin-xlnx:


src/com/xilinx/rapidwright/design/tools/RegroupInstances.java
```diff
@@ -0,0 +1,145 @@
...
+            System.out.println("    #   partition2/microblaze_0_local_memory_2/ilmb_v10");
+            return;
+        }
+
+        CodePerfTracker t = new CodePerfTracker("Regroup Instances");
+        Design design = null;
+        t.start("Read Design");
+        if (args[0].endsWith(".dcp")) {
+            design = Design.readCheckpoint(args[0], CodePerfTracker.SILENT);
+            design.unplaceDesign();
```
eddieh-xlnx, line 130: Can we document and issue a warning here that using this class on placed DCPs will result in everything (including instances that are not regrouped) from being unplaced (and unrouted)?

eddieh-xlnx: Am I right that unplacing is necessary because `ECOTools.refactorCell()` does not currently work for non-leaf cells which are placed? Let's leave that as a note.

eddieh-xlnx (commented):
Looks reasonable enough.

src/com/xilinx/rapidwright/design/tools/RegroupInstances.java
clavin-xlnx: > Can we document and issue a warning here that using this class on placed DCPs will result in everything (including instances that are not regrouped) from being unplaced (and unrouted)?

Warning added.

> Am I right that unplacing is necessary because `ECOTools.refactorCell()` does not currently work for non-leaf cells which are placed? Let's leave that as a note.

Yes, I've added a note.

eddieh-xlnx (commented):
Since you are using hierarchical paths, should the instances be uniquified prior to being regrouped?

clavin-xlnx:
> Since you are using hierarchical paths, should the instances be uniquified prior to being regrouped?

`ECOTools.refactorCell()` should check and unique-ify as needed: https://github.com/Xilinx/RapidWright/blob/e4c5030b1a59b81af329544abf4b3d34719da72d/src/com/xilinx/rapidwright/eco/ECOTools.java#L1221-L1222

PR #1205, opened 2025-05-29
LUT Input Cone Optimization

clavin-xlnx:


eddieh-xlnx (commented):
Since you are using hierarchical paths, should the instances of given pins be be uniquified prior to being optimized?

PR #1209, opened 2025-06-03
[ECOTools] Augments connectNet() to handle dual-LUT pin usage scenarios

clavin-xlnx:
This is to address the issue found in #1198.  It identifies scenarios where multiple LUTs are placed on the same complex and when a shared pin now needs to be connected to a different physical pin since the input net is now different.  This also adds a check to ensure the method only runs on disconnected pins and issues an error if a pin is still connected to a net.


test/src/com/xilinx/rapidwright/eco/TestECOTools.java
```diff
@@ -326,6 +327,95 @@ public void testConnectNetSwapSource() {
...
+        return design;
+    }
+    
+    @Test 
+    public void testConnectNetFailsWithoutDisconnect() { 
+        Design design = genConnectNetSwitchLUTInputTestDesign();
+        Cell testCell = design.getCell("cell1");
+        Net testNet = design.getNet("net7");
+        
+        // Ensure error is thrown in pin is still connected to a net
```
clavin-xlnx:
```suggestion
        // Ensure error is thrown if pin is still connected to a net
```

test/src/com/xilinx/rapidwright/eco/TestECOTools.java
```diff
@@ -326,12 +327,127 @@ public void testConnectNetSwapSource() {
         }
     }
 
+    /**
+     * Generated with CodeGenerator.genCodeForTestSite()
+     * From https://github.com/Xilinx/RapidWright/discussions/1198#discussioncomment-13335689
+     */
+    private Design genDiscussion1198TestCase() {
```
eddieh-xlnx, line 334: @clavin-xlnx Is this statement correct?

eddieh-xlnx (approved):
I fixed the issue with SPIs being wrongly disconnected when a companion LUT is using it, but the other two things we discussed:
- Re-using the same physical pin as the companion LUT when the net to be connected is the same
- Re-claiming the physical pins associated with unconnected logical pins

is more involved than I anticipated so I left a note for future us.

Also, I don't have a test to check that we fail gracefully for when there are not enough physical pins for the case when both LUTs combined require more than 6 unique ones.

test/src/com/xilinx/rapidwright/eco/TestECOTools.java
clavin-xlnx: Yes, thanks.  Used `CodeGenerator` to extract the test case.

test/src/com/xilinx/rapidwright/eco/TestECOTools.java
```diff
@@ -326,12 +327,128 @@ public void testConnectNetSwapSource() {
...
+        Assertions.assertEquals("C5", spiA5.getName());
+        Assertions.assertSame(targetNet, spiA5.getNet());
+        Assertions.assertSame(origNet, spiA2.getNet());
+
+        // Now disconnect from cell0 too
+        ECOTools.disconnectNet(design, cell0.getEDIFHierCellInst().getPortInst(pin));
+        Assertions.assertNull(spiA2.getNet());
+        // Connect that to targetNet too
+        ECOTools.connectNet(design, cell0, pin, targetNet);
+        // Currently it use the existing A2 input
```
eddieh-xlnx:
```suggestion
        // Check that it overwrites its existing A2 input
```

Issue #1220, opened 2025-06-15
[LUTTools] LUTTools.getLUTSize(c) return value for xcvp1902 device


coherent17:
Hi RapidWright Experts,

For versal device, the LUTTools.getLUTSize(cell) return 12 for some cells.
Can you help verify if it is a bug or not?

Example code snippet:
```java
    private static void dumpLUT() {
        int [] LUTCount = new int[16];
        for (Cell c : design.getCells()) {
            if(!LUTTools.isCellALUT(c)) continue;
            LUTCount[LUTTools.getLUTSize(c)]++;
            if (LUTTools.getLUTSize(c) > 6) {
                System.out.println("[WARNING] LUT" + LUTTools.getLUTSize(c) + ": " + c.getName() + " BEL: " + c.getBEL().getName());
            }
        }
        
        System.out.println("LUT Count:");
        int total = 0;
        for (int i = 1; i <= 15; i++) {
            total += LUTCount[i];
            System.out.println("LUT" + i + ": " + LUTCount[i]);
        }
        System.out.println("Total LUTs: " + total);
    }
```

Example output:
[WARNING] LUT12: XXX BEL: A6LUT
[WARNING] LUT12: XXX BEL: D5LUT
...

Some LUT5 and LUT6 cell in versal device LUTTools.getLUTSize(cell) will return 12.

LUT1: 16XXXX
LUT2: 77XXXX
LUT3: 17XXXX
LUT4: 49XXX
LUT5: 72XXX
LUT6: 38XXXX
LUT7 ~ LUT11: 0
LUT12: 50660

Thanks!

clavin-xlnx:
It looks like this is being caused by the Versal macro `LUT6CY` which is composed of two special `LUT5` instances, `LUTCY1` and `LUTCY2`.  I've pushed #1221 that attempts to fix this accounting problem, please give it a try and let us know.  Thanks!

coherent17:
Hi @clavin-xlnx,

Thanks for the fix, it's fixed on my side!

PR #1221, opened 2025-06-16
[LUTTools] Properly size LUTCY* primitives

clavin-xlnx:
This resolves an issue with the LUT6CY macro (composed of LUTCY1 and LUTCY2) present in Versal devices as reported in #1220.


eddieh-xlnx (approved):

eddieh-xlnx:
So `LUTTools.isCellALUT()` will continue returning `0` for `LUTCY*` primitives?

PR #1222, opened 2025-06-17
[ECOTools] Fix connectNet() for pins that don't need routing

eddieh-xlnx:
`CARRY8.CIN` can be logically connected to `VCC`, but does not physically need routing since it can be sourced from inside the BEL.

Do not error out in such cases.

### DIFF
--- a/src/com/xilinx/rapidwright/eco/ECOTools.java
+++ b/src/com/xilinx/rapidwright/eco/ECOTools.java
@@ -364,7 +364,8 @@ public class ECOTools {
         // Modify the physical netlist
         EDIFCell ecGnd = netlist.getHDIPrimitive(Unisim.GND);
         EDIFCell ecVcc = netlist.getHDIPrimitive(Unisim.VCC);
-        nextNet: for (EDIFHierNet ehn : netToPortInsts.keySet()) {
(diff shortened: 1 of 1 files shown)


clavin-xlnx (approved):

src/com/xilinx/rapidwright/eco/ECOTools.java
```diff
@@ -514,8 +516,15 @@ public static void connectNet(Design design,
...
                         createExitSitePinInst(design, ehpi, newPhysNet);
                     } else {
-                        // TODO: Find a new physical pin mapping
-                        throw new RuntimeException("ERROR: No logical-physical pin mapping found for pin '" + ehpi + "'");
+                        if (LUTTools.isCellALUT(cell)) {
+                            // TODO: Find a new physical pin mapping
+                            throw new RuntimeException("ERROR: No logical-physical pin mapping found for pin '" + ehpi + "'");
+                        } else {
+                            // Assume that sink does not need routing (e.g. CARRY8.CIN can be connected to VCC but does
+                            // not need physically routing)
```
eddieh-xlnx:
```suggestion
                            // Assume that sink does not need routing (e.g. CARRY8.CIN may already be connected to VCC
                            //  but does not need physically routing)
```

PR #1226, opened 2025-06-19
[EDIFNetlist] Update library when renaming cells; add API to remove blackbox prop

clavin-xlnx:


src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -341,7 +341,13 @@ public EDIFPort createPort(EDIFPort port) {
     }
 
     public void rename(String newName) {
+        if (getLibrary() != null) {
+            getLibrary().removeCell(this);
+        }
         setName(newName);
+        if (getLibrary() != null) {
```
eddieh-xlnx: I'm seeing that `EDIFLibrary.removeCell()` above calls `EDIFCell.clearLibrary()` -- are you seeing it enter this if branch?

src/com/xilinx/rapidwright/edif/EDIFCellInst.java
```diff
@@ -274,6 +274,11 @@ public boolean isBlackBox() {
         return false;
     }
 
+    public void removeBlackBoxProperty() {
+        removeProperty(BLACK_BOX_PROP);
+        removeProperty(BLACK_BOX_PROP_VERSAL);
```
eddieh-xlnx, line 279: For my education, how do these properties influence Vivado behaviour?

eddieh-xlnx: Can the blackbox-ness of a cell inst not be determined by looking at its cell and seeing if that's a blackbox? Is it possible that that cell doesn't exist at all?

src/com/xilinx/rapidwright/edif/EDIFCell.java
clavin-xlnx: Good catch, I've refactored the code to avoid loosing the reference to the library.

src/com/xilinx/rapidwright/edif/EDIFCellInst.java
clavin-xlnx: This change is mostly for consistency sake.  Leaving a property that says a cell is says `black_box=true` is confusing.  This is not the primary mode of deciding if a cell is a black box.

eddieh-xlnx (approved):

eddieh-xlnx (approved):

PR #1229, opened 2025-06-20
[NetlistBrowser] Fix extra click issue with instances

clavin-xlnx:
There is a bug where if the user clicks on a hierarchical instance more than once, it will not populate its children properly.


eddieh-xlnx:
Also smuggled a small fix in: https://github.com/Xilinx/RapidWright/pull/1229/commits/1d722edbfddf0408d587e1c2e7abf569ac934054

Issue #1228, opened 2025-06-20
Possible missing SiteInst


reillymck:
I might be missing the proper way to do this, but I am trying to interact with this site in Rapidwright. Here is a screenshot of the site in Vivado:

![Image](https://github.com/user-attachments/assets/a421b17c-48d6-4616-a649-bfb6682be8ca)

However, in rapidwright there does not exist a SiteInst. `design.getSiteInst("SLICE_X4Y124")` returns None. If I try and do `design.createSiteInst("SLICE_X4Y124")`, I get the following error message:
```
Traceback (most recent call last):
  File "com.xilinx.rapidwright.design.Design.java", line 0, in com.xilinx.rapidwright.design.Design.createSiteInst
Exception: Java Exception

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<console>", line 1, in <module>
java.lang.java.lang.RuntimeException: java.lang.RuntimeException: ERROR: Failed to create site instance for site SLICE_X4Y124
```
I have a script that iterates through SiteInsts and adds routethrus and constant generators to the EDIFNetlist. It is missing this one because there is no SiteInst here, and I am lost on how to detect it.

[quad_func_dpc_edf.zip](https://github.com/user-attachments/files/20841803/quad_func_dpc_edf.zip)

reillymck:
Update, I made the incorrect assumption that site instance names were always the name of the site. I used the function getSiteInstFromSiteName, and I was able to retrieve the instance.

reillymck:
I also see that `SiteInst.getNetFromSiteWire("A6LUT_O6")` for this site returns None.
`SiteInst.getConnectedNets()` returns an empty list for this site

I do think this behavior is unexpected, since signals from LUT constant generators are present in these calls for other site_inst objects.

clavin-xlnx:
> I also see that `SiteInst.getNetFromSiteWire("A6LUT_O6")` for this site returns None. `SiteInst.getConnectedNets()` returns an empty list for this site

This is by design.  When there are not cells in a SLICE other than a static source LUT, these `SiteInst` are special and do not appear in the general netlist.  

> I do think this behavior is unexpected, since signals from LUT constant generators are present in these calls for other site_inst objects.

Unfortunately, the net to site wire routing for static sources is inconsistent in DCPs from Vivado.  Sometimes the net to site wire mappings are optional and in this scenario, I believe that is the case.

PR #1231, opened 2025-06-20
Fix DesignTools.getConnectionPips for sinks driven by lut constant generators

reillymck:
Related to my issue #1228, I was tracing backwards from the sinks of the LUT constant generator in the static site. I wanted to use DesignTools.getConnectionPips() as a way to trace back to the SitePinInst of the driving site. This function failed as the biDir list is null and the current pip in the iteraction does not equal the source node of the global constant net. I assume in general this function will fail for SitePinInsts that are driven by LUT constant generators. I added some code to fix this.

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -33,9 +33,9 @@
 	<classpathentry kind="lib" path="jars/kryo-5.2.1.jar"/>
 	<classpathentry kind="lib" path="jars/minlog-1.3.1.jar"/>
 	<classpathentry kind="lib" path="jars/jython-standalone-2.7.2.jar"/>
-	<classpathentry kind="lib" path="jars/rapidwright-api-lib-2024.2.2.jar">
+	<classpathentry kind="lib" path="jars/rapidwright-api-lib-2024.2.3.jar">
 		<attributes>
-			<attribute name="javadoc_location" value="jar:platform:/resource/RapidWright/jars/rapidwright-api-lib-2024.2.2-javadoc.jar!/"/>
+			<attribute name="javadoc_location" value="jar:platform:/resource/RapidWright/jars/rapidwright-api-lib-2024.2.3-javadoc.jar!/"/>
 		</attributes>
 	</classpathentry>
 	<classpathentry kind="lib" path="jars/jgrapht-core-1.3.0.jar"/>
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
 
 env:
-  RAPIDWRIGHT_VERSION: v2024.2.2-beta
+  RAPIDWRIGHT_VERSION: v2024.2.3-beta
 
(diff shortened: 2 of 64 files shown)


clavin-xlnx:
Thanks @reillymck, will take a look next week.

eddieh-xlnx:
Hi @reillymck. Thanks for your contribution! Can you build a test or share a testcase which exhibits the problem that you are solving here please?

As an example, the test for bidir PIPs here:
https://github.com/Xilinx/RapidWright/blob/4e804cfe7257e2d7172a29db33a78f77a58acf37/test/src/com/xilinx/rapidwright/design/TestDesignTools.java#L1416-L1450

Also, wouldn't this affect all nets that don't have bidir PIPs, rather than just static nets?

reillymck:
Hey @eddieh-xlnx, thanks for the feedback. Also, sorry for the delayed response. 
        What is happening is there is a LUT  with no inputs and an equation "O6=1" driving the clk pin of another site. The FFs on that site are configured as LCDEs to be used as routethrus (the CLKINV BEL flips the signal as it enters the site). So when you query the net that is driving the CLK site pin, it shows up as `GLOBAL_LOGIC1`. However, if you follow the path of the pips on the connection, it brings you to the LUT's site. The reason this breaks `getConnectionPIPs`, is because the function initializes the source pip by calling Net..getSource().getConnectedNode(). 
        To answer your question about nets that do not have bidir pips, the function as it stands in the main branch is already able to handle these. This is because the `reverseNodeToPIPMap` will keep returning pips without any problems until it reaches the source pip. However, with my constant generator signal, it never reaches the source pip that is located after the global constant tieoff. Instead, it runs out of pips (because it has reached the true source pip). The current assumption that the function makes is that `reverseNodeToPIPMap` only returns `null` when a bidirectional pip exists in the path. So in my case, it tries to iterate an empty bidirectional pip list which throws an exception.  The only other case that I can think of where it may try to iterate an empty list is if there is a broken net with a pip missing for some reason. In that case, I think the function should still fail, which is why I have the check that net is indeed a constant net. 
I'm not quite sure how to replicate this "organic" Vivado design I have using the testing tools, but here is the design checkpoint with a script that will raise an exception if you run it on the code in main vs returning a list of correct pips if you run it on my branch. 
[test_getConnectionPips.zip](https://github.com/user-attachments/files/20982987/test_getConnectionPips.zip)

reillymck:
@eddieh-xlnx I was wondering what I needed to do to get this merged in. Do I need to add a full test case in TestDesignTools?

eddieh-xlnx:
Sorry that I haven't gotten to this. A test would certainly help me understand what the problem you're trying to solve is. It would be good to extract the relevant subtree from the relevant net of the DCP using https://github.com/Xilinx/RapidWright/blob/master/src/com/xilinx/rapidwright/util/CodeGenerator.java

Issue #1239, opened 2025-06-27
DesignTools.copyImplementation sometimes results in unplaced non-VCC/GND cells


korbin:
I'm working on a highly occupied (99%+) UltraScale+ design and am using `DesignTools.copyImplementation` to piece together separately-implemented (but identical) pblocks from multiple SLRs.

https://github.com/Xilinx/RapidWright/blob/4e804cfe7257e2d7172a29db33a78f77a58acf37/src/com/xilinx/rapidwright/design/DesignTools.java#L2840-L2841

For some reason, `dest.getSiteInstFromSite()` here is returning:

`SiteInst(name="STATIC_SOURCE_SLICE_X117Y252", type="SLICEL", site="SLICE_X117Y252")`

for a source SiteInst of:

`SiteInst(name="SLICE_X117Y252", type="SLICEL", site="SLICE_X117Y252")`

This is causing a small number of sites to be unplaced in the merged `Design`. Most Cells within the same bundle get copied exactly as expected, copying the routing works, but a handful of sites break here.

I don't believe the source to `dest.getSiteInstFromSite` is published or I would debug further. The part in question is a`xcvu35p-fsvh2104`

korbin:
Update: there's not a lot of documentation on what STATIC_SOURCE slices specifically *are*, but, copying this logic:

https://github.com/Xilinx/RapidWright/blob/e8126ba7adc471c3b254e1e3246da3601af9c15e/src/com/xilinx/rapidwright/design/tools/RelocationTools.java#L225-L227

into `DesignTools.copyImplementation` (near the existing null check) has resulted in the issue going away.

korbin:
Thanks for the additional info.

This was initially confusing because these `STATIC_SOURCE` sites were located within floorplanned `EXCLUDE_PLACEMENT` pblocks.

Upon further digging, I figured out what was happening. In this specific case, the culprit was my "greybox" creator function (could be useful for #356):

```Scala
def makeGreybox(dcp: Design, cellName: String): Unit = {
    val netlist = dcp.getNetlist()
    val top = netlist.getTopCell()
    val futureGreybox = netlist.getHierCellInstFromName(cellName).getInst()

    val oldCell = futureGreybox.getCellType()
    val greybox = new EDIFCell(futureGreybox.getCellType().getLibrary(), oldCell.getName() + "_gb")

    for (port <- oldCell.getPorts().asScala) {
      val netsPre = port.getInternalNets().asScala
[...]

Issue #356, opened 2022-03-04
Adding an API to insert a buffer cell (LUT1 or FF) on a module's ports


clavin-xlnx:
One useful capability would be to automatically add a buffer cell (LUT1 or FDRE) similar to the way "update_design -buffer_ports" operates in Vivado.  This would help with integrating multiple designs together into a single one.

PR #1242, opened 2025-06-30
Inline Flop Insertion/Removal Tool 

clavin-xlnx:
A tool designed to attached (and remove in a separate invocation) inline flip-flops to ports of an out-of-context DCP.  This is to help routability for kernels once put in-context that will be replicated later.

### DIFF
--- a/src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
+++ b/src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
@@ -68,7 +68,7 @@ public class InlineFlopTools {
 
     /**
      * Add flip flops inline on all the top-level ports of an out-of-context design.
-     * This is useful for placed out-of-context kernels prior to routing so that
+     * This is useful for out-of-context kernels prior to placement and routing so that
      * after the flops have been placed, the router is forced to route connections
      * of each of the ports to each of the flops. This can help alleviate congestion
      * when the kernels are placed/relocated in context. Note this assumes the
--- a/src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
+++ b/src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
@@ -39,6 +39,7 @@ import com.xilinx.rapidwright.design.SitePinInst;
 import com.xilinx.rapidwright.design.Unisim;
 import com.xilinx.rapidwright.design.blocks.PBlock;
 import com.xilinx.rapidwright.device.BEL;
+import com.xilinx.rapidwright.device.Series;
 import com.xilinx.rapidwright.device.Site;
 import com.xilinx.rapidwright.eco.ECOPlacementHelper;
 import com.xilinx.rapidwright.edif.EDIFCell;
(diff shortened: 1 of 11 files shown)


src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
```diff
@@ -0,0 +1,265 @@
...
+
+    public static final String INLINE_SUFFIX = "_rw_inline_flop";
+
+    /**
+     * Add flip flops inline on all the top-level ports of an out-of-context design.
+     * This is useful for placed out-of-context kernels prior to routing so that
+     * after the flops have been placed, the router is forced to route connections
+     * of each of the ports to each of the flops. This can help alleviate congestion
+     * when the kernels are placed/relocated in context. Note this assumes the
+     * design is not implemented as in most contexts it will be placed and routed
```
eddieh-xlnx, line 75: Here it says design is assumed to be not implemented (and there's the assertion to check that it's not placed, below) but on line 71 above it says:

> This is useful for placed out-of-context kernels prior to routing ...

Contradiction?

eddieh-xlnx (approved):

korbin:
This is very similar to what I am doing with the `createGreybox` function [here](https://github.com/Xilinx/RapidWright/issues/1239#issuecomment-3030308925), and if you squint, could be useful for #356.

I am using LUT1s for this purpose, though unclocked D-only or Q-only FDREs seemed to work as well, I wasn't sure which would be superior.

If using a D-only or Q-only FDRE _doesn't_ cause Vivado to emit static GND/VCC LUTs and routes (I didn't check that closely), FDREs might be better than LUT1s.

src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
```diff
@@ -0,0 +1,265 @@
...
+
+    private static final String CLK_OPT = "--clk";
+    private static final String PBLOCK_OPT = "--pblock";
+    private static final String REMOVE_FLOPS_OPT = "--remove_flops";
+
+    public static final String INLINE_SUFFIX = "_rw_inline_flop";
+
+    /**
+     * Add flip flops inline on all the top-level ports of an out-of-context design.
+     * This is useful for placed out-of-context kernels prior to routing so that
```
clavin-xlnx:
```suggestion
     * This is useful for out-of-context kernels prior to placement and routing so that
```

src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
clavin-xlnx: Thanks, I have reworded the comment.

clavin-xlnx:
> If using a D-only or Q-only FDRE _doesn't_ cause Vivado to emit static GND/VCC LUTs and routes (I didn't check that closely), FDREs might be better than LUT1s.

In theory, it should not, but may depend on the architecture.  There are inverters built into control set signal site inputs so GND doesn't have to be sourced explicitly.  We went with FDREs in this case over LUT1s as they can affect the timing of the port-connected paths and by moving the FDREs farther away from the out-of-context module, *potentially*, you could coax greater slack out of the connection.

PR #1254, opened 2025-07-21
Versal LUT Input Routing Fixes and Test

clavin-xlnx:


test/src/com/xilinx/rapidwright/eco/TestECOTools.java
```diff
@@ -445,6 +445,29 @@ public void testConnectNetSwitchLUTInput() {
...
+        Cell c = design.getCell("processor/upper_parity_lut");
+        String pin = "I2";
+        String siteWire = c.getSiteWireNameFromLogicalPin(pin);
+        SitePinInst spi = c.getSitePinFromLogicalPin(pin, null);
+        SiteInst si = c.getSiteInst();
+        Net net = si.getNetFromSiteWire(siteWire);
+        EDIFHierPortInst portInst = c.getEDIFHierCellInst().getPortInst(pin);
+        ECOTools.disconnectNet(design, portInst);
+        // TODO Question: should disconnectNet() be unrouting IMR routethrus? Is this next line necessary?
+        si.unrouteIntraSiteNet(spi.getBELPin(), c.getBELPin(portInst.getPortInst()));
```
eddieh-xlnx: I think `disconnectNet()` should unroute/remove IMRs to leave a valid design behind. If it doesn't I would consider that a bug that should be fixed.

PR #1253, opened 2025-07-17
[EDIFTools] Add support for connecting logical nets through both hierarchical directions

clavin-xlnx:

PR #1258, opened 2025-07-23
[EDIFTools] Fix connectPortInstsThruHier when finding source pin

eddieh-xlnx:
Fixes issue introduced in #1253

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFTools.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFTools.java
@@ -701,7 +701,7 @@ public class EDIFTools {
                 // Need to go up in hierarchy ...
                 EDIFHierNet currNet;
                 if (pinIsOutput) {
-                    // ... fom the target pin (which is an input)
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -698,10 +698,17 @@ public static void connectPortInstsThruHier(EDIFHierNet net, EDIFHierPortInst pi
                 }
             } 
             if (commonAncestor.getDepth() <= pinInst.getDepth()) {
-                // Need to go up in hierarchy
-                EDIFHierNet currNet = pin.getHierarchicalNet();
+                // Need to go up in hierarchy ...
+                EDIFHierNet currNet;
+                if (pinIsOutput) {
+                    // ... fom the target pin (which is an input)
```
clavin-xlnx:
```suggestion
                    // ... from the target pin (which is an input)
```

PR #1260, opened 2025-07-29
Netlist Comparator Inst and Property Tracking

EddieRydell:
This PR extends EDIFNetlistComparator to provide more granular context for reported differences by tracking:
* The source instance (EDIFCellInst) associated with the difference, when applicable
* The property key for property-related mismatches

These additions are reflected in the EDIFDiff class and its toString() method, which now includes inst= and property= annotations to improve traceability during debugging.

### DIFF
--- a/src/com/xilinx/rapidwright/edif/compare/EDIFNetlistComparator.java
+++ b/src/com/xilinx/rapidwright/edif/compare/EDIFNetlistComparator.java
@@ -335,18 +335,11 @@ public class EDIFNetlistComparator {
         addDiff(type, gold, test, parentCell, parentLibrary, notEqualString, null, null);
     }
 
-    private void addDiff(
-            EDIFDiffType type,
-            Object gold,
-            Object test,
-            EDIFCell parentCell,
-            EDIFLibrary parentLibrary,
-            String notEqualString,
-            EDIFCellInst sourceInst,
-            String propertyKey) {
+    private void addDiff(EDIFDiffType type, Object gold, Object test, EDIFCell parentCell,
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/edif/compare/EDIFNetlistComparator.java
```diff
@@ -307,14 +326,28 @@ private void checkDiff(Object checkGold, Object checkTest, EDIFDiffType type, Ob
...
+            addDiff(type, gold, test, parentCell, parentLibrary, notEqualString, inst, propertyKey);
         }
     }
 
     private void addDiff(EDIFDiffType type, Object gold, Object test, EDIFCell parentCell,
             EDIFLibrary parentLibrary, String notEqualString) {
+        addDiff(type, gold, test, parentCell, parentLibrary, notEqualString, null, null);
+    }
+
+    private void addDiff(
```
clavin-xlnx: Could you re-arrange the parameters to be on 2 or 3 lines to match the similar method prototype above?  It would make it easier to 'diff' it visually if they were laid out in a similar fashion.

clavin-xlnx (commented):
Thanks for the contribution!  I was going to ask if you could help us update our tests to check for the added functionality, but I realized, we had no tests for this class!  So, I went and wrote a simple one (https://github.com/Xilinx/RapidWright/pull/1263).  Give us a few days to get it merged in and then if you could please augment it to do some very basic checking for the improvements you are making, that would be great.

clavin-xlnx (approved):
Thank you!

PR #1262, opened 2025-07-29
Add support for shared sink SitePinInsts in LUTInputConeOpt

clavin-xlnx:


src/com/xilinx/rapidwright/eco/LUTInputConeOpt.java
```diff
@@ -186,6 +212,16 @@ private static Cell createAndConnectCell(Design design, EDIFHierCellInst hierLut
         List<EDIFHierPortInst> pinsToConnect = new ArrayList<>();
         pinsToConnect.add(input);
         pinsToConnect.add(new EDIFHierPortInst(hierLutInst.getParent(), lutInst.getOrCreatePortInst("O")));
+
+        // Check for other cells physically sharing the same SitePinInst input
+        List<EDIFHierPortInst> otherPins = getSharedSinksInSite(design, input);
+        if (otherPins.size() > 0) {
+            // Include other sinks in the optimization by disconnecting them also and
+            // connecting them to the optimized LUT output
+            ECOTools.disconnectNet(design, otherPins);
```
eddieh-xlnx, line 225: Will `input` always be present in `otherPins`, whether there are shared sinks or not? If so, would `otherPins.size() > 0` always be true?

clavin-xlnx: `input` should not be present in `otherPins`, therefore, if the input is not shared, `otherPins` should be empty.

clavin-xlnx:
> Is `EDIFHierPortInst input` the right argument? It might be for a non-leaf port inst (e.g. an input to a `LUT6_2` macro) but that's not always the case (e.g. a `CE` input shared between leaf cells from different hierarchies packed into the same slice).

This is the same argument provided by the user.  We are just looking to see if it is placed and if it is, if it shares the site net with any other placed cells.  

> It seems like `SitePinInst input` might be a better argument as that reflects placement constraints, but of course, then it won't work on unplaced designs.

Correct, we want this to work for unplaced, placed, and routed designs.  `SitePinInst` requires some effort on the caller and abstracts away some of the details I believe we need.

eddieh-xlnx:
> > Is `EDIFHierPortInst input` the right argument? It might be for a non-leaf port inst (e.g. an input to a `LUT6_2` macro) but that's not always the case (e.g. a `CE` input shared between leaf cells from different hierarchies packed into the same slice).
> 
> This is the same argument provided by the user/caller of `optimizedLUTInputCone()`. We are just looking to see if it is placed and if it is, if it shares the site net with any other placed cells.

My view is to look more holistically at the whole thing. If I provide an input `cell1/CE` and it unavoidably modifies `cell2/CE` too (or perhaps `cell3/A1` and `cell4/A1`) without telling me that might seem a little imprecise/if not pushy, particularly if for some reason I didn't want/need it to change `cell4/A1` for some reason. Just raising this concern, I'll let you judge if it is important enough to deal with. 

> > It seems like `SitePinInst input` might be a better argument as that reflects placement constraints, but of course, then it won't work on unplaced designs.
> 
> Correct, we want this to work for unplaced, placed, and routed designs. `SitePinInst` requires some effort on the caller and abstracts away some of the details I believe we need.

On the flip side, this entire issue isn't applicable when operating on an unplaced design. Put another way, behaviour of `LUTInputConeOpt` may be different between placed and unplaced designs.
It could also be argued that

PR #1264, opened 2025-08-04
Fix an issue on calculateUtilization function for Versal devices.

kylepham-amd:


clavin-xlnx (approved):
Thanks for the contribution and including the tests!

PR #1265, opened 2025-08-04
Full TCL parser for XDC

jakobwenzel:
Hi there!

This PR implements a full TCL parser for XDC. The parser handles package pin, `create_clock` and `set_property` constraints. If the Parser encounters unsupported commands or command options, the parsed TCL code is converted back to strings and returned as `UnsupportedConstraintElement`s. Support for more constraint types can be easily added by implementing more TCL commands.

The parser can be called with a netlist and then supports cell lookup via `get_cells` with a number of filtering flags. If called without a netlist, `get_cells` support is more limited.

As discussed previously, this introduces a dependency on a modified version of JACL ([upstream](https://tcljava.sourceforge.net/docs/website/index.html), [modified version](https://github.com/jakobwenzel/jacl)). This is not yet on Maven Central, so I have not added the actual dependency in the Gradle build. CI will fail for now. For testing, I have added the jars in a directory inside RapidWright ([see this branch](https://github.com/jakobwenzel/RapidWright/tree/full-tcl-parser-dependencies)).

In my use case, the source Verilog is restructured before creating an EDIF. Therefore, XDC needed to be rewritten as well. This led to a few design decisions in the XDC parser:
* The EDIF cell lookup is customizable via the interface `EDIFCellLookup` (with default implementations for lookup with and without a netlist). My rewriting implementation is not part of this PR.
* `EDIFCellLookup` has a notion of original (pre-rewrite) and final (post-rewrite) design names.
* In XDC code, cells returned by `get_cells` calls can be stringified (and appended to get a sub-cell) and be passed to `get_cells` again. For my use case, the exact Java Instance needed to be found again. Stringifying cells therefore returns a hash code, which can be converted back to Java instances in further XDC calls. This is handled by `TclHashIdentifiedObject`.

## Open Questions
* I have a testcase which uses https://github.com/alexforencich/verilog-ethernet, which has quite complex XDC. This code is under MIT license. Currently, my testcase just points to an EDIF file on my disk... I assume it is ok to put an EDIF of that under https://github.com/eddieh-xlnx/RapidWrightDCP if a license notice is included? (Will open a PR if I get a go ahead) Should we fetch the XDC from upstream or copy it into our Git somewhere?
* As there now are quite a few classes for XDC, it might make sense to move everything out of `ipi` into its own package. I was hesistant because of backwards compatibility.
* In `XDCParser`, the `parseXDC` methods used to return `Map<String,PackagePinConstraint>`. The new parser returns instances of `XDCConstraints`. We can either break backwards compatibility and have change the existing methods' return type, or we keep the old return type and choose a new method name for returning `XDCConstraints` (currently implemented, as `parseXDCNew`).
* I assume you would want to handle publishing JACL onto Maven Central to have everything under your control?

I am planning to add some more Javadoc in the coming days.

Jakob

### DIFF
--- a/src/com/xilinx/rapidwright/ipi/BlockStitcher.java
+++ b/src/com/xilinx/rapidwright/ipi/BlockStitcher.java
@@ -30,6 +30,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Queue;
 import java.util.Set;
@@ -137,7 +138,7 @@ public class BlockStitcher {
      * @param design
      * @param constraints
      */
-    public void stitchDesign(Design design, HashMap<String,PackagePinConstraint> constraints) {
+    public void stitchDesign(Design design, Map<String,PackagePinConstraint> constraints) {
         boolean debug = false;
         EDIFNetlist n = design.getNetlist();
         // Create a reverse parent net map (Parent Net -> Children Nets: all logical nets that are physically equivalent)
@@ -271,7 +272,7 @@ public class BlockStitcher {
                 Net portNet = null;
                 portNet = e.getValue();
 
-                Site site = design.getDevice().getSiteFromPackagePin(constraints.get(portName).getName());
+                Site site = design.getDevice().getSiteFromPackagePin(constraints.get(portName).getPackagePin());
                 if (site == null) {
                     MessageGenerator.briefMessage("WARNING: It appears that the I/O called " + portName + " is not assigned to a package pin!");
                     continue nextPort;
@@ -294,7 +295,7 @@ public class BlockStitcher {
                 }
 
                 String ioStandard = constraints.get(portName).getIoStandard();
-                String pkgPin = constraints.get(portName).getName();
+                String pkgPin = constraints.get(portName).getPackagePin();
                 EDIFNet logNet = e.getKey().getPortInst().getNet();
                 design.createAndPlaceIOB(portName, isPortOutput ? PinType.OUT : PinType.IN, pkgPin, ioStandard, portNet, logNet);
             }
@@ -562,7 +563,7 @@ public class BlockStitcher {
         runtimes[2] = System.currentTimeMillis();
         System.out.println("Total Blocks : " + totalBlocks);
         String xdcFileName = args[1].replace(".edf", ".xdc");
-        HashMap<String,PackagePinConstraint> constraints = null;
+        Map<String,PackagePinConstraint> constraints = null;
         if (new File(xdcFileName).exists()) {
             constraints = XDCParser.parseXDC(xdcFileName,stitched.getDevice());
         } else {
--- a/src/com/xilinx/rapidwright/ipi/ClockConstraint.java
+++ b/src/com/xilinx/rapidwright/ipi/ClockConstraint.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2022, Advanced Micro Devices, Inc.
+ * All rights reserved.
+ *
+ * Author: Jakob Wenzel
+ *
+ * This file is part of RapidWright.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.xilinx.rapidwright.ipi;
+
+import java.util.Locale;
+
+public class ClockConstraint implements Constraint<ClockConstraint>{
+    private String clockName;
+    private double period;
+
+    private String portName;
+
+    public ClockConstraint(String clockName, double period, String portName) {
+        this.clockName = clockName;
+        this.period = period;
+        this.portName = portName;
+    }
+
+    @Override
+    public ClockConstraint clone() {
+        return new ClockConstraint(clockName, period, portName);
+    }
+
+    public String getClockName() {
+        return clockName;
+    }
+
+    public void setClockName(String clockName) {
+        this.clockName = clockName;
+    }
+
+    public double getPeriod() {
+        return period;
+    }
+
+    public void setPeriod(double period) {
+        this.period = period;
+    }
+
+    public String asXdc() {
+        String periodString = String.format(Locale.US, "%.3f", period);
+        return "create_clock -period "+periodString+" -name "+clockName+" [get_ports "+portName+"]";
+    }
+
+    public String getPortName() {
+        return portName;
+    }
+
+    public void setPortName(String portName) {
+        this.portName = portName;
+    }
+}
--- a/src/com/xilinx/rapidwright/ipi/Constraint.java
+++ b/src/com/xilinx/rapidwright/ipi/Constraint.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright (c) 2024, Advanced Micro Devices, Inc.
+ * All rights reserved.
+ *
+ * Author: Jakob Wenzel
+ *
+ * This file is part of RapidWright.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.xilinx.rapidwright.ipi;
+public interface Constraint<T extends Constraint<T>> {
+    T clone();
+}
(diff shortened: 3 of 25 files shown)


jakobwenzel:
> Thanks @jakobwenzel, this is awesome!
> 
> > ## Open Questions
> > 
> > * I have a testcase which uses https://github.com/alexforencich/verilog-ethernet, which has quite complex XDC. This code is under MIT license. Currently, my testcase just points to an EDIF file on my disk... I assume it is ok to put an EDIF of that under https://github.com/eddieh-xlnx/RapidWrightDCP if a license notice is included? (Will open a PR if I get a go ahead) Should we fetch the XDC from upstream or copy it into our Git somewhere?
> 
> Would it make sense to simply combine the compiled design (EDIF and XDC) into a single DCP? I think that might make the most sense and then we could extract the XDC directly if needed.
> 

Great Idea! The testcase uses the source XDC instead of an XDC that has gone through Vivado, but manually building a DCP should work.

> > * As there now are quite a few classes for XDC, it might make sense to move everything out of `ipi` into its own package. I was hesistant because of backwards compatibility.
> 
> Sure, I am open to having an `xdc` package if that can be abstracted more or less cleanly enough.
> 

Ok, I will do some refactoring.

> > * In `XDCParser`, the `parseXDC` methods used to return `Map<String,PackagePinConstraint>`. The new parser returns instances of `XDCConstraints`. We can either break backwards compatibility and have change the existing methods' return type, or we keep the old return type and choose a new method name for returning `XDCConstraints` (currently implemented, as `parseXDCNew`).
> 
> I think we can break the old method as it wasn't very robust. I'm not sure that it was being used broadly anyways. Having this in place will make it much more feasible/useful to use.
> 

Ok, will replace the old method.

> > * I assume you would want to handle publishing JACL onto Maven Central to have everything under your control?
> 
> Do you have a sense for how much modification would be needed going forward on JACL? I'm not sure that we would want to adopt/maintain JACL, but if it is static enough, we could possibly just publish a snapshot after legal scans and such.

I believe it would be a very static thing. The only changes I made so far were making more stuff public to enable re-stringifying parsed TCL code for unsupported constraints

clavin-xlnx:
> I believe it would be a very static thing. The only changes I made so far were making more stuff public to enable re-stringifying parsed TCL code for unsupported constraints

Sounds good, I will look at getting the code reviewed/scanned to see if we can provide a static build on Maven Central as a dependency.  It might take a while so no hurry on the other changes.

src/com/xilinx/rapidwright/design/xdc/parser/CellObject.java
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright (c) 2022, Advanced Micro Devices, Inc.
```
clavin-xlnx: All the headers probably need to be updated to reflect the correct year and your university.

src/com/xilinx/rapidwright/design/xdc/parser/CellObject.java
```diff
@@ -0,0 +1,94 @@
...
+package com.xilinx.rapidwright.design.xdc.parser;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import com.xilinx.rapidwright.design.xdc.UnsupportedConstraintElement;
+
+/**
+ * TCL representation of a list of cells
```
clavin-xlnx: We should probably change all instances of `TCL --> Tcl` (See https://en.wikipedia.org/wiki/Tcl#:~:text=The%20name%20originally%20comes%20from,written%20Tcl%20rather%20than%20TCL.)

src/com/xilinx/rapidwright/design/xdc/parser/CreateClockCommand.java
```diff
@@ -0,0 +1,91 @@
...
+                    case "-waveform":
+                        //Just skip the waveform specification
+                        ++i;
+                        break;
+                    default:
+                        throw new RuntimeException("expected -name | -period | -waveform but got " + objv[i]);
+                }
+            } else {
+
+                DesignObject obj = DesignObject.requireUnwrapTclObject(interp, objv[i], cellLookup);
```
clavin-xlnx, line 72:
```suggestion
            } else {
                DesignObject obj = DesignObject.requireUnwrapTclObject(interp, objv[i], cellLookup);
```

src/com/xilinx/rapidwright/design/xdc/parser/DesignObject.java
```diff
@@ -0,0 +1,132 @@
...
+import tcl.lang.TclList;
+import tcl.lang.TclObject;
+
+/**
+ * Base class for any object that will be referenced from TCL
+ * @param <T> Representation of Cells
+ */
+public abstract class DesignObject<T> {
+    /**
+     * Cast an
```
clavin-xlnx: Maybe add a little bit more detail?

src/com/xilinx/rapidwright/design/xdc/parser/GetCellsCommand.java
```diff
@@ -0,0 +1,270 @@
...
+                    }
+                    cellNameStr = TclHashIdentifiedObject.unpackAsString(interp, argv[i].toString(), cellLookup);
+                    break;
+            }
+
+        }
+
+        if (cellNameStr != null && !regexpFlag && filter == null && !hierFlag) {
+
+            simpleGetCells(interp, cellNameStr);
```
clavin-xlnx:
```suggestion
        if (cellNameStr != null && !regexpFlag && filter == null && !hierFlag) {
            simpleGetCells(interp, cellNameStr);
```

src/com/xilinx/rapidwright/design/xdc/parser/GetCellsCommand.java
```diff
@@ -0,0 +1,270 @@
...
+    private void complexGetCells(Interp interp, boolean hierFlag, boolean regexFlag, String cellNames, TclObject filterArg, TclObject[] argv) throws TclException {
+        Stream<T> candidateSupplier = null;
+        List<Pair<Predicate<T>, Function<T, String>>> filter = new ArrayList<>();
+
+        if (!hierFlag) {
+            throw new RuntimeException("no hier flag? not implemented");
+        }
+
+
+        if (filterArg != null) {
```
clavin-xlnx, line 163:
```suggestion
        }

        if (filterArg != null) {
```

src/com/xilinx/rapidwright/design/xdc/parser/NameDesignObject.java
```diff
@@ -0,0 +1,94 @@
...
+                    new UnsupportedConstraintElement.SyntaxConstraintElement("["),
+                    new UnsupportedConstraintElement.NameConstraintElement(type.getXdcCommand()),
+                    new UnsupportedConstraintElement.SyntaxConstraintElement("]")
+                    );
+        }
+        boolean braces = objects.size()!=1 || objects.stream().anyMatch(o->o.contains("*") || o.contains("$"));
+
+
+
+        List<UnsupportedConstraintElement> before = new ArrayList<>(Arrays.asList(
```
clavin-xlnx:
```suggestion
        boolean braces = objects.size()!=1 || objects.stream().anyMatch(o->o.contains("*") || o.contains("$"));

        List<UnsupportedConstraintElement> before = new ArrayList<>(Arrays.asList(
```

src/com/xilinx/rapidwright/design/xdc/parser/NoNetlistLookup.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override
+    public Class<String> getCellClass() {
+        return String.class;
+    }
+
+    @Override
+    public String getInstFromFinalName(String s) {
+        return s;
+    }
+}
```
clavin-xlnx: Newline

src/com/xilinx/rapidwright/design/xdc/parser/RegularEdifCellLookup.java
```diff
@@ -0,0 +1,108 @@
...
+package com.xilinx.rapidwright.design.xdc.parser;
+
+import java.util.stream.Stream;
+
+import com.xilinx.rapidwright.edif.EDIFCellInst;
+import com.xilinx.rapidwright.edif.EDIFHierCellInst;
+import com.xilinx.rapidwright.edif.EDIFNetlist;
+import com.xilinx.rapidwright.edif.EDIFPropertyValue;
+
+public class RegularEdifCellLookup extends EdifCellLookup<EDIFHierCellInst> {
```
clavin-xlnx, line 38: Maybe add a brief Javadoc description?

src/com/xilinx/rapidwright/design/xdc/XDCParser.java
```diff
@@ -0,0 +1,275 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+/**
+ *
+ */
```
clavin-xlnx:
```suggestion

```

src/com/xilinx/rapidwright/design/xdc/parser/DesignObject.java
jakobwenzel: Oops, forgot to finish this line

jakobwenzel:
I have added my testcases. I think this PR is now ready.

I expanded my old testcase to include checking if parsing constraints, stringifying, then parsing again gives identical results to just parsing. To make this work for parsing without netlists, this required a few changes in the parser.

Previously, to parse constraints without netlists, you could either set lookup to null, or use NoNetlistLookup. I realized that NoNetlistLookup offered no benefits, so I removed it.

jakobwenzel:
I logged the test failure as #1320 . This can be worked around by removing the true at https://github.com/jakobwenzel/RapidWright/blob/full-tcl-parser/test/src/com/xilinx/rapidwright/design/xdc/TestXDCParser.java#L985 for a performance penalty.

clavin-xlnx (approved):

PR #1268, opened 2025-08-06
[RWRoute] Handle Missing INTF Tile

clavin-xlnx:
Example failure in https://github.com/Xilinx/RapidWright/discussions/1245#discussioncomment-14003055.


eddieh-xlnx:
Turns out the VP1202 really doesn't have tiles with type `INTF_LOCF_B[LR]_TILE`.

I did notice a few more `INTF_*` tiles that were present on the VP1202, so I added those too.

PR #1270, opened 2025-08-09
[ECOTools.connectNet()] Fixes static net connection and no source issues

clavin-xlnx:


test/src/com/xilinx/rapidwright/eco/TestECOTools.java
```diff
@@ -1118,4 +1119,66 @@ public void testDiscussion1245(String device) {
...
+//        Cell clkIBUF = design.createAndPlaceCell(top, "clkIBUF", Unisim.IBUF, "IOB_X29Y0/IOB_M");
+//        ECOTools.connectNet(design, clkIBUF, "I", clk);
+//        
+//        Cell gateIBUF = design.createAndPlaceCell(top, "gateIBUF", Unisim.IBUF, "IOB_X29Y1/IOB_M");
+//        ECOTools.connectNet(design, gateIBUF, "I", gate);
+//
+//        Net clkBUF = design.createNet("clkBUF");
+//        ECOTools.connectNet(design, clkIBUF, "O", clkBUF);
+//        Net gateBUF = design.createNet("gateBUF");
+//        ECOTools.connectNet(design, gateIBUF, "O", gateBUF);
```
eddieh-xlnx, line 1147: Are we expecting a primitives fix imminently? This code seems to attach sources to these nets, which sounds like it could defeat the purpose of this test?

test/src/com/xilinx/rapidwright/eco/TestECOTools.java
```diff
@@ -1118,4 +1119,66 @@ public void testDiscussion1245(String device) {
...
+        ECOTools.connectNet(design, ff2, "Q", ff2Feedback);
+
+        Net vcc = design.getVccNet();
+        Net gnd = design.getGndNet();
+        ECOTools.connectNet(design, ff1, "CE", vcc);
+        ECOTools.connectNet(design, ff2, "CE", vcc);
+        ECOTools.connectNet(design, ff1, "R", gnd);
+        ECOTools.connectNet(design, ff2, "R", gnd);
+
+        RWRoute.routeDesignFullNonTimingDriven(design);
```
eddieh-xlnx, line 1181: Are you expecting a fully routed design after this call? Might it be helpful to check using Vivado?

eddieh-xlnx (commented):
Title mentions "static net" but I'm not seeing any static net specific behaviour/testing?

test/src/com/xilinx/rapidwright/eco/TestECOTools.java
clavin-xlnx: It turns out there are device-specific Versal primitive mappings that we are not capturing properly.  This needs to be fixed before this code can be enabled, but should be part of the test in the future.  I don't think this PR should wait for the primitive mappings update.

test/src/com/xilinx/rapidwright/eco/TestECOTools.java
clavin-xlnx: No, the clock won't be routed and currently Vivado is inserting IBUFs because I can't insert them yet.  When we get the primitive mappings fixed, that can be added.

test/src/com/xilinx/rapidwright/eco/TestECOTools.java
```diff
@@ -1118,4 +1119,66 @@ public void testDiscussion1245(String device) {
...
+        ECOTools.connectNet(design, ff1, "Q", ff1Feedback);
+        ECOTools.connectNet(design, ff2, "D", ff2Feedback);
+        ECOTools.connectNet(design, ff2, "Q", ff2Feedback);
+
+        Net vcc = design.getVccNet();
+        Net gnd = design.getGndNet();
+        ECOTools.connectNet(design, ff1, "CE", vcc);
+        ECOTools.connectNet(design, ff2, "CE", vcc);
+        ECOTools.connectNet(design, ff1, "R", gnd);
+        ECOTools.connectNet(design, ff2, "R", gnd);
```
clavin-xlnx, line 1179: > Title mentions "static net" but I'm not seeing any static net specific behaviour/testing?

Here we are connecting GND and VCC to the flip flops.

eddieh-xlnx: Missed that, thanks!

test/src/com/xilinx/rapidwright/eco/TestECOTools.java
eddieh-xlnx: Agreed.

PR #1278, opened 2025-08-25
[DesignTools] Handles VCC/GND connections properly in populateBlackBox()

clavin-xlnx:
If the top level design is driving one of the ports of the black box with GND or VCC, this should recognize that it should switch the parent net to their physical net equivalents (`GLOBAL_LOGIC[1|0]`).

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -945,6 +945,8 @@ public class DesignTools {
         // We need to prefix all cell and net names with the hierarchicalCellName as a prefix
         Net vcc = design.getVccNet();
         Net gnd = design.getGndNet();
+        Net vccCell = cell.getVccNet();
+        Net gndCell = cell.getGndNet();
         for (SiteInst si : cell.getSiteInsts()) {
             for (Cell c : new ArrayList<Cell>(si.getCells())) {
                 c.updateName(hierarchicalCellName + "/" + c.getName());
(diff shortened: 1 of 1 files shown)


eddieh-xlnx (approved):

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -955,6 +957,15 @@ public static void populateBlackBox(Design design, String hierarchicalCellName,
                 }
             }
             design.addSiteInst(si);
+            // Update GND/VCC site routing to point to destination design's GND/VCC nets
+            for (String siteWire : si.getSiteWiresFromNet(vcc)) {
+                BELPin pin = si.getSiteWirePins(siteWire)[0];
+                si.routeIntraSiteNet(vcc, pin, pin);
+            }
```
eddieh-xlnx: Hang on, isn't this redundant? You're asking for all sitewires from the `vcc` net, and then you're re-routing those sitewires onto the same `vcc` net?

eddieh-xlnx (commented):
Added comment

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx: The problem is we are merging two different designs together.  They each have an instance of the GND and VCC nets.  If I don't update the reference pointers to the destination design's GND or VCC net, this method will fail because it checks references rather then `.equals()`:
https://github.com/Xilinx/RapidWright/blob/f6df54954f3280003360eb9af69d56bb01eb713f/src/com/xilinx/rapidwright/design/DesignTools.java#L3494

eddieh-xlnx: Hmm. So `vcc` is from the destination design, but `si.getSiteWiresFromNet(vcc)` will return the sitewires painted by a different `vcc` from the source design (because it uses `Net.equals()`) but that is not what the code you quoted uses.

clavin-xlnx: `SiteInst.getSiteWiresFromNet()` uses `Net.equals()` under the hood so it relies on the name of the `Net` rather than the actual reference to do the lookup.

clavin-xlnx: It comes down to using reference equivalence or object equivalence when checking if two nets are equal.  I think, in theory, reference equivalence should work under normal conditions.  However, we haven't standardized everywhere in the code base on that idea, especially in the closed source portion of the code.  For more unusual scenarios, like combining two designs together, checking by object equivalence makes things easier during the translation process.

PR #1280, opened 2025-08-29
[RWRoute] Versal fixes for clocking, static nets and general routing

clavin-xlnx:
### Versal Clock Improvements
1. Several clocking methods were not checking against existing used nodes which created the possibility for multi-clock designs to use the same nodes.  This PR ensures that clocks cannot accidentally use the same node.
2. There are a finite set of routing and distribution tracks in UltraScale/UltraScale+ and Versal architectures.  Once a numbered track is chosen, the clock net will only successfully route by staying on that track.  In a multi-clock design, we must avoid using the same track for the same clock region to avoid routing failures.  Hence, only clocks that are mutually exclusive based on the clock regions they occupy can use the same track, otherwise, they should be avoided.  This PR adds basic logic to avoid track collisions by making sure the centroid of a clock doesn't use a track already used by a previously routed clock net.
3. Some clock pin sinks are not always served by a leaf clock buffer (LCB).  For example, when driving a clock off chip through an IOB, this routing path is different.  This PR adds code to route these pins appropriately.
4. When a sink on the clock net is the PS, it is driven by the neighboring clock region through an LCB.  This adds logic to detect and correctly handle these sinks.
5. In Versal, it appears that not all clock regions can serve as clock roots (centroids).  It is still unclear as to how to predict which clock regions can serve as a clock root.   This PR handles this issue by iterative searching across a row of candidate clock regions until a successful clock root (centroid) is found.
6. Follows the Vivado router pattern in Versal XPHY of BEL pin remapping when driven by DMC (memory controller) to use the dedicated pins.
7. The Versal clock router was allowing multiple clock roots to be generated (VROUTE->VDISTR transitions).  The warning from Vivado was: `WARNING: [Constraints 18-11790] Suboptimal number of clock roots found for net 'soc_vip_i/versal_cips_0/inst/pspmc_0/inst/pl0_ref_clk', expecting single clock root, but found 2, Delay estimation/calculation will be inaccurate for this clock net.`

### DIFF
--- a/src/com/xilinx/rapidwright/router/VersalClockRouting.java
+++ b/src/com/xilinx/rapidwright/router/VersalClockRouting.java
@@ -43,6 +43,7 @@ import com.xilinx.rapidwright.design.DesignTools;
 import com.xilinx.rapidwright.design.Net;
 import com.xilinx.rapidwright.design.SitePinInst;
 import com.xilinx.rapidwright.device.ClockRegion;
+import com.xilinx.rapidwright.device.Device;
 import com.xilinx.rapidwright.device.IntentCode;
 import com.xilinx.rapidwright.device.Node;
 import com.xilinx.rapidwright.device.PIP;
@@ -65,13 +66,17 @@ public class VersalClockRouting {
     private static final EnumSet<IntentCode> allRouteTypes;
     
     static {
-        hRouteTypes = EnumSet.of(IntentCode.NODE_GLOBAL_HROUTE,
+        hRouteTypes = EnumSet.of(
+                IntentCode.NODE_GLOBAL_HROUTE,
                 IntentCode.NODE_GLOBAL_HROUTE_HSR, 
                 IntentCode.NODE_GLOBAL_HROUTE_MED, 
                 IntentCode.NODE_GLOBAL_HROUTE_SLOW); 
         vRouteTypes = EnumSet.of(IntentCode.NODE_GLOBAL_VROUTE);
-        allRouteTypes = EnumSet.of(IntentCode.NODE_GLOBAL_VROUTE, IntentCode.NODE_GLOBAL_HROUTE,
-                IntentCode.NODE_GLOBAL_HROUTE_HSR, IntentCode.NODE_GLOBAL_HROUTE_MED,
+        allRouteTypes = EnumSet.of(
+                IntentCode.NODE_GLOBAL_VROUTE, 
+                IntentCode.NODE_GLOBAL_HROUTE,
+                IntentCode.NODE_GLOBAL_HROUTE_HSR, 
+                IntentCode.NODE_GLOBAL_HROUTE_MED,
                 IntentCode.NODE_GLOBAL_HROUTE_SLOW);
     }
     
@@ -232,7 +237,8 @@ public class VersalClockRouting {
                 IntentCode.NODE_GLOBAL_VDISTR,
                 IntentCode.NODE_GLOBAL_VDISTR_LVL1, 
                 IntentCode.NODE_GLOBAL_VDISTR_LVL2,
-                IntentCode.NODE_GLOBAL_VDISTR_LVL21, 
+                IntentCode.NODE_GLOBAL_VDISTR_LVL21,
+                IntentCode.NODE_GLOBAL_VDISTR_LVL3, 
                 IntentCode.NODE_GLOBAL_GCLK);
 
         Set<NodeWithPrevAndCost> startingPoints = new HashSet<>();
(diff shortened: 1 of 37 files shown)


src/com/xilinx/rapidwright/router/VersalClockRouting.java
```diff
@@ -102,65 +127,86 @@ public static Node routeBUFGToNearestRoutingTrack(Net clk) {
...
+    
+    public static Integer getRoutingTrackIndex(Node node) {
+        if (node.getIntentCode() == IntentCode.NODE_GLOBAL_VROUTE) {
+            String wireName = node.getWireName();
+            return Integer.parseInt(wireName.substring(wireName.indexOf(VROUTE) + VROUTE.length()));
+        }
+        return null;
+    }
+    
+    private static boolean isCentroidCandidateAvailable(Node node, Set<Integer> unavailableTracks) {
```
eddieh-xlnx, line 140: Is it insufficient to use `getNodeStatus()` to identify unavailable tracks? As I understand, after routing one clock, all its nodes gets preserved and thus blocked for use by anyone else.

src/com/xilinx/rapidwright/router/VersalClockRouting.java
```diff
@@ -261,18 +347,16 @@ public static Map<ClockRegion, Node> routeVerticalToHorizontalDistributionLines(
...
                 if (targetCR.equals(curr.getTile().getClockRegion()) &&
                     curr.getIntentCode() == IntentCode.NODE_GLOBAL_GCLK &&
                     parent.getIntentCode() == IntentCode.NODE_GLOBAL_HDISTR_LOCAL) {
                     List<Node> path = curr.getPrevPath();
                     for (int i = 1; i < path.size(); i++) {
                         Node node = path.get(i);
-                        NodeStatus status = getNodeStatus.apply(node);
-                        if (status == NodeStatus.INUSE) {
-                            break;
-                        }
```
eddieh-xlnx, line 274: I have a inkling that this code was for incremental clock routing. However -- assuming we have some -- incremental tests still seem to be passing. Maybe we need to run the ECO flow.

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -279,45 +301,141 @@ private static void symmetricClockRoutingVersal(Net clk, Device device, Function
...
+        Pair<Node, ClockRegion> centroidResult = findCentroid(clk, centroidHRouteNode, centroid,
+                noVrouteNeeded, getNodeStatus, unavailableTracks);
+        Node vroute = centroidResult.getFirst();
+        centroid = centroidResult.getSecond();
+        
+        Map<ClockRegion, Node> upDownDistLines = VersalClockRouting
+                .routeToHorizontalDistributionLines(clk, vroute, usedCRsAndNonLCBPinsTuple.getFirst(),
+                        false, getNodeStatus);
+
+        // Route non-LCB driven pins
```
eddieh-xlnx: Can you give an example of what a non-LCB pin is?

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -386,6 +504,17 @@ public static Map<RouteNode, List<SitePinInst>> getLCBPinMappings(List<SitePinIn
...
 
+    private static boolean centroidEligible(ClockRegion candidate) {
+        // Check if we can use this clock region
+        Tile lowerLeft = candidate.getLowerLeft();
+        Tile upperLeft = candidate.getUpperLeft();
+        int tileHeight = 0;
+        if (lowerLeft != null && upperLeft != null) {
+            tileHeight = Math.abs(lowerLeft.getRow() - upperLeft.getRow());
+        }
+        return tileHeight > 40;
```
eddieh-xlnx: Where does 40 come from?

src/com/xilinx/rapidwright/rwroute/RWRoute.java
```diff
@@ -228,6 +232,73 @@ public static void preprocess(Design design) {
         DesignTools.makePhysNetNamesConsistent(design);
         DesignTools.createPossiblePinsToStaticNets(design);
         DesignTools.createMissingSitePinInsts(design);
+        if (series == Series.Versal) {
+            updateVersalXPHYPinsForDMC(design);
+        }
+    }
+
+    private static void updateVersalXPHYPinsForDMC(Design design) {
```
eddieh-xlnx: Can this be moved out of RWRoute?

eddieh-xlnx (commented):
Let's run this against Sashimi first

src/com/xilinx/rapidwright/router/VersalClockRouting.java
clavin-xlnx: I was running into situations with multiple clocks that one clock would choose a track, and then another clock would choose the same track but in another clock region.  Then, when it came to the stage where the router had to route a clock region that would have both clocks, it would fail.  The clock tracks must be mutually exclusive.

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
clavin-xlnx: If the clock is driving an IOB offchip, it does not pass through an LCB.

src/com/xilinx/rapidwright/rwroute/RWRoute.java
clavin-xlnx: Relocated to DesignTools.

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
clavin-xlnx: This is the minimum height of a clock region that we would want to attempt using as a centroid candidate.  The problem is that there are a lot of CRs in Versal devices that have very small CRs that don't have the clocking resources available.

clavin-xlnx: I've updated the code to define 40 as a named static variable.

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
clavin-xlnx: I updated the comment to include an example.

src/com/xilinx/rapidwright/router/VersalClockRouting.java
clavin-xlnx: Not sure, since this is Versal-specific code, would this matter in the ECO flow?

PR #1286, opened 2025-09-19
[DesignTools] createCeClkOfRoutethruFFToVCC() to handle AND2B1L/OR2L

eddieh-xlnx:
`DesignTools.createCeClkOfRoutethruFFToVCC()` is responsible for examining the physical netlist to discover the presence of FFs configured as routethrus and ensuring that their `CE` pins are connected to VCC, and their `CLK` pins are connected to GND (which is supplied via inversion from VCC). This operation only applied to US/US+ but not to Versal, since Versal now has a mux to bypass the flop without needing routethru.

It turns out this same operation is also necessary for the `AND2B1L` and `OR2L` unisims (and perhaps more; these seem to be the only two officially supported ones according to [UG974 for US/US+](https://docs.amd.com/r/en-US/ug974-vivado-ultrascale-libraries) and [UG1344 for Versal](https://docs.amd.com/r/en-US/ug1344-versal-architecture-libraries))

Do we want to change the name of this method (with an alias to the old name)? (Answered: No)

Depends on https://github.com/Xilinx/RapidWright/pull/1285 passing first.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -3363,7 +3363,7 @@ public class DesignTools {
     }
 
     /**
-     * Examines a design for FFs configured as routethrus or as AND/OR functions, ensuring that the CE pins of
+     * Examines a design for FFs configured as routethrus or as AND2B1L/OR2L functions, ensuring that the CE pins of
      * the used BELs are connected via the correct intra-site routing to VCC and their CLK pins are connected to
      * GND (which is supplied via inversion from VCC).
      * @param design Design to be processed.
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -3363,7 +3363,7 @@ public class DesignTools {
     }
 
     /**
-     * Examines a design for FFs configured as routethrus or as AND2B1L/OR2L functions, ensuring that the CE pins of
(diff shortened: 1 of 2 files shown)


clavin-xlnx (approved):
Looks good to me.  I'm not sure if a name change is necessary unless you feel strongly we should.  Perhaps just add a Javadoc to the method that describes that AND2B1L and OR2L fall into the category as well.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3361,24 +3361,34 @@ public static void createPossiblePinsToStaticNets(Design design) {
         createCeSrRstPinsToVCC(design);
     }
 
+    /**
+     * Examines a design for FFs configured as routethrus or as AND/OR functions, ensuring that the CE pins of
```
clavin-xlnx:
```suggestion
     * Examines a design for FFs configured as routethrus or as AND2B1L/OR2L functions, ensuring that the CE pins of
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3361,24 +3361,34 @@ public static void createPossiblePinsToStaticNets(Design design) {
         createCeSrRstPinsToVCC(design);
     }
 
+    /**
+     * Examines a design for FFs configured as routethrus or as AND2B1L/OR2L functions, ensuring that the CE pins of
```
eddieh-xlnx:
```suggestion
     * Examines a design for FFs configured as routethrus or as AND2B1L/OR2L functionality, ensuring that the CE pins of
```

PR #1293, opened 2025-10-03
Add method to prevent Vivado from grouping together certain ports into buses

abutt-amd:
Vivado incorrectly groups ports of the form `port[0]`, `port[1]`, etc into a single bus of ports of the form `port[1:0]`. This prevents the interface from matching with a desired blackbox in the netlist. The new method `ensurePreservedInferfaceVivado` appends `[]` in front of ports of this form which instructs Vivado to keep these port separate. A new method `renamePort` was added to `EDIFCell` to make this possible.


src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -326,6 +326,37 @@ public EDIFPort getPortByPortInstName(String portInstName) {
...
     }
 
+    /**
+     * Rename port with name currName to newName.
+     *
+     * @param currName The current name of the port that will be renamed.
+     * @param newName The new name of the port after renaming.
+     */
+    public void renamePort(String currName, String newName) {
+        EDIFPort port = getPortByPortInstName(currName);
```
clavin-xlnx, line 340: It is possible somewhere, someone might accidentally be calling this method (a lot) where the newName is the same as currName.  It might be worth it to just quickly do a String.equals() on the two before hand to quickly exit out and return to avoid that work.

src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -326,6 +326,37 @@ public EDIFPort getPortByPortInstName(String portInstName) {
...
+
+        if (port == null) {
+            throw new RuntimeException("Attempting to rename port " + currName
+                    + " that does not exist on cell " + getName());
+        }
+
+        EDIFPort newPort = new EDIFPort(newName, port.getDirection(), port.getWidth());
+        addPort(newPort);
+
+        for (int i = 0; i < port.getWidth(); i++) {
```
clavin-xlnx: Some bussed ports don't always start at zero. For example, you could have `my_signal[5:2]`.  There is a helper method for this scenario `EDIFPort.getBitBlastedIndicies()` which could be used.  This could also probably be added to your renamePort test below.

src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -326,6 +326,37 @@ public EDIFPort getPortByPortInstName(String portInstName) {
...
+            if (portInst == null) {
+                continue;
+            }
+            EDIFNet net = removeInternalPortMapEntry(portInst.getFullName());
+            net.removePortInst(portInst);
+            net.createPortInst(newPort, newPort.isBus() ? i : -1);
+            addInternalPortMapEntry(newPort.getPortInstNameFromPort(i), net);
+        }
+
+        removePort(port);
```
clavin-xlnx:
```suggestion
        removePort(port);
```

src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -1757,4 +1757,23 @@ public static EDIFNetlist createFlatNetlist(EDIFNetlist netlist, String partName
...
+     * Prevents vivado from forming buses out of ports of the form `port[0]`, `port[1]`, etc.
+     *
+     * @param netlist  The netlist to preserve the interface of.
+     *
+     */
+    public static void ensurePreservedInterfaceVivado(EDIFNetlist netlist) {
+        EDIFCell top = netlist.getTopCell();
+        List<String> portsToRename = new ArrayList<>();
+        for (EDIFPort p : top.getPorts()) {
+            if (!p.isBus() && !p.getName().startsWith("[]") && p.getName().endsWith("]")) {
```
clavin-xlnx: Let's create a `public static final String VIVADO_PRESERVE_PORT_INTERFACE = "[]";` or similar and add a Javadoc about why we are using this specific String.  Most of us have bad memories and it would be good to document this behavior for our future selves.

src/com/xilinx/rapidwright/util/VivadoTools.java
```diff
@@ -473,4 +473,42 @@ public static float getWorstSetupSlack(Path dcp, Path workdir, boolean encrypted
         }
         return Float.NaN;
     }
+
+    /**
+     * Open a netlist in Vivado and write a new DCP.
```
clavin-xlnx:
```suggestion
     * Open a DCP in Vivado and write a new DCP.
```

src/com/xilinx/rapidwright/util/VivadoTools.java
```diff
@@ -473,4 +473,42 @@ public static float getWorstSetupSlack(Path dcp, Path workdir, boolean encrypted
...
+        sb.append(createTclDCPLoadCommand(dcp, encrypted));
+        sb.append("write_checkpoint ").append(outputDcp).append(" -force; ");
+        sb.append("write_edif ").append(outputEdif).append(" -force; ");
+
+        VivadoTools.runTcl(outputLog, sb.toString(), true);
+        return outputDcp;
+    }
+
+    /**
+     * Open a netlist in Vivado then write a new DCP and parse the new DCP back into a Design.
```
clavin-xlnx:
```suggestion
     * Exports the design to Vivado and then writes a new DCP from Vivado, loads and returns it.
```

src/com/xilinx/rapidwright/util/VivadoTools.java
```diff
@@ -473,4 +473,42 @@ public static float getWorstSetupSlack(Path dcp, Path workdir, boolean encrypted
...
+    /**
+     * Open a netlist in Vivado and write a new DCP.
+     *
+     * @param dcp        Path to DCP to open in Vivado.
+     * @param outputName Name of the output design.
+     * @param workdir    Directory to work within.
+     * @param encrypted  Indicates whether DCP contains encrypted EDIF cells.
+     * @return Path to output dcp after round-trip.
+     */
+    public static Path roundTripDCPInVivado(Path dcp, String outputName, Path workdir, boolean encrypted) {
```
clavin-xlnx: What do you think about refactoring the name to something like `roundTripDCPThruVivado()`?

src/com/xilinx/rapidwright/util/VivadoTools.java
```diff
@@ -473,4 +473,42 @@ public static float getWorstSetupSlack(Path dcp, Path workdir, boolean encrypted
...
+    }
+
+    /**
+     * Open a netlist in Vivado then write a new DCP and parse the new DCP back into a Design.
+     *
+     * @param design  The design to open in Vivado.
+     * @param workdir Directory to work within.
+     * @return True if the port exists in Vivado, false otherwise.
+     */
+    public static Design roundTripDCPInVivado(Design design, Path workdir) {
```
clavin-xlnx: Refactor name? (see comment above)

src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -1757,4 +1757,23 @@ public static EDIFNetlist createFlatNetlist(EDIFNetlist netlist, String partName
...
+                portsToRename.add(p.getName());
+            }
+        }
+        for (String p : portsToRename) {
+            top.renamePort(p, "[]" + p);
```
clavin-xlnx: (see comment above about using a static string).

test/shared/com/xilinx/rapidwright/util/VivadoToolsHelper.java
```diff
@@ -68,4 +68,24 @@ public static void assertCanBeFullyPlacedAndRoutedByVivado(Design design, Path d
...
+        if (!FileTools.isVivadoOnPath()) {
+            return;
+        }
+        int beforePortCount = design.getNetlist().getTopCell().getPorts().size();
+        Design newDesign = VivadoTools.roundTripDCPInVivado(design, dir);
+        int afterPortCount = newDesign.getNetlist().getTopCell().getPorts().size();
+        Assertions.assertEquals(beforePortCount, afterPortCount);
+    }
+
+    public static void assertDifferentPortCountAfterRoundTripInVivado(Design design, Path dir) {
```
clavin-xlnx: These two methods are quite similar.  Could we simplify by creating a unified method that returns the number of top ports?  In this case the assertion could be moved out to the test.  Alternatively, you could have another parameter `boolean expectEquals` to simplify the two methods.

test/src/com/xilinx/rapidwright/edif/TestEDIFTools.java
```diff
@@ -426,4 +428,23 @@ public void testCopyEDNOnDCPWrite(@TempDir Path dir) {
...
+
+    @Test
+    public void testEnsurePreservedInterfaceVivado(@TempDir Path dir) {
+        Design design = RapidWrightDCP.loadDCP("bnn.dcp");
+        EDIFNetlist netlist = design.getNetlist();
+        EDIFCell topCell = netlist.getTopCell();
+
+        for (EDIFPort p : topCell.getPorts()) {
+            System.out.println(p);
+        }
```
clavin-xlnx: We can remove this for tests.

test/src/com/xilinx/rapidwright/edif/TestEDIFCell.java
```diff
@@ -117,4 +117,21 @@ public void testIsUniquifiedRemoveAndAdd() {
...
+    @Test
+    public void testRenamePort() {
+        Design design = RapidWrightDCP.loadDCP("bnn.dcp");
+        EDIFNetlist netlist = design.getNetlist();
+        EDIFCell topCell = netlist.getTopCell();
+
+        EDIFPort p = topCell.getPort("dmem_i_V_ce0");
+        topCell.renamePort("dmem_i_V_ce0", "test_port");
+
+        Assertions.assertNull(topCell.getPort("dmem_i_V_ce0"));
```
clavin-xlnx, line 129: Maybe consider a few other corner cases (if ports like this don't exist, you could always add them):
- `my_signal[5:2]`
- `my_signal[0:0]`
- `my_signal[5:5]`

src/com/xilinx/rapidwright/edif/EDIFCell.java
```diff
@@ -342,8 +346,12 @@ public void renamePort(String currName, String newName) {
 
         EDIFPort newPort = new EDIFPort(newName, port.getDirection(), port.getWidth());
         addPort(newPort);
+        int[] indices = {0};
+        if (port.isBus()) {
+            indices = port.getBitBlastedIndicies();
+        }
```
clavin-xlnx:
```suggestion
        int[] indices = port.isBus() ? port.getBitBlastedIndicies() : {0};
```

clavin-xlnx: This pattern avoids a branch instruction.

test/src/com/xilinx/rapidwright/edif/TestEDIFTools.java
```diff
@@ -435,16 +435,34 @@ public void testEnsurePreservedInterfaceVivado(@TempDir Path dir) {
...
+        EDIFTools.removeVivadoBusPreventionAnnotations(design.getNetlist());
+        EDIFPort testPort0 = topCell.getPort("test_port[0]");
+        Assertions.assertNotNull(testPort0);
+        Assertions.assertFalse(testPort0.getName().startsWith("[]"));
+        EDIFPort testPort1 = topCell.getPort("test_port[1]");
+        Assertions.assertNotNull(testPort1);
+        Assertions.assertFalse(testPort1.getName().startsWith("[]"));
+        EDIFPort testPort2 = topCell.getPort("test_port[2]");
+        Assertions.assertNotNull(testPort2);
+        Assertions.assertFalse(testPort2.getName().startsWith("[]"));
```
clavin-xlnx: It would probably be best to replace `"[]"` with references to `EDIFTools.VIVADO_PRESERVE_PORT_INTERFACE`, despite it being much more verbose.

PR #1294, opened 2025-10-07
Adds external library for netlists to compensate for black boxed cells

clavin-xlnx:
Often when IPs are included in a design, the corresponding EDIF representation is encrypted in a `.edn` file along side the EDIF.  This can present a challenge when attempting to place and/or route that design in RapidWright as it doesn't know what is inside the black boxed cells.  This allows for a substitute netlist with just the bare bones representation to be supplied externally to enable a place and or route solution to be achieved.  

The easiest way to invoke is to call:
```
EDIFNetlist.setExternalLibrary(EDIFLibrary external);
```
after a Design (DCP/EDIF) has been loaded.  This will automatically populate any black boxes with their corresponding substitute cells from the external library.  These cells will return to the black box state when `EDIFNetlist.collapseMacroUnisims()` is called (a default when writing out a DCP).  This is to ensure that when the DCP is written out, that Vivado can fully re-assemble the design from the encrypted and non-encrypted parts.  The substitute netlists cannot be part of the output.

Questions to answer:
 - The external library will undoubtedly contain references to primitive cell types which are only implicitly provided through `setExternalLibrary()`.  Do we want to change this?
   - Answer: No, keeping the primitive library implicit simplifies things.
 - Should we assume the state of the external library is with Unisim macros expanded or collapsed? 
   - Answer: Assume the external library is expanded upon call to `EDIFNetlist.setExternalLibrary(EDIFLibrary external)`


test/src/com/xilinx/rapidwright/rwroute/TestRWRoute.java
```diff
@@ -613,4 +616,96 @@ public void testDiscussion1245_20250807(boolean forceLagPin) {
...
+            VivadoTools.runTcl(outputLog, tclScript, true);
+
+            Design d = Design.readCheckpoint(dcp);
+            EDIFNetlist externalLib = EDIFTools.readEdifFile(subNetlist);
+            d.getNetlist().setExternalLibrary(externalLib.getHDIPrimitivesLibrary());
+
+            RWRoute.routeDesignFullNonTimingDriven(d);
+
+            d.getNetlist().blackBoxExternalCells();
+            d.getNetlist().setExternalLibrary(null);
```
eddieh-xlnx, line 705: Is this a necessity prior to writing out a DCP (which `VivadoToolsHelper.assertFullyRouted()` does behind the scene)? What badness happens if a user forgets to do this? Ideally, can it be done automatically?

test/src/com/xilinx/rapidwright/rwroute/TestRWRoute.java
```diff
@@ -613,4 +616,96 @@ public void testDiscussion1245_20250807(boolean forceLagPin) {
...
+            tclCmds.add("set cell_to_write [get_cells [lsort -unique [get_cells [get_property PARENT "
+                    + "[get_cells -hierarchical -filter {is_du_within_envelope==1}]]]] "
+                    + "-filter {is_du_within_envelope!=1}]");
+            tclCmds.add("write_cell_to_edif $cell_to_write " + subNetlist.toString());
+            FileTools.writeLinesToTextFile(tclCmds, tclScript.toString());
+            VivadoTools.runTcl(outputLog, tclScript, true);
+
+            Design d = Design.readCheckpoint(dcp);
+            EDIFNetlist externalLib = EDIFTools.readEdifFile(subNetlist);
+            d.getNetlist().setExternalLibrary(externalLib.getHDIPrimitivesLibrary());
```
eddieh-xlnx, line 700: I'm a bit confused by this: why would the now un-blackboxed cells appear in the HDIPrimitives library?

clavin-xlnx: The Tcl script simply puts all the relevant cells in the same library to keep things simple.  The external library is only one library, not a netlist.  Using the HDIPrimitives library is the easiest one-library option.

test/src/com/xilinx/rapidwright/rwroute/TestRWRoute.java
clavin-xlnx: Yes, this needs to happen for the design to be able to be loaded correctly back into Vivado.  If not, the EDIF will be written out with an inconsistent state and any encrypted cells won't be loaded back into their original cell locations.  We could potentially automate this for writing out a DCP, but it might be tricky and I think needs some thought to be robust.  For now I prefer we handle the situation manually.

PR #1299, opened 2025-10-17
[Write EDIF] Validate .edn files attached to netlist before DCP write

clavin-xlnx:
Sometimes if there are "*.edn" files in the same directory as the ".edf" file that is loaded with a DCP, unrelated .edn files can be mistakenly included in a design that has no encrypted cells.  This PR aims to avoid that issue by searching through the .edn files and making sure there are no matching black box cells that exist in the netlist.

### DIFF
--- a/test/src/com/xilinx/rapidwright/edif/TestEDIFTools.java
+++ b/test/src/com/xilinx/rapidwright/edif/TestEDIFTools.java
@@ -481,6 +481,8 @@ public class TestEDIFTools {
         FileTools.writeStringToTextFile("Fake EDIF", edn.toString());
 
         // We need to read an external EDIF in order to trigger the .edn search
+        // When provided with an external EDIF, RapidWright will look for encrypted cells in the same directory as the 
+        // .edn file, keeping a record of possible encrypted cells as it goes.  
         d = Design.readCheckpoint(dcp, edf);
 
         Assertions.assertEquals(1, d.getNetlist().getEncryptedCells().size());


test/src/com/xilinx/rapidwright/edif/TestEDIFTools.java
```diff
@@ -465,4 +466,28 @@ public void testRemoveVivadoBusPreventionAnnotations(@TempDir Path dir) {
...
+        d.writeCheckpoint(dcp);
+        // All our test DCPs have unencrypted EDIF inside, we need to create an external
+        // one
+        d.getNetlist().exportEDIF(edf);
+        FileTools.writeStringToTextFile("Fake EDIF", edn.toString());
+
+        // We need to read an external EDIF in order to trigger the .edn search
+        d = Design.readCheckpoint(dcp, edf);
+
+        Assertions.assertEquals(1, d.getNetlist().getEncryptedCells().size());
```
eddieh-xlnx, line 488: If I'm understanding correctly: `readCheckpoint` (will continue to) pull all `*.edn` files in the same directory as the `*.dcp` -- regardless of whether they're needed or not (and in this case, absolutely not). Perhaps that's worth a comment here as it took a minute to parse.

clavin-xlnx, line 488:
```suggestion
        // We need to read an external EDIF in order to trigger the .edn search
        // When provided with an external EDIF, RapidWright will look for encrypted cells in the same directory as the 
        // .edn file, keeping a record of possible encrypted cells as it goes.  
        d = Design.readCheckpoint(dcp, edf);

        Assertions.assertEquals(1, d.getNetlist().getEncryptedCells().size());
```

clavin-xlnx:
> > This PR only affects `writeCheckpoint()`. Is an equivalent optimization to `readCheckpoint()` relevant?
> 
> No. Since encrypted cells can contain other encrypted cells, we can only make the assumption that there are either no encrypted cells or some encrypted cells.

It could be argued that we could have this check after loading a DCP to search for matching black boxes in the netlist, but I prefer the lazy approach because if the user loads the DCP and never writes it out, then the check would be in vain and there would be no real change in behavior.

clavin-xlnx:
> > > This PR only affects `writeCheckpoint()`. Is an equivalent optimization to `readCheckpoint()` relevant?
> > 
> > 
> > No. Since encrypted cells can contain other encrypted cells, we can only make the assumption that there are either no encrypted cells or some encrypted cells.
> 
> That's a good point! I think that same argument apply to this `writeCheckpoint` PR too, and so your PR description is a bit misleading.
> 
> > ... unrelated .edn files can be included unnecessarily in the load script when written out.
> 
> whereas I think you mean the following?
> 
> "... unrelated .edn can cause a load script to be unnecessarily written out."

Thanks, updated.  That extra little revelation came out during testing.

PR #1300, opened 2025-10-18
[ECOTools] disconnectNet() to use skipUnrouteIntraSite property

eddieh-xlnx:
Since #1254, `ECOTools.disconnectNet()` now removes all intra-site routing.

This PR adds the property `rapidwright.ecotools.disconnectNet.skipUnrouteIntraSite` to disable this new behaviour which can be helpful for when `disconnectNet()` is followed by `connectNet()` that will re-use this intra-site routing.

### DIFF
--- a/test/src/com/xilinx/rapidwright/design/TestCell.java
+++ b/test/src/com/xilinx/rapidwright/design/TestCell.java
@@ -68,6 +68,29 @@ public class TestCell {
 
     @ParameterizedTest
     @CsvSource({
+            "false,[G5]",
+            "true,'[G5, G1, G2, G3, G4, G6]'",
+    })
+    public void testGetAllCorrespondingSitePinNamesLUTRouteThru(boolean considerLutRoutethru, String expectedSitePins) {
+        Design d = new Design("testGetAllCorrespondingSitePinNamesLUTRouteThru", Device.KCU105);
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/eco/ECOTools.java
```diff
@@ -117,10 +118,16 @@ public static void disconnectNet(Design design,
...
      *                         reused by {@link #connectNet(Design, Map, Map)}.
+     * By default, this method will unroute any intra-site routing associated with the disconnected
+     * pin. The Java property "rapidwright.ecotools.disconnectNet.skipUnrouteIntraSite" disables this
+     * behaviour which can be helpful for when disconnectNet() is followed by connectNet() that
+     * will re-use this intra-site routing.
      */
     public static void disconnectNet(Design design,
                                      List<EDIFHierPortInst> pins,
                                      Map<Net, Set<SitePinInst>> deferredRemovals) {
+        final boolean unrouteIntraSite = !Params.isParamSet("rapidwright.ecotools.disconnectNet.skipUnrouteIntraSite");
```
clavin-xlnx, line 129: I think as long as the param is only used in one place, I'm fine keeping it as a String literal.  As soon as it is used in two or more places, I think it should be defined as a static somewhere.

PR #1304, opened 2025-10-22
Add HierNet clock detection command line tool

mmatlin:
Adds a tool named `NetlistClockDetection` that finds all terminal hierarchical port insts (pins) which gate changes to the given hierarchical net's value, i.e., finds all terminal pins which may be drivers of the net through non-clocked primitives and all likely terminal clock signal pins that gate signals which drive the net through clocked primitives.

For example, when run on the output hierarchical net of a LUT2 whose two inputs come from a flop and a LUT, this will return the terminal hierarchical port inst(s) driving the flop's clock pin as well as any gating drivers of the input LUT's inputs, determined recursively. For the schematic below, the tool would return that the gating drivers/likely clocks leading to the hierarchical net `foo` are `clk`, `a`, `b`, and `c`.
```
         ____________
clk ----|C   FDRE    |
... ----|CE         Q|_
 in ----|D           | \       ____________
... ----|R           |  \     |    LUT2    |  (HierNet foo)
         ‾‾‾‾‾‾‾‾‾‾‾‾    \____|I0         O|----------------- out
         ____________    /‾‾‾‾|I1          |
        |    LUT3    |  /      ‾‾‾‾‾‾‾‾‾‾‾‾
  a ----|I0          | /
  b ----|I1         O|‾
  c ----|I2          |
         ‾‾‾‾‾‾‾‾‾‾‾‾
```
The CLI tool takes any number of hierarchical nets to find gating drivers of and the result is returned as a JSON object with input hierarchical net names as keys and arrays of their found gating driver pins as values. For the above schematic, if the tool was called with just `foo` as an input, it would return `{"foo": ["clk", "a", "b", "c"]}`.

Currently, only Versal devices are supported, and the list of Versal primitives' clock input pins was determined by retrieving the `IS_CLOCK` property in Vivado for all primitives' pins, which I'm not sure if is a fully accurate approach.


src/com/xilinx/rapidwright/design/tools/NetlistClockDetection.java
```diff
@@ -0,0 +1,180 @@
...
+import com.xilinx.rapidwright.edif.EDIFHierCellInst;
+import com.xilinx.rapidwright.edif.EDIFHierNet;
+import com.xilinx.rapidwright.edif.EDIFHierPortInst;
+import com.xilinx.rapidwright.edif.EDIFNetlist;
+import com.xilinx.rapidwright.edif.EDIFTools;
+
+public class NetlistClockDetection {
+
+    /** A map whose keys are FPGA series and whose values are maps from the names of
+     * those series' primitives which have clock input pins to those clock input pins.
```
clavin-xlnx:
```suggestion
    /** 
     * A map whose keys are FPGA series and whose values are maps from the names of
     * those series' primitives which have clock input pins to those clock input pins.
```
Just tidying up the Javadoc.

src/com/xilinx/rapidwright/design/tools/NetlistClockDetection.java
```diff
@@ -0,0 +1,180 @@
...
+
+import com.xilinx.rapidwright.design.Design;
+import com.xilinx.rapidwright.device.Series;
+import com.xilinx.rapidwright.edif.EDIFHierCellInst;
+import com.xilinx.rapidwright.edif.EDIFHierNet;
+import com.xilinx.rapidwright.edif.EDIFHierPortInst;
+import com.xilinx.rapidwright.edif.EDIFNetlist;
+import com.xilinx.rapidwright.edif.EDIFTools;
+
+public class NetlistClockDetection {
```
clavin-xlnx, line 50: It would be great if we could add a brief summary of what this class does in a Javadoc.

src/com/xilinx/rapidwright/edif/EDIFHierNet.java
```diff
@@ -284,6 +284,11 @@ public List<EDIFHierPortInst> getLeafHierPortInsts(boolean includeSourcePins, bo
                         if (upNet != null) {
                             queue.add(upPort.getHierarchicalNet());
                         }
+                    } else {
+                        // Add top-level hierarchical port insts
+                        if ((includeSinkPins && p.isOutput()) || (includeSourcePins && p.isInput())) {
+                            leafCellPins.add(p);
+                        }
```
clavin-xlnx:
```suggestion
                    } else if ((includeSinkPins && p.isOutput()) || (includeSourcePins && p.isInput())) {
                        // Add top-level hierarchical port insts
                        leafCellPins.add(p);
```
Simplifying.  It would be great to have a test that demonstrates how this change modifies the behavior of this method.  See https://github.com/Xilinx/RapidWright/blob/e166aebab504d5080945a022d15478dcb4f59438/test/src/com/xilinx/rapidwright/edif/TestEDIFHierNet.java#L44 for an example.

src/com/xilinx/rapidwright/design/tools/NetlistClockDetection.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        EDIFNetlist nl = args[0].endsWith(".dcp") ? Design.readCheckpoint(args[0]).getNetlist() : EDIFTools.readEdifFile(args[0]);
+
+        Map<String, Set<String>> primsToClockPins = seriesPrimsToClockPins.get(nl.getDevice().getSeries());
+
+        if (primsToClockPins == null) {
+            throw new RuntimeException("ERROR: NetlistClockDetection only supports likely clock pin detection for Versal-targeting netlists. "
+                + "The provided netlist targets " + nl.getDevice().getSeries() + ".");
+        }
```
clavin-xlnx: Perhaps we could move this into its own helper method, making the `seriesPrimsToClockPins` map `private` and have a method like: `public static Map<String, Set<String>> getPrimsToClockPins(Series series)`.   That way, if user want to call `getGatingDrivers()` from another context, they could benefit from the `Series` check.

Alternatively, you change the method signature of `getGatingDrivers(EDIFHierNet hNet, Map<String, Set<String>> primsToClockPins, ...` to `getGatingDrivers(EDIFHierNet hNet, Series series,...` and hide all of the map lookup/checking.

src/com/xilinx/rapidwright/design/tools/NetlistClockDetection.java
```diff
@@ -0,0 +1,180 @@
...
+                } else {
+                    out.add(hPI);
+                }
+            }
+        }
+
+        return out;
+    }
+
+    static Set<EDIFHierPortInst> getGatingDrivers(EDIFHierNet hNet, Map<String, Set<String>> primsToClockPins) {
```
clavin-xlnx: We should probably get a test added to our regressions so that if we change something that breaks it, we can catch it in CI.  Since this is a new file, the corresponding test would go in `$RAPIDWRIGHT_PATH/test/src/com/xilinx/rapidwright/design/tools/TestNetlistClockDetection.java`.  We use JUnit for tests.  There are several examples in the test area, you can also check our docs (https://www.rapidwright.io/docs/Setup_JUnit5_Test_in_RapidWright.html) for help on getting them setup.  Reach out to me directly if you have any trouble.

clavin-xlnx:
```suggestion
    public static Set<EDIFHierPortInst> getGatingDrivers(EDIFHierNet hNet, Map<String, Set<String>> primsToClockPins) {
```
Seems like we should make this available publicly?

src/com/xilinx/rapidwright/design/tools/NetlistClockDetection.java
```diff
@@ -0,0 +1,180 @@
...
+     * inputs, determined recursively.
+     *
+     * @param hNet             The hierarchical net to find gating drivers of.
+     * @param primsToClockPins A map from primitive names to their input clock pins, if
+     *                         the primitive does have input clock pins. This is
+     *                         specific to the FPGA series.
+     * @param encountered      A set of already-encountered EDIFHierNets to skip during
+     *                         netlist traversal.
+     */
+    static Set<EDIFHierPortInst> getGatingDrivers(EDIFHierNet hNet, Map<String, Set<String>> primsToClockPins, Set<EDIFHierNet> encountered) {
```
clavin-xlnx:
```suggestion
    private static Set<EDIFHierPortInst> getGatingDrivers(EDIFHierNet hNet, Map<String, Set<String>> primsToClockPins, Set<EDIFHierNet> encountered) {
```
I'm assuming we don't want outside users calling this one?

clavin-xlnx (commented):
@mmatlin - thanks for the PR!  Only some minor edits and refactoring needed.  Although, it looks like a few tests failed in our CI.  To check what failed and see the stack traces, you'll want to click on the failing build:

image

Click on "Summary":
image

Download and extract the "Java Test Report multi-threaded" locally:
image

Open index.html in the extracted folder in a browser and it should show failing tests:

image

src/com/xilinx/rapidwright/design/tools/NetlistClockDetection.java
mmatlin: Yep, that's probably right, in which case would you prefer that I include a separate Javadoc for the overloaded version that doesn't take `Set<EDIFHierNet> encountered`?

clavin-xlnx: Sure, I think refactoring the Javadoc for the public facing method makes sense.

test/src/com/xilinx/rapidwright/design/tools/TestNetlistClockDetection.java
```diff
@@ -0,0 +1,394 @@
...
+
+import com.xilinx.rapidwright.support.RapidWrightDCP;
+
+public class TestNetlistClockDetection {
+
+    @Test
+    public void testNetlistClockDetection(@TempDir Path dir) {
+        Path path = RapidWrightDCP.getPath("test_clock_detection.edif");
+
+        String[] args = new String[] {
```
clavin-xlnx: Let's select a meaningful subset (~5-15 examples) to reduce this test to something that could be viewed as more of an example of how to call `NetlistClockDetection`.  If there are too many corner cases that we can't cover, that is fine, maybe break the test up into pieces.  I think we just want to have a test that can be smaller so it is easier to maintain.

One idea to simplify is to create a map with the arguments as keys and the expected results as values.  Then you could copy the keys of the map into the `String[] args`.  This would allow you to have the expected map upfront and you could check the values against the result in the loop.

mmatlin: Agreed that it was too much, I've pushed a commit that narrows down the test to 10 individual cases.

PR #1307, opened 2025-10-31
[ConstraintTools] Tools to handle the xdc constraint on pblock

coherent17:
As discussed in discussion(#1235), a method to get the pblock from XDCConstraint.
When the tcl intepreter is ready (#1265), several xdc constraints can be further added in this class.

### DIFF
--- a/src/com/xilinx/rapidwright/design/xdc/ConstraintTools.java
+++ b/src/com/xilinx/rapidwright/design/xdc/ConstraintTools.java
@@ -35,7 +35,7 @@ import com.xilinx.rapidwright.design.ConstraintGroup;
  * Created on: Oct 31, 2025
  */
 public class ConstraintTools {
-    public static Map<String, PBlock> getPBlockFromXDCConstraints(Design d){
+    public static Map<String, PBlock> getPBlockFromXDCConstraints(Design d) {
         Map<String, PBlock> pblockMap = new HashMap<>();
         for (ConstraintGroup cg : ConstraintGroup.values()) {
             for (String tclLine : d.getXDCConstraints(cg)) {


src/com/xilinx/rapidwright/design/XDC.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.xilinx.rapidwright.design;
```
clavin-xlnx: I think we are going to try and add a new package for the XDC/Tcl parsing stuff in #1265 (see my recent comment https://github.com/Xilinx/RapidWright/pull/1265#issuecomment-3474160150).  Could we refactor this class so it will be in the `com.xilinx.rapidwright.design.xdc` package?

src/com/xilinx/rapidwright/design/ConstraintsTools.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public static PBlock getPBlock(String name) {
+        return pblockMap.get(name);
+    }
+
+    public static HashMap<String, PBlock> getPBlockMap() {
+        return pblockMap;
+    }
+}
```
clavin-xlnx: Let's add a newline at the end of the file

test/src/com/xilinx/rapidwright/design/TestXDC.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.xilinx.rapidwright.design;
```
clavin-xlnx: Similarly, since we'll refactor the `ConstraintTools`, we should also put the corresponding test in `com.xilinx.rapidwright.design.xdc`.

test/src/com/xilinx/rapidwright/design/TestConstraintsTools.java
```diff
@@ -0,0 +1,41 @@
...
+    public void testParsePBlockFromXDCConstraints() {
+        String dcpPath = RapidWrightDCP.getString("microblazeAndILA_3pblocks.dcp");
+        Design d = Design.readCheckpoint(dcpPath);
+        ConstraintsTools ct = new ConstraintsTools(d);
+        Assertions.assertEquals(3, ct.getPBlockMap().size());
+        Assertions.assertTrue(ct.getPBlockMap().containsKey("pblock_dbg_hub"));
+        Assertions.assertTrue(ct.getPBlockMap().containsKey("pblock_base_mb_i"));
+        Assertions.assertTrue(ct.getPBlockMap().containsKey("pblock_u_ila_0"));
+    }
+}
```
clavin-xlnx: Let's add a newline at the end of the file

src/com/xilinx/rapidwright/design/ConstraintsTools.java
```diff
@@ -0,0 +1,65 @@
...
+public class ConstraintsTools {
+    private static HashMap<String, PBlock> pblockMap = new HashMap<>();
+    private Design d;
+
+    public ConstraintsTools(Design d){
+        this.d = d;
+        parsePBlockFromXDCConstraints();
+    }
+
+    private void parsePBlockFromXDCConstraints(){
```
clavin-xlnx: Sorry, a few more comments.  Could we change this method so that it is `static` and takes as input a `Design` object and returns the `Map<String, PBlock>`?  That would follow the pattern of the `*Tool` classes.  We typically don't want to instantiate classes like this that are a collection of tools.  To be clear, this would be the new signature:
```
public static Map<String, PBlock> parsePBlocksFromXDC(Design design) 
```
Then you could remove the constructors and other methods.

src/com/xilinx/rapidwright/design/XDC.java
coherent17: Great, looking forward to handle those design constraints soon in RapidWright, this would be really helpful! I'd add a XDC.java under design folder.

src/com/xilinx/rapidwright/design/ConstraintsTools.java
coherent17: I'd removed the constructor, and pass design as the argument, thanks

src/com/xilinx/rapidwright/design/XDC.java
clavin-xlnx: Sorry if I was unclear, I thought the class name you had was fine, what I meant was to relocate the file to the new package `com.xilinx.rapidwright.design.xdc`.  We don't have any files there yet, so you'll have to create that directory.  So, the expected file would be `src/com/xilinx/rapidwright/design/xdc/ConstraintTools.java`.  The package name is generally the directory path in the project and the class name is the root file name in Java.

test/src/com/xilinx/rapidwright/design/TestXDC.java
clavin-xlnx: Similarly, this file would be `test/src/com/xilinx/rapidwright/design/xdc/TestConstraintTools.java`.

src/com/xilinx/rapidwright/design/XDC.java
coherent17: Sorry I misunderstanding the project name and the class name. I'd updated the file to the package   ```com.xilinx.rapidwright.design.xdc```.

src/com/xilinx/rapidwright/design/xdc/ConstraintTools.java
```diff
@@ -0,0 +1,54 @@
...
+import com.xilinx.rapidwright.design.blocks.PBlock;
+import com.xilinx.rapidwright.design.ConstraintGroup;
+
+/**
+ * A collection of methods to access design constraints.
+ *
+ * Created on: Oct 31, 2025
+ */
+public class ConstraintTools {
+    public Map<String, PBlock> getPBlockFromXDCConstraints(Design d){
```
clavin-xlnx:
```suggestion
    public static Map<String, PBlock> getPBlockFromXDCConstraints(Design d) {
```
We want to make this method static so that users don't have to create an instance of `ConstraintTools`, when the method is static, the method can be called directly `ConstraintTools.getPBlockFromXDCConstraints()`.

coherent17: Got it, thanks for the suggestion. I've updated the ```static``` keyword for the method.

src/com/xilinx/rapidwright/design/xdc/ConstraintTools.java
```diff
@@ -0,0 +1,54 @@
...
+import com.xilinx.rapidwright.design.blocks.PBlock;
+import com.xilinx.rapidwright.design.ConstraintGroup;
+
+/**
+ * A collection of methods to access design constraints.
+ *
+ * Created on: Oct 31, 2025
+ */
+public class ConstraintTools {
+    public static Map<String, PBlock> getPBlockFromXDCConstraints(Design d){
```
clavin-xlnx:
```suggestion
    public static Map<String, PBlock> getPBlockFromXDCConstraints(Design d) {
```
Sorry, one more nit -- we have a space between the ')' and '{' at the end of most Java constructs.

coherent17: Will RapidWright provide some formating style for java in the future?
Ex: clang-format or other formatting tools.

clavin-xlnx: That is a good point, we should have a formatter solution but we don't.  We mostly follow the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html), but I think there might be a few places where we deviate (see https://github.com/Xilinx/RapidWright/blob/master/CODING_STYLE.md).  Ideally we would have a solution that was fully automated and had plug-ins for the major IDEs (IntelliJ, Eclipse, VSCode, etc).

PR #1306, opened 2025-10-30
Add or-tools 9.14.6206, update protobuf to 4.31.1, update commons-io to 2.20.0

clavin-xlnx:
Adds and updates the following dependencies: 

ADD com.google.ortools:ortools-java:9.14.6206
ADD com.google.ortools:ortools-linux-x86-64:9.14.6206
ADD com.google.ortools:ortools-darwin-x86-64:9.14.6206
ADD com.google.ortools:ortools-win32-x86-64:9.14.6206
ADD com.google.ortools:ortools-linux-aarch64:9.14.6206
ADD com.google.ortools:ortools-darwin-aarch64:9.14.6206
ADD net.java.dev.jna:jna-platform:5.14.0
ADD net.java.dev.jna:jna:5.14.0
UPGRADE com.google.protobuf:protobuf-java:4.31.1 (from 3.25.0)
UPGRADE commons-io:commons-io:2.20.0 (from 2.11.0)

Also includes a minor improvement to the `DesignComparator`.

PR #1309, opened 2025-10-31
Array Builder Automated Placement and Various Fixes

abutt-amd:
Adds automated module placement to array builder and fixes various other bugs, as well as improving performance. Dependent on #1306 getting merged and then array_builder getting updated. Still needs a lot of cleanup, but opening as a draft for now, now that or-tools is getting added. I might end up splitting this into at least 2 separate PRs (one for array builder changes and one for performance explorer / inline flop tools).


clavin-xlnx:
Yes, let's split it into two PRs.  It also looks like some of the commits will need to be augmented with the sign off message.

PR #1311, opened 2025-11-04
[VersalClockRouting] VDISTR tree storage map for Versal

clavin-xlnx:
Versal has an improved feature over UltraScale/UltraScale+ clock networks for vertical clock distribution (Vertical H tree -- see slide 6 of https://www.amd.com/content/dam/amd/en/documents/products/adaptive-socs-and-fpgas/vivado/reducing-clock-skew-in-versal-devices.pdf).  
image

In order to take advantage of this, there are certain vertical distribution path patterns that should be followed to minimize skew and balance multiple occupied clock regions in a clock domain.  

This PR updates the Versal Clock Router employed by RWRoute to load a new RapidWright data file that contains a map with vertical distribution patterns pre-computed offline for all Versal devices.  There are entries for every contiguous span of occupied logic (Y1->Y1, Y1->Y2, Y1->Y3, ... Y1->Y10), for each valid clock root option for the given contiguous span, with a preferred clock root being used all of the time.

Note: The Calibrated Deskew feature described in the document above is not something we intend to support at this time.  If users wish to take advantage of this feature in their projects, they will need to import their designs back into Vivado.  

Ultimately, this modification led to a reduction of 100's of picoseconds of skew for a multi-SLR device.

### DIFF
--- a/src/com/xilinx/rapidwright/util/DataVersions.java
+++ b/src/com/xilinx/rapidwright/util/DataVersions.java
@@ -395,6 +395,6 @@ public class DataVersions {
         dataVersionMap.put("data/partdump.csv", new Pair<>("partdump-csv", "7e74868f1d6f8a1b3e4084097ef72d1b"));
         dataVersionMap.put("data/parts.db", new Pair<>("parts-db", "516e1d369674451dc29498a99d7f00f5"));
         dataVersionMap.put("data/unisim_data.dat", new Pair<>("unisim-data-dat", "a13bffcefb95e410acb4ce02cab2399e"));
-        dataVersionMap.put("data/versal_vdistr_trees.dat", new Pair<>("versal-vdistr-trees-dat", "27d6515bc6a12e0b258b5be1f7095abc"));
+        dataVersionMap.put("data/versal_vdistr_trees.dat", new Pair<>("versal-vdistr-trees-dat", "3f31320ae45277511538d37cc7127309"));
     }
 }


src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -547,6 +567,7 @@ private static ClockRegion findCentroid(Net clk, Device device) {
                 centroid = centroid.getNeighborClockRegion(-1, 0);
             }
         }
+
```
clavin-xlnx, line 570:
```suggestion
```

src/com/xilinx/rapidwright/rwroute/GlobalSignalRouting.java
```diff
@@ -373,28 +375,46 @@ private static void symmetricClockRoutingVersal(Net clk, Device device,
...
+            maxY = Math.max(maxY, cr.getInstanceY());
+        }
+        Device device = origCentroid.getDevice();
+        // Clock roots on Versal appear to only be possible on odd-numbered columns
+        int clkRootXCoord = origCentroid.getColumn() % 2 == 0 ? origCentroid.getInstanceX() + 1
+                : origCentroid.getInstanceX();
+        Integer preferredYCoord = VersalClockRouting.getPreferredClockRootYCoord(device, minY, maxY);
+        int clkRootYCoord = preferredYCoord == null ? origCentroid.getInstanceY() : preferredYCoord;
+
+        // If the current column doesn't work, try both neighboring ones
```
clavin-xlnx:
```suggestion
        // If the current column doesn't work, try neighboring ones on both sides
```

clavin-xlnx:
> > Note: The Calibrated Deskew feature described in the document above is not something we intend to support at this time. If users wish to take advantage of this feature in their projects, they will need to import their designs back into Vivado.
> 
> From your [slides](https://www.amd.com/content/dam/amd/en/documents/products/adaptive-socs-and-fpgas/vivado/reducing-clock-skew-in-versal-devices.pdf) it seems there is "basic deskew" (sufficient for intra-SLR) and "calibrated deskew" (for multiple SLR devices). Are you saying we support the former?

Yes, just basic deskew using the H-trees for horizontal distribution.

src/com/xilinx/rapidwright/router/VersalClockRouting.java
```diff
@@ -282,46 +287,54 @@ public static Map<ClockRegion, Node> routeVrouteToVerticalDistributionLines(Net
             verticalSpineCRs.add(device.getClockRegion(i, x));
         }
 
-        nextClockRegion: for (ClockRegion cr : verticalSpineCRs) {
+        VersalClockTree clkTree = getVersalClockTree(device, minY, maxY);
+        if (clkTree == null) {
+            System.err.println("ERROR: No clock tree found for " + device + " Y" + minY + "-Y" + maxY
```
eddieh-xlnx, line 293: Is the plan (in a future PR, prior to final 2025.2.0 release) to support more devices than just the V80 that the data file in this PR supports? If so, should this be turned into a Exception (after which any code to do suboptimal vertical routing can also be removed?)

eddieh-xlnx (approved):
Is it possible to get a quick test to ensure that the balanced H-tree does indeed get taken, as opposed to the suboptimal routing before?

src/com/xilinx/rapidwright/router/VersalClockRouting.java
clavin-xlnx: I've pushed a file that supports all Versal devices (except xcvp1902).  I kept the exception but added an assert so any tests will fail if the data isn't found.  I've also added a test that checks the vertical distribution data is being used in the clock router.

PR #1316, opened 2025-11-07
Parallelize Router Preprocessing

abutt-amd:
For some large designs router preprocessing takes longer than actually routing the design. This PR parallelizes the bottlenecks of router preprocessing where possible. Before these changes preprocess takes 243 seconds on a design that nearly fills 2 SLRs on the V80, and after these changes preprocess takes 74 seconds (a 3.3x speedup). Performance comparison run on my VDI with 8 threads.

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -2299,6 +2299,7 @@ public class DesignTools {
             if (bel == null) continue;
             String logicalPinName = p.getPortInst().getName();
             Set<String> physPinMappings;
+            // Need to synchronize on the cell to prevent concurrent modification of the logical to physical pin map
             synchronized (c) {
                 physPinMappings = c.getAllPhysicalPinMappings(logicalPinName);
             }


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2204,6 +2210,18 @@ public static boolean placeCell(Cell c, Design design) {
...
     public static List<SitePinInst> createMissingSitePinInsts(Design design, Net net) {
+        return createMissingSitePinInsts(design, net, null);
+    }
+
+    /**
+     * Creates any and all missing SitePinInsts for this net.  This is common as a placed
+     * DCP will not have SitePinInsts annotated and this information is generally necessary
+     * for routing to take place.
+     * @param design The current design of this net.
+     * @return The list of pins that were created or an empty list if none were created.
```
clavin-xlnx, line 2223: It looks like the previous version of this method was missing the Javadoc for the parameter `net`.  Would you mind updating the previous version and this version as well as documenting the new `siteInstToNetSiteWiresMap`?  I think it would be good to explain why this map should be added and how it contributes to the speedup.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2338,10 +2366,41 @@ public static String getRoutedSitePin(Cell cell, Net net, String logicalPinName)
...
+     * @return The name of the first site pin on the cell's site to which the pin is routed.
+     */
+    public static String getRoutedSitePinFromPhysicalPin(Cell cell, Net net, String belPinName,
+                                                         Map<SiteInst, Map<Net, List<String>>> siteInstToNetSiteWiresMap) {
+        List<String> sitePins = getAllRoutedSitePinsFromPhysicalPin(cell, net, belPinName, siteInstToNetSiteWiresMap);
         return (!sitePins.isEmpty()) ? sitePins.get(0) : null;
     }
 
+    public static Map<SiteInst, Map<Net, List<String>>> getSiteInstToNetSiteWiresMap(Design design) {
+        ConcurrentMap<SiteInst, Map<Net, List<String>>> siteInstToNetSiteWiresMap = new ConcurrentHashMap<>();
```
clavin-xlnx:
```suggestion
        Map<SiteInst, Map<Net, List<String>>> siteInstToNetSiteWiresMap = new ConcurrentHashMap<>();
```
A common pattern we try to follow is to use the interface type as the reference for common data structures.  This enables a bit more flexibility in other potential contexts.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2338,10 +2366,41 @@ public static String getRoutedSitePin(Cell cell, Net net, String logicalPinName)
...
+     * @param belPinName The physical pin name of the cell
+     * @return The name of the first site pin on the cell's site to which the pin is routed.
+     */
+    public static String getRoutedSitePinFromPhysicalPin(Cell cell, Net net, String belPinName,
+                                                         Map<SiteInst, Map<Net, List<String>>> siteInstToNetSiteWiresMap) {
+        List<String> sitePins = getAllRoutedSitePinsFromPhysicalPin(cell, net, belPinName, siteInstToNetSiteWiresMap);
         return (!sitePins.isEmpty()) ? sitePins.get(0) : null;
     }
 
+    public static Map<SiteInst, Map<Net, List<String>>> getSiteInstToNetSiteWiresMap(Design design) {
```
clavin-xlnx: This might be a good place to explain in a Javadoc the purpose of this map, and you can reference this method  in the above method parameter Javadoc description by using this syntax `{@link #getSiteInstToNetSiteWiresMap(Design)}`.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2352,15 +2411,35 @@ public static String getRoutedSitePinFromPhysicalPin(Cell cell, Net net, String
...
+    }
+
+    /**
+     * Gets all site pins that are currently routed to the specified cell pin.  If
+     * the site instance is not routed, it will return null.
+     * @param cell The cell with the pin of interest.
+     * @param net The physical net to which this pin belongs
+     * @param belPinName The physical pin name of the cell
+     * @return A list of site pin names on the cell's site to which the pin is routed.
+     * @since 2023.1.2
```
clavin-xlnx: Probably update the Javadoc here as well, but the map description can be brief.  Also, we only mandate use of the `@since` tag in the closed source portion, but since it was on the original, you might as well update it to `@since 2025.2.0`.

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx: Maybe update the Javadoc with a quick note telling the caller that this method is now multi-threaded based on the settings in `ParallelismTools`.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2460,23 +2539,39 @@ public static List<String> getAllRoutedSitePinsFromPhysicalPin(Cell cell, Net ne
...
-                Net parentNet = design.getNet(parentEhn.getHierarchicalNetName());
-                if (parentNet != null) {
-                    // 'net' is not a parent net (which normally causes createMissingSitePinInsts(Design, Net)
-                    // to analyze its parent net) but that parent net also exist in the design and has been/
-                    // will be analyzed in due course, so skip doing so here
-                    continue;
+        Map<SiteInst, Map<Net, List<String>>> siteInstToNetSiteWiresMap = DesignTools.getSiteInstToNetSiteWiresMap(design);
+        final CountUpDownLatch threadsOutstanding = new CountUpDownLatch();
+        int numNets = design.getNets().size();
+        int numJobs = ParallelismTools.maxParallelism() * 100;
```
clavin-xlnx, line 2595: We should describe in a comment why 100 (experimentally was the best performing...)

src/com/xilinx/rapidwright/rwroute/PartialRouter.java
```diff
@@ -853,7 +855,8 @@ public static void main(String[] args) {
 
         // Reads in a design checkpoint and routes it
         String[] rwrouteArgs = Arrays.copyOfRange(args, 2, args.length);
-        Design routed = routeDesignWithUserDefinedArguments(Design.readCheckpoint(args[0]), rwrouteArgs);
+        Design d = Design.readCheckpoint(args[0]);
+        Design routed = routeDesignWithUserDefinedArguments(d, rwrouteArgs);
```
clavin-xlnx:
```suggestion
        Design routed = routeDesignWithUserDefinedArguments(Design.readCheckpoint(args[0]), rwrouteArgs);
```
I think we can revert this, just a leftover change from debugging.

clavin-xlnx: In fact, do we need to make any changes in `PartialRouter.java` at all?  If not, let's just revert it.

src/com/xilinx/rapidwright/design/DesignTools.java
abutt-amd: ConcurrentMap is an interface. My understanding is you should use ConcurrentMap instead of Map when you want to use the ConcurrentHashMap in threaded contexts, but please correct me if I'm wrong.

src/com/xilinx/rapidwright/rwroute/PartialRouter.java
abutt-amd: Yeah you're right I'll revert Partial Router completely

src/com/xilinx/rapidwright/design/DesignTools.java
clavin-xlnx: Yes, `ConcurrentMap` is an interface, however, `Map` is also an interface and `Map` is more generic than `ConcurrentMap` so we prefer the more generic one.  For example, if you create a method signature that takes a parameter that is a `Map<T>`, we can call that method with any kind of map we want.  However, if we create a method to take a `ConcurrentMap<T>`, then we are forced to provide a `ConcurrentMap` implementation which might be slower depending on the context.  

If for some reason in the future, we wanted to overload this method where the user could provide a map, perhaps they would want a faster variant that was only for single-threaded context.  In this case, they would want to switch `ConcurrentMap` to `Map` to make that extra flexibility possible.

abutt-amd:
This is passing test cases locally but is now dependent on an update to the rapidwright api.

clavin-xlnx (approved):
I've looked over the code and don't have any major comments.  But I think I will defer to @eddieh-xlnx as I think he might be able to provide more substantial feedback on the parallelization techniques.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2265,13 +2295,20 @@ public static List<SitePinInst> createMissingSitePinInsts(Design design, Net net
 
         EDIFNetlist netlist = design.getNetlist();
         EDIFHierNet parentEhn = null;
+        Map<EDIFHierPortInst, Cell> cellCache = new HashMap<>();
         for (EDIFHierPortInst p :  physPins) {
-            Cell c = design.getCell(p.getFullHierarchicalInstName());
+            Cell c = cellCache.containsKey(p) ? cellCache.get(p) : design.getCell(p.getFullHierarchicalInstName());
+            if (!cellCache.containsKey(p)) {
+                cellCache.put(p, c);
+            }
```
eddieh-xlnx: Look at using `computeIfAbsent` here. Otherwise, you're doing 4 lookups.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2265,13 +2295,20 @@ public static List<SitePinInst> createMissingSitePinInsts(Design design, Net net
...
             if (c == null) continue;
             BEL bel = c.getBEL();
             if (bel == null) continue;
             String logicalPinName = p.getPortInst().getName();
-            Set<String> physPinMappings = c.getAllPhysicalPinMappings(logicalPinName);
+            Set<String> physPinMappings;
+            synchronized (design.getCell(c.getName())) {
```
eddieh-xlnx: 1. Why do you need to synchronize on the cell here? I don't think any pin mappings need to be modified in this method, nor do cells need to be created.
2. Why is it necessary to lookup the cell again here?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3305,67 +3455,85 @@ public static void printSiteInstInfo(SiteInst siteInst, PrintStream ps) {
...
+                futures.add((Future<Void>) f);
             }
-
-            if (parentPhysNet != null) {
-                design.movePinsToNewNetDeleteOldNet(net, parentPhysNet, true);
-            }
+        }
+        while (!futures.isEmpty()) {
+            ParallelismTools.joinFirst(futures);
         }
```
eddieh-xlnx: `joinFirst()` is for when you want to get the result from the first Future that completes so that it can be operated on. You want this instead:

```suggestion
        ParallelismTools.join(futures);
```

eddieh-xlnx: You'll probably need to change `futures` to a `List` too.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2332,16 +2385,66 @@ public static String getRoutedSitePin(Cell cell, Net net, String logicalPinName)
...
+    /**
+     * Creates a two-level map from SiteInst -> Net -> list of SiteWires. Calculating this map ahead of time ensures
+     * that {@link #createMissingSitePinInsts(Design, Net, Map)} does not need to iterate over the SiteInst ctags
+     * for each net. Instead, this method iterates over the ctags once per SiteInst and stores the resulting
+     * Net -> list of SiteWires map.
+     * @param design The design to create a SiteInst -> Net -> list of SiteWires map for.
+     * @return
+     */
+    @SuppressWarnings("unchecked")
+    public static Map<SiteInst, Map<Net, List<String>>> getSiteInstToNetSiteWiresMap(Design design) {
```
eddieh-xlnx: I'm not convinced this is necessary. My quick take is that this Map is only used by each Net (i.e. `createMissingSitePinInsts(Design, Net, ...)`, `getRoutedSitePinFromPhysicalPin(Cell, Net, ...)`, `getAllRoutedSitePinsFromPhysicalPin(Cell, Net, ...)`) which indicates to me that a global map (keyed on `SiteInst`) is not the best thing to do.

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: Actually, can you convince me that `Map<EDIFHierPortInst, Cell> cellCache` adds anything here? Since you're iterating over `physPins`, is it expected that you will ever lookup one of those pins more than once?

I could maybe see value in `Map<EDIFHierCellInst, Cell>`, but even then, how often does the same pin connect to the same cell? I'd need to see some data.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3305,67 +3455,85 @@ public static void printSiteInstInfo(SiteInst siteInst, PrintStream ps) {
...
+                                parentPhysNet = design.getVccNet();
+                            } else if (parentNetName.equals(EDIFTools.LOGICAL_GND_NET_NAME)) {
+                                parentPhysNet = design.getGndNet();
+                            } else {
+                                parentPhysNet = design.getNet(parentHierNet.getHierarchicalNetName());
+                            }
+
+                            if (parentPhysNet == null) {
+                                synchronized (design) {
+                                    if (!net.rename(parentHierNet.getHierarchicalNetName())) {
```
eddieh-xlnx, line 3511: Depending on how many things you have to rename (admittedly, there shouldn't be many), this might be quite slow as you're essentially enforcing an atomic section here on the `design` singleton. It's likely better to save everything you need to rename into a concurrent collection (e.g. `ConcurrentLinkedQueue`) so that the actual renaming can happen outside of the parallel section.

Same for the `movePinstToNewNetDeleteOldNet()` call below.

src/com/xilinx/rapidwright/design/DesignTools.java
abutt-amd: 1. We need to synchronize on the cell because of the `populateL2PMap` method called in `getAllPhysicalPinMappings` is not thread-safe. In the rare case that this map doesn't already exist and `getAllPhysicalPinMappings` is called on the same cell at the same time on two different threads there will be concurrent modification exceptions on the creation on the `pinMappingsL2P` map.

2. If we directly synchronize on c we will be synchronizing on a local variable which won't actually synchronize on the cell across threads (based on my understanding of the java synchronization semantics)

src/com/xilinx/rapidwright/design/DesignTools.java
abutt-amd: Good idea, I did not see this API. I will change to use `join`.

src/com/xilinx/rapidwright/design/DesignTools.java
abutt-amd: This map is actually very important. This optimization alone results in a 2x speedup of `createMissingSitePinInsts`. The reason it's keyed on SiteInst and why it results in such a big speedup is because of the way the ctags are stored on the SiteInst. If we try to lookup per net of the SiteInst we need to iterate over every ctag each time we look up a net on a given site inst, but when we precalculate this map we only need to iterate over the ctags once per SiteInst.

src/com/xilinx/rapidwright/design/DesignTools.java
abutt-amd: Yeah you're right, this is a remnant of an optimization I tried but didn't work out. I'll look a bit more and see whether an EDIFHierCellInst map makes sense.

src/com/xilinx/rapidwright/design/DesignTools.java
abutt-amd: Yeah my assumption here is that the number of nets we need to rename should be low, but i wouldn't be opposed to making a change here if you think it would be better.

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: 1. You're right -- it's `populateL2PMap()` that needs to be synchronized on.

2.  My understanding is that `c` == `design.getCell(c.getName())`. There is no local **copy** that you're synchronizing on -- both `c` and `design.getCell(c.getName())` refer to the same object.

abutt-amd: 2. Looking at this more you're right, I'll remove the additional lookup.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2282,7 +2319,12 @@ public static List<SitePinInst> createMissingSitePinInsts(Design design, Net net
                     if (siteWireNet == null) {
                         if (isVersal && net.isStaticNet() && bel.isLUT()) {
                             siteWireNet = net;
-                            si.routeIntraSiteNet(net, belPin, belPin);
+                            synchronized (si) {
+                                si.routeIntraSiteNet(net, belPin, belPin);
+                            }
+                            if (siteInstToNetSiteWiresMap != null) {
+                                siteInstToNetSiteWiresMap.get(si).put(net, si.getSiteWiresFromNet(net));
```
eddieh-xlnx: `si.getSiteWiresFromNet()` is linear complexity, and returns a whole new list rather than adding to the one that already exists. Is it possible to just add incrementally to the existing list?

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2226,7 +2248,11 @@ public static List<SitePinInst> createMissingSitePinInsts(Design design, Net net
             // Likely net inside encrypted IP, let's see if we can infer anything from existing
             // physical description
             for (SiteInst siteInst : new ArrayList<>(net.getSiteInsts())) {
-                for (String siteWire : new ArrayList<>(siteInst.getSiteWiresFromNet(net))) {
+                Map<Net, List<String>> netSiteWiresMap = siteInstToNetSiteWiresMap != null ?
+                        siteInstToNetSiteWiresMap.get(siteInst) : null;
+                List<String> netSiteWires = netSiteWiresMap != null ?
+                        netSiteWiresMap.get(net) : siteInst.getSiteWiresFromNet(net);
+                for (String siteWire : new ArrayList<>(netSiteWires)) {
```
eddieh-xlnx: I don't think this needs to a copy anymore (it did in the past).

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2256,6 +2282,10 @@ public static List<SitePinInst> createMissingSitePinInsts(Design design, Net net
 
                         currPin = net.createPin(pin.getName(), siteInst);
                         newPins.add(currPin);
+                        if (siteInstToNetSiteWiresMap != null) {
+                            siteInstToNetSiteWiresMap.get(siteInst)
```
eddieh-xlnx: `si.getSiteWiresFromNet()` is linear complexity, and returns a whole new list rather than adding to the one that already exists. Is it possible to just add incrementally to the existing list?

Also, is `computeIfAbsent()` necessary since there must have been a list to begin with?

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: 1. I would add a comment to make it more obvious to future maintainers.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2256,6 +2282,10 @@ public static List<SitePinInst> createMissingSitePinInsts(Design design, Net net
...
+                                    .computeIfAbsent(net, k -> new ArrayList<>()).add(currPin.getSiteWireName());
```
eddieh-xlnx: Is `computeIfAbsent()` necessary since to get here it must have been present?
```suggestion
                        if (netSiteWiresMap != null) {
                            netSiteWiresMap.add(currPin.getSiteWireName());
```

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2352,15 +2455,43 @@ public static String getRoutedSitePinFromPhysicalPin(Cell cell, Net net, String
...
+     * @param cell The cell with the pin of interest.
+     * @param net The physical net to which this pin belongs
+     * @param belPinName The physical pin name of the cell
+     * @param siteInstToNetSiteWiresMap An optional two-level map from siteInst -> net -> siteWires.
+     *                                  See {@link #getSiteInstToNetSiteWiresMap(Design)} for more information.
+     * @return A list of site pin names on the cell's site to which the pin is routed.
+     * @since 2025.2.0
+     */
+    public static List<String> getAllRoutedSitePinsFromPhysicalPin(Cell cell, Net net, String belPinName, Map<SiteInst,
+                                                                   Map<Net, List<String>>> siteInstToNetSiteWiresMap) {
```
eddieh-xlnx: Could this take `Map<Net, List<String>>` directly instead, since `SiteInst` is not ambiguous?

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: Sorry, I think I misread some of the logic behind it.

The other improvement I would make would be to make it a:
```
Map<SiteInst, Map<Net, List<Integer>>>
```
instead, since you don't need to be operating with sitewire names (which are converted back to integer indices anyway prior to `SiteInst.getSiteWirePins()`).

Can you share with what design you used to measure a 2x improvement?

abutt-amd: Yeah it's a 12x32 tile array created by array builder. I can share a copy with you if you want but it's too big to upload to github.

abutt-amd: Also changing to use the integer indices directly is a good idea but will require some changes to the rapidwright api.

src/com/xilinx/rapidwright/design/DesignTools.java
abutt-amd: I think it is necessary to return the whole new list because we are running routeIntraSiteNet so I don't know which site wires will be used by the net. Let me know if you know some way to avoid regenerating the site wires each time.

src/com/xilinx/rapidwright/design/DesignTools.java
abutt-amd: Removed

abutt-amd: Actually it seems like this copy is still necessary because when the map exists a concurrent modification exception can be triggered when adding to the netSiteWires.

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: Isn't `routeIntraSiteNet()` 3 lines above is only called to route one sitewire: `BELPin.getSiteWireName()` (or `.getSiteWireIndex()`)?

src/com/xilinx/rapidwright/design/DesignTools.java
eddieh-xlnx: ~Actually, `Net` is not ambiguous either.~

EDIT: Actually, you need to lookup vccNet

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2454,29 +2582,46 @@ public static List<String> getAllRoutedSitePinsFromPhysicalPin(Cell cell, Net ne
...
+        List<List<Net>> partitionedNets = Lists.partition(designNets, (int) Math.ceil((double) numNets / numJobs));
+        List<Future<Void>> futures = new ArrayList<>();
+        for (List<Net> nets : partitionedNets) {
+            Future<?> f = ParallelismTools.submit(() -> {
+                for (Net net : nets) {
+                    if (net.isUsedNet()) {
+                        continue;
+                    }
+                    EDIFHierNet ehn = net.getLogicalHierNet();
+                    EDIFHierNet parentEhn = (ehn != null) ? netlist.getParentNet(ehn) : null;
```
eddieh-xlnx, line 2606: When the parent net map is not already generated, `netlist.getParentNet()` will generate it and it is not synchronized.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2332,16 +2382,64 @@ public static String getRoutedSitePin(Cell cell, Net net, String logicalPinName)
...
+        int numJobs = ParallelismTools.maxParallelism() * 100;
+        List<Cell> designCells = new ArrayList<>(design.getCells());
+        List<List<Cell>> partitionedCells = Lists.partition(designCells, (int) Math.ceil((double) numCells / numJobs));
+        for (List<Cell> cells : partitionedCells) {
+            Future<?> f = ParallelismTools.submit(() -> {
+                for (Cell c : cells) {
+                    SiteInst inst = c.getSiteInst();
+                    siteInstToNetSiteWiresMap.put(inst, inst.getNetToSiteWireIndicesMap());
+                }
+            });
```
eddieh-xlnx, line 2434: Shouldn't you be iterating over all `SiteInst`s, rather than all `Cell`s since each `SiteInst` can (and typically does) contain more than one `Cell`?

eddieh-xlnx:
Here are my average-of-3 runtime results of `DesignTools.createMissingSitePinInsts()` (with a parent net map already generated) on a 32-thread (16 core) bare metal server, for a 12x12 array that was shared with me on 2025-10-15:
* Original: 156.2s
* This PR at 07f3af373: 25.1s
* This PR, but with the new `DesignTools.getSiteInstToNetSiteWiresMap()` method just returning `null`: 37.4s

I'm working on a simplification of this PR that brings this down to: ~20s that I'll push shortly, so don't start working on any of the comments above.

PR #1318, opened 2025-11-07
Add ability to constrain flop harness to specific side of PBlock

abutt-amd:
To ensure routability of ArrayBuilder designs, it is important that there is a routing path from a precompiled module in the direction of the input of the next precompiled module. A side map is provided to PerformanceExplorer which specifies which side of the pblock a given top-level I/O should have its flop placed on. Also unroutes top-level I/O nets that exit the pblock to prevent node overlaps between blocks placed near each other in the array.

### DIFF
--- a/src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
+++ b/src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
@@ -502,17 +502,21 @@ public class InlineFlopTools {
         }
     }
 
+    private static final Map<Series,String[]> staticPinsMap;
+    static {
+        staticPinsMap = new HashMap<>();
+        staticPinsMap.put(Series.Series7, new String[]{"CE", "SR"});
+        staticPinsMap.put(Series.UltraScale, new String[]{"CKEN1", "CKEN2", "CKEN3", "CKEN4", "SRST1", "SRST2"});
+        staticPinsMap.put(Series.UltraScalePlus, staticPinsMap.get(Series.UltraScale));
+        staticPinsMap.put(Series.Versal, new String[]{"CKEN1", "CKEN2", "CKEN3", "CKEN4", "RST"});
+    }
+
     private static String[] getStaticPins(Design design) {
-        String[] versalStaticPins = new String[]{"CKEN1", "CKEN2", "CKEN3", "CKEN4", "RST"};
-        String[] ultrascaleStaticPins = new String[]{"CKEN1", "CKEN2", "CKEN3", "CKEN4", "SRST1", "SRST2"};
-        String[] series7StaticPins = new String[]{"CE", "SR"};
-        if (design.getSeries() != Series.Versal && design.getSeries() != Series.UltraScale
-                && design.getSeries() != Series.UltraScalePlus && design.getSeries() != Series.Series7) {
+        String[] staticPins = staticPinsMap.get(design.getSeries());
+        if (staticPins == null) {
             throw new RuntimeException("Unsupported device series for removing inline flops");
         }
-        return design.getSeries() == Series.Versal ? versalStaticPins :
-                              design.getSeries() == Series.UltraScalePlus
-                              || design.getSeries() == Series.UltraScale ? ultrascaleStaticPins : series7StaticPins;
+        return staticPins;
     }
 
     /**


src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
```diff
@@ -94,6 +96,103 @@ public static void createAndPlaceFlopsInlineOnTopPortsNearPins(Design design, St
...
+                                                      Map<EDIFPort, PBlockSide> portSideMap) {
+        assert (design.getSiteInsts().isEmpty());
+        Site start = keepOut.getAllSites("SLICE").iterator().next(); // TODO this is a bit wasteful
+        boolean exclude = true;
+
+        EDIFHierNet clk = design.getNetlist().getHierNetFromName(clkNet);
+
+        Set<SiteInst> siteInstsToRoute = new HashSet<>();
+
+        for (Map.Entry<EDIFPort, PBlockSide> entry : portSideMap.entrySet()) {
```
clavin-xlnx: I often just import `Entry` (`for (Entry<EDIFPort, PBlockSide> entry : ...`)

src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
```diff
@@ -94,6 +96,103 @@ public static void createAndPlaceFlopsInlineOnTopPortsNearPins(Design design, St
...
+            EDIFPort port = entry.getKey();
+            PBlockSide side = entry.getValue();
+            if (port.getName().equals(clkNet)) {
+                continue;
+            }
+            Site shiftedSite = shiftSiteToSide(design.getDevice(), start, keepOut, side);
+            if (port.isBus()) {
+                for (int i : port.getBitBlastedIndicies()) {
```
clavin-xlnx: This is a lot of duplicated code.  I realize this is partly due to the nature of bussed vs. single bit ports.  One way to work around this is to use this pattern:
```java
            for (int i : (port.isBus() ? port.getBitBlastedIndicies() : new int[] { 0 })) {
                EDIFPortInst portInst = port.getInternalPortInstNameFromIndex(i));
```
A better solution might be to modify `EDIFPort.getBitBlastedIndicies()` so that for the single bit case it returns a static `new int[] { 0 }`.

src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
```diff
@@ -135,7 +232,9 @@ private static void createAndPlaceFlopsInlineOnTopPorts(Design design, String cl
                     if (centroidPlacement) {
                         netCentroidFlipFlopPlacement(design, inst, keepOut, clk, siteInstsToRoute);
                     } else {
-                        Pair<Site, BEL> loc = nextAvailPlacement(design, siteItr);
+                        Iterator<Site> siteItr = ECOPlacementHelper.spiralOutFrom(start, keepOut, exclude).iterator();
```
clavin-xlnx: I see this code is also duplicated, sorry for the bad example.  Perhaps we can refactor this one as well?

src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
```diff
@@ -272,14 +437,17 @@ public static void removeInlineFlops(Design design) {
         Net vcc = design.getVccNet();
         Set<SitePinInst> vccPins = new HashSet<>();
         pinsToRemove.put(vcc, vccPins);
-        String[] staticPins = new String[]{"CKEN1", design.getSeries() == Series.Versal ? "RST" : "SRST1"};
+        String[] staticPins = new String[]{"CKEN1", "CKEN2", design.getSeries() == Series.Versal ? "RST" : "SRST1"};
```
clavin-xlnx: Is this change also true for Series7/UltraScale/UltraScale+?

src/com/xilinx/rapidwright/design/blocks/PBlockSide.java
```diff
@@ -0,0 +1,31 @@
+/*
```
clavin-xlnx, line 1: There is already an `Edge` enum in this package, is it not feasible to try and use it also?  Another option is `TimingDirection` in the `com.xilinx.rapidwright.timing` package.

If you still want to keep it, I think we should refactor it to the same package as `PBlock`, which is `com.xilinx.rapidwright.design.blocks`.

src/com/xilinx/rapidwright/util/PerformanceExplorer.java
```diff
@@ -288,6 +301,38 @@ public void setEnsureExternalRoutability(boolean ensureExternalRoutability) {
...
+            externalRoutabilitySideMap = new HashMap<>();
+            List<String> lines = FileTools.getLinesFromTextFile(getExternalRoutabilitySideFile());
+
+            for (String line : lines) {
+                String[] splitLine = line.split("\\s+");
+                String portRegex = splitLine[0];
+                String pblockSide = splitLine[1].toUpperCase();
+                for (EDIFPort port : netlist.getTopCell().getPorts()) {
+                    if (port.getBusName().matches(portRegex) ||
+                            port.getName().matches("\\" + EDIFTools.VIVADO_PRESERVE_PORT_INTERFACE + portRegex)) {
```
clavin-xlnx: Just an FYI - `String.matches()` is notoriously slow.  I think in this case the regex is unique per line, so refactoring it to compile the regex ahead of time is not going to help.  The performance of this method is probably not going to be an issue, so I don't think a change is needed here.

src/com/xilinx/rapidwright/util/PerformanceExplorer.java
```diff
@@ -288,6 +301,38 @@ public void setEnsureExternalRoutability(boolean ensureExternalRoutability) {
...
 
+    public String getExternalRoutabilitySideFile() {
+        return externalRoutabilitySideFile;
+    }
+
+    public void setExternalRoutabilitySideFile(String externalRoutabilitySideFile) {
+        this.externalRoutabilitySideFile = externalRoutabilitySideFile;
+    }
+
+    public Map<EDIFPort, PBlockSide> getExternalRoutabilitySideMap(EDIFNetlist netlist) {
```
clavin-xlnx: Since this map is most associated with the `InlineFlopTools` set of methods, should we refactor it by making it static and moving it there?  I think you would only need to make the filename a parameter.

src/com/xilinx/rapidwright/util/PerformanceExplorer.java
```diff
@@ -506,16 +560,48 @@ public void getBestDesignPerPBlock() {
...
             d.writeCheckpoint(runDirectory + File.separator + pblock + "_best.dcp");
         }
     }
 
+    public void unrouteNetsThatLeavePBlock(Design d, PBlock pBlock) {
+        DesignTools.makePhysNetNamesConsistent(d);
+        d.getNetlist().resetParentNetMap();
+        d.getNetlist().expandMacroUnisims();
+        for (EDIFPort p : d.getNetlist().getTopCell().getPorts()) {
+            int[] indices = p.isBus() ? p.getBitBlastedIndicies() : new int[]{0};
```
clavin-xlnx: Looks like you used the pattern here!

src/com/xilinx/rapidwright/util/PerformanceExplorer.java
```diff
@@ -506,16 +560,48 @@ public void getBestDesignPerPBlock() {
...
                 InlineFlopTools.removeInlineFlops(d);
+                unrouteNetsThatLeavePBlock(d, getPBlock(pblockNum));
             }
 
+            EDIFTools.ensurePreservedInterfaceVivado(d.getNetlist());
             d.writeCheckpoint(runDirectory + File.separator + pblock + "_best.dcp");
         }
     }
 
+    public void unrouteNetsThatLeavePBlock(Design d, PBlock pBlock) {
```
clavin-xlnx: This looks like a handy method and seems like it should be labeled `static`.  Should we refactor it to more generic location, `NetTools` perhaps?

src/com/xilinx/rapidwright/util/PerformanceExplorer.java
abutt-amd: Yeah I know it's slow but I figured it wasn't that big of an issue here. The number of regex and port pairs to check should be relatively low.

src/com/xilinx/rapidwright/design/blocks/PBlockSide.java
abutt-amd: Edge has a bunch of options that I don't need and TimingDirection seems like it would be a confusing name. I can refactor into `com.xilinx.rapidwright.design.blocks` unless you really want me to use Edge/TimingDirection.

src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
abutt-amd: This is actually a bit more complicated than I thought and there are different values for versal, ultrascale, and series7. I refactored to handle each of these cases separately.

src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
abutt-amd: We decided that getBitBlastedIndicies should be updated to return `{ 0 }` when called on a non-bused port. I will refactor further once this change has been made.

abutt-amd: Updated to use new `getBitBlastedIndices` semantics.

src/com/xilinx/rapidwright/design/NetTools.java
```diff
@@ -517,4 +521,40 @@ public static Node findClockRootVRoute(Net net) {
...
+     * where the top-level port nets leave the PBlock if the InlineFlopTools flop harness was used and the top-level
+     * port net also has a sink inside the PBlock.
+     *
+     * @param d The design to unroute nets from.
+     * @param pBlock The bounding box for the placed and routed design.
+     */
+    public static void unrouteTopLevelNetsThatLeavePBlock(Design d, PBlock pBlock) {
+        DesignTools.makePhysNetNamesConsistent(d);
+        d.getNetlist().resetParentNetMap();
+        d.getNetlist().expandMacroUnisims();
```
clavin-xlnx: Are these necessary and should they be part of this method?  I could imagine scenarios where these methods might be called unnecessarily and lead to a slow down.  We could update the Javadoc to be clear about the requirements of the state of the design before calling this method.  

It would also be good to have a test.

src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
```diff
@@ -437,7 +409,16 @@ public static void removeInlineFlops(Design design) {
         Net vcc = design.getVccNet();
         Set<SitePinInst> vccPins = new HashSet<>();
         pinsToRemove.put(vcc, vccPins);
-        String[] staticPins = new String[]{"CKEN1", "CKEN2", design.getSeries() == Series.Versal ? "RST" : "SRST1"};
+        String[] versalStaticPins = new String[]{"CKEN1", "CKEN2", "CKEN3", "CKEN4", "RST"};
```
clavin-xlnx: Could we add a small test to exercise these improvements to make sure we don't break them unintendedly later?

abutt-amd: I added a parameterized test that exercises this code by testing every possible BEL for an inline flop to be placed at.

src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
```diff
@@ -24,6 +24,7 @@
 
 import java.util.*;
```
clavin-xlnx: Doesn't look like this is a change you made in  this PR, but we avoid using the "*" import pattern.  Would you mind fixing this in this PR so that all the imports are explicit?

src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
```diff
@@ -200,7 +200,7 @@ public static void createAndPlacePortFlopsOnSide(Design design, String clkNet, P
      */
     private static void createAndPlaceFlopsInlineOnTopPorts(Design design, String clkNet, PBlock keepOut,
                                                             boolean centroidPlacement) {
-        assert (design.getSiteInsts().isEmpty());
+//        assert (design.getSiteInsts().isEmpty());
```
clavin-xlnx: Why was this commented out?  Seems like we might want to keep it for testing?

src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
```diff
@@ -363,8 +363,8 @@ private static Pair<Site, BEL> nextAvailPlacement(Design design, Iterator<Site>
         return null;
     }
 
-    private static Cell createAndPlaceFlopInlineOnTopPortInst(Design design, EDIFPortInst portInst, Pair<Site, BEL> loc,
-                                                              EDIFHierNet clk) {
+    public static Cell createAndPlaceFlopInlineOnTopPortInst(Design design, EDIFPortInst portInst, Pair<Site, BEL> loc,
```
clavin-xlnx, line 383: If this is becoming `public`, perhaps we could add a Javadoc?  Probably can reuse most of what is in the Javadoc in a method above.

test/src/com/xilinx/rapidwright/design/tools/TestInlineFlopTools.java
```diff
@@ -0,0 +1,176 @@
...
+        Design design = new Design("inline_flops", partName);
+        design.setDesignOutOfContext(true);
+        design.setAutoIOBuffers(false);
+        EDIFCell topCell = design.getNetlist().getTopCell();
+        // Create clock port and net
+        EDIFPort clkPort = topCell.createPort("clk", EDIFDirection.INPUT, 1);
+        EDIFNet clkNet = topCell.createNet("clk");
+        clkNet.createPortInst(clkPort);
+        EDIFHierNet logicalClkNet = new EDIFHierNet(design.getNetlist().getTopHierCellInst(), clkPort.getInternalNet());
+        Net clk = design.createNet(logicalClkNet);
```
clavin-xlnx: You can actually simplify this code by using the fact that `Design.createNet(String)` will create a logical net for you:
```suggestion
        Net clk = design.createNet("clk");
        EDIFHierNet logicalClkNet = clk.getLogicalHierNet();
        logicalClkNet.getNet().createPortInst(clkPort);
```

test/src/com/xilinx/rapidwright/design/tools/TestInlineFlopTools.java
```diff
@@ -0,0 +1,176 @@
...
+        logicalInNet.createPortInst(inPort);
+        EDIFHierNet hierInNet = new EDIFHierNet(design.getNetlist().getTopHierCellInst(), inPort.getInternalNet());
+        Net inNet = design.createNet(hierInNet);
+
+        // Create out port and net
+        EDIFPort outPort = topCell.createPort("out", EDIFDirection.OUTPUT, 1);
+        EDIFNet logicalOutNet = topCell.createNet("out");
+        logicalOutNet.createPortInst(outPort);
+        EDIFHierNet hierOutNet = new EDIFHierNet(design.getNetlist().getTopHierCellInst(), outPort.getInternalNet());
+        Net outNet = design.createNet(hierOutNet);
```
clavin-xlnx: You can use the same simplfication from above in these two situations as well.

test/src/com/xilinx/rapidwright/design/TestNetTools.java
```diff
@@ -42,15 +50,15 @@ public class TestNetTools {
...
      * The method NetTools.isGlobalClock(Net net) returns true for these nets and false for others.
-     * 
+     *
      * All Versal/Ultrascale+/Series-7 designs under the RapidWrightDCP path are available to be checked, except for the following:
      *      picoblaze4_ooc_X6Y60_X6Y65_X10Y60_X10Y65.dcp    The source cell bufgce_inst of this design is not placed yet.
      *      picoblaze_ooc_X10Y235_unreadable_edif.dcp       This design would call vivado to generate an readable EDIF file.
-     * 
-     * @param pathAndlobalClockNames Use ' ' as the delimiter to split this string, where the first element is the path to the DCP file, 
+     *
+     * @param pathAndlobalClockNames Use ' ' as the delimiter to split this string, where the first element is the path to the DCP file,
```
clavin-xlnx: Not your initial change, but would you mind refactoring `pathAndlobalClockNames` to `pathAndGlobalClockNames`?

src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
abutt-amd: I think this should actually be removed because we now support placing inline flops for designs that have already been placed, instead of just synthesized designs. The only reason I didn't run into issues here is because I wasn't running with assertions enabled.

abutt-amd: I'm going to add some tests the exercise this.

abutt-amd: Added two new tests in TestInlineFlopTools

src/com/xilinx/rapidwright/design/tools/InlineFlopTools.java
```diff
@@ -494,6 +502,19 @@ public static void removeInlineFlops(Design design) {
...
+        String[] ultrascaleStaticPins = new String[]{"CKEN1", "CKEN2", "CKEN3", "CKEN4", "SRST1", "SRST2"};
+        String[] series7StaticPins = new String[]{"CE", "SR"};
+        if (design.getSeries() != Series.Versal && design.getSeries() != Series.UltraScale
+                && design.getSeries() != Series.UltraScalePlus && design.getSeries() != Series.Series7) {
+            throw new RuntimeException("Unsupported device series for removing inline flops");
+        }
+        return design.getSeries() == Series.Versal ? versalStaticPins :
+                              design.getSeries() == Series.UltraScalePlus
+                              || design.getSeries() == Series.UltraScale ? ultrascaleStaticPins : series7StaticPins;
+    }
```
clavin-xlnx, line 520: Seems like this could be handled with a static map, for example:
```java
        private static Map<Series,String[]> staticPinsMap;
        static { 
             staticPinsMap = new HashMap<>();
             staticPinsMap.put(Series.Series7, new String[]{"CE", "SR"});
             staticPinsMap.put(Series.UltraScale, new String[]{"CKEN1", "CKEN2", "CKEN3", "CKEN4", "SRST1", "SRST2"});
             staticPinsMap.put(Series.UltraScalePlus, staticPinsMap.get(Series.UltraScale)); 
             staticPinsMap.put(Series.Versal, new String[]{"CKEN1", "CKEN2", "CKEN3", "CKEN4", "RST"});
        }

        private static String[] getStaticPins(Design design) {
             String[] staticPins = staticPinsMap.get(design.getSeries());
             if (staticPins == null) {
                 throw new RuntimeException("Unsupported device series for removing inline flops");
             }
             return staticPins;
        }
```

clavin-xlnx (approved):

PR #1317, opened 2025-11-07
Only remove top level Vivado bus prevention annotations

abutt-amd:
Previously the code on array_builder to remove Vivado bus prevention annotations would recursively remove bus prevention annotations. This method was prone to creating inconsistencies in the netlist and is not actually necessary for the goal of ensuring consistent interfaces after roundtripping to Vivado.

PR #1319, opened 2025-11-07
Automate placement in ArrayBuilder

abutt-amd:
Automates placement of arrays in ArrayBuilder. Provides two different ways to pick a placement, a greedy method and a CP-SAT method using OR-Tools. Should not be merged until after #1318, #1316 and #1317.

### DIFF
--- a/src/com/xilinx/rapidwright/design/tools/ArrayBuilder.java
+++ b/src/com/xilinx/rapidwright/design/tools/ArrayBuilder.java
@@ -73,7 +73,6 @@ import com.xilinx.rapidwright.edif.EDIFCellInst;
 import com.xilinx.rapidwright.edif.EDIFDirection;
 import com.xilinx.rapidwright.edif.EDIFHierCellInst;
 import com.xilinx.rapidwright.edif.EDIFHierNet;
-import com.xilinx.rapidwright.edif.EDIFHierPortInst;
 import com.xilinx.rapidwright.edif.EDIFNet;
 import com.xilinx.rapidwright.edif.EDIFNetlist;
 import com.xilinx.rapidwright.edif.EDIFPort;
@@ -750,6 +749,7 @@ public class ArrayBuilder {
 
         Design array = null;
         List<String> modInstNames = null;
+        // List containing pairs of (x,y) coordinates with the moduleInst name placed at that ideal (x,y) coordinate
         List<Pair<Pair<Integer, Integer>, String>> idealPlacementList = null;
         if (ab.getTopDesign() == null) {
             array = new Design("array", ab.getKernelDesign().getPartName());
@@ -968,7 +968,7 @@ public class ArrayBuilder {
             // Automatically find bounding PBlock based on used Slices, DSPs, and BRAMs
             Set<Site> usedSites = new HashSet<>();
             for (SiteInst siteInst : array.getSiteInsts()) {
-                if (siteInst.getName().contains("STATIC_SOURCE_SLICE")) {
+                if (siteInst.getName().contains(SiteInst.STATIC_SOURCE)) {
                     continue;
                 }
                 if (isSLICE(siteInst) || isBRAM(siteInst) || isDSP(siteInst)) {
@@ -1007,6 +1007,7 @@ public class ArrayBuilder {
         if (newRowMap.isEmpty()) {
             return placement;
         }
+        // Map from (x,y) coordinate to the moduleInst name placed at that ideal (x,y) coordinate
         Map<Pair<Integer, Integer>, String> newPlacement = new HashMap<>(placement);
 
         // Check if row updates are unique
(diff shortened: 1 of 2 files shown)


src/com/xilinx/rapidwright/design/tools/ArrayBuilder.java
```diff
@@ -612,19 +747,43 @@ public static void main(String[] args) {
             }
             modules.add(m);
         }
-        t.stop().start("Place Instances");
 
         Design array = null;
         List<String> modInstNames = null;
+        List<Pair<Pair<Integer, Integer>, String>> idealPlacementList = null;
```
clavin-xlnx, line 753: I think longer term, we may want to create an object here, but for now, just add a comment to describe what all the fields stand for.  A `List<Pair<Pair<Integer, Integer>, String>>` is a bit abstract.

src/com/xilinx/rapidwright/design/tools/ArrayBuilder.java
```diff
@@ -766,16 +951,26 @@ public static void main(String[] args) {
...
+        }
         array.getNetlist().consolidateAllToWorkLibrary();
+        array.flattenDesign();
 
-        if (ab.getOutOfContext()) {
+        if (ab.isOutOfContext()) {
             // Automatically find bounding PBlock based on used Slices, DSPs, and BRAMs
             Set<Site> usedSites = new HashSet<>();
             for (SiteInst siteInst : array.getSiteInsts()) {
+                if (siteInst.getName().contains("STATIC_SOURCE_SLICE")) {
```
clavin-xlnx: Where is `STATIC_SOURCE_SLICE` defined?  There is a `SiteInst.STATIC_SOURCE` string that is `STATIC_SOURCE`.

src/com/xilinx/rapidwright/design/tools/ArrayBuilder.java
```diff
@@ -784,11 +979,71 @@ public static void main(String[] args) {
...
         array.writeCheckpoint(ab.getOutputName());
         t.stop().printSummary();
     }
 
+    private static Map<Pair<Integer, Integer>, String> foldIdealPlacement(Map<Pair<Integer, Integer>, String> placement,
+                                                                          Map<Integer, Integer> newRowMap) {
+        if (newRowMap.isEmpty()) {
+            return placement;
+        }
+        Map<Pair<Integer, Integer>, String> newPlacement = new HashMap<>(placement);
```
clavin-xlnx, line 1011: Again, just add a comment describing `Map<Pair<Integer, Integer>, String>`.

src/com/xilinx/rapidwright/design/tools/ArrayNetlistGraph.java
```diff
@@ -0,0 +1,430 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class ArrayNetlistGraph {
```
clavin-xlnx, line 63: Let's add a brief Javadoc description for what these classes do.

src/com/xilinx/rapidwright/design/tools/ArrayNetlistGraph.java
```diff
@@ -0,0 +1,430 @@
...
+        graph = new DefaultDirectedGraph<>(NetlistEdge.class);
+    }
+
+    public ArrayNetlistGraph(Design array, List<String> modules) {
+        this(array, modules, null);
+    }
+
+    public ArrayNetlistGraph(Design array, List<String> modules, Map<EDIFPort, PBlockSide> sideMap) {
+        this();
+        EDIFHierCellInst top = array.getNetlist().getTopHierCellInst();
```
clavin-xlnx: It may make sense to store a local reference to `EDIFNetlist` , for example `EDIFNetlist netlist = array.getNetlist();`. As it is being used below in the loop over modules.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3426,6 +3426,64 @@ public static void makePhysNetNamesConsistent(Design design) {
         ParallelismTools.join(futures);
     }
 
+    /**
+     * Make a single physical Net's name consistent with its logical (EDIF) netlist.
+     * @param design Design containing the net
+     * @param net The physical Net to make consistent
+     * @return The parent physical Net or null if already consistent
+     */
+    public static Net makePhysNetNameConsistent(Design design, Net net) {
```
clavin-xlnx, line 3435: I'm assuming we added this method to be more agile in performance?

abutt-amd: Yes, I only really use it when routing just the clock for now but I plan to use this to help avoid the preprocessing overhead more in the future.

PR #1321, opened 2025-11-12
[EDIFPort] Refactor getBitBlastedIndices()

clavin-xlnx:
This PR makes two changes:

1) EDIFTools.getBitBlastedIndicies() name was refactored to EDIFTools.getBitBlastedIndices() and the previous spelling was marked for deprecation due to a mispelling.

2) The return value of this method for a single bit port was changed such that instead of returning null, it now returns an array with a single value in it  `{0}`.  This behavior lends itself to simplifying code that traverses over ports such that it removes the need to check if it is a bussed port or single bit port.

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFPort.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFPort.java
@@ -378,7 +378,7 @@ public class EDIFPort extends EDIFPropertyObject {
         if (isBus()) {
             int lastLeftBracket = getName().lastIndexOf('[');
             assert(lastLeftBracket != -1);
-            return getName().contains(":") ? 
+            return getName().indexOf(':', lastLeftBracket) != -1 ? 
                 EDIFTools.bitBlastBus(getName().substring(lastLeftBracket)) : 
                 new int[] { Integer.parseInt(getName().substring(lastLeftBracket, getName().length() - 1)) };
         }


src/com/xilinx/rapidwright/edif/EDIFPort.java
```diff
@@ -356,13 +356,33 @@ public boolean isBus() {
...
+     * @return The integer list of indices of this port, or {0} for a single bit
+     *         port.
+     */
+    public int[] getBitBlastedIndices() {
+        if (isBus()) {
+            int lastLeftBracket = getName().lastIndexOf('[');
+            if (getName().contains(":"))
+                return EDIFTools.bitBlastBus(getName().substring(lastLeftBracket));
+            if (getName().contains("["))
+                return new int[] { Integer.parseInt(getName().substring(lastLeftBracket, getName().length() - 1)) };
```
eddieh-xlnx:
```suggestion
            assert(lastLeftBracket != -1);
            return getName().contains(":") ? 
                EDIFTools.bitBlastBus(getName().substring(lastLeftBracket)) : 
                new int[] { Integer.parseInt(getName().substring(lastLeftBracket, getName().length() - 1)) };
```

src/com/xilinx/rapidwright/edif/EDIFTools.java
```diff
@@ -1736,7 +1736,7 @@ public static EDIFNetlist createFlatNetlist(EDIFNetlist netlist, String partName
         for (EDIFPort topPort : netlist.getTopCell().getPorts()) {
             EDIFPort flatPort = flatTop.createPort(topPort);
             if (flatPort.isBus()) {
-                int[] indicies = flatPort.getBitBlastedIndicies();
+                int[] indicies = flatPort.getBitBlastedIndices();
```
eddieh-xlnx: (Variable) misspelled here and below too.

clavin-xlnx: Sorry, I don't see the misspellings you are referring to.

src/com/xilinx/rapidwright/edif/EDIFPort.java
clavin-xlnx: I've never seen a bussed port that did not have a `[#]` or `[#:#]` at the end.  I've re-written to remove the `if`s.

src/com/xilinx/rapidwright/edif/EDIFTools.java
eddieh-xlnx:
```suggestion
                int[] indices = flatPort.getBitBlastedIndices();
```

src/com/xilinx/rapidwright/edif/EDIFTools.java
eddieh-xlnx:
```suggestion
                    flatNet.createPortInst(flatPort, indices[i++]);
```

src/com/xilinx/rapidwright/edif/EDIFPort.java
```diff
@@ -356,13 +356,33 @@ public boolean isBus() {
...
+     * when needing to iterate over a port's PortInst objects.
+     * 
+     * @return The integer list of indices of this port, or {0} for a single bit
+     *         port.
+     */
+    public int[] getBitBlastedIndices() {
+        if (isBus()) {
+            int lastLeftBracket = getName().lastIndexOf('[');
+            assert(lastLeftBracket != -1);
+            return getName().contains(":") ?
```
eddieh-xlnx: `contains()` searches the whole name from the beginning. I think we can do better by starting from the left bracket.


```suggestion
            return getName().indexOf(':', lastLeftBracket) != -1 ? 
```

src/com/xilinx/rapidwright/edif/EDIFTools.java
eddieh-xlnx: This suggestion seems to be broken

src/com/xilinx/rapidwright/edif/EDIFTools.java
clavin-xlnx: Thanks, I saw it and I think I have fixed all instances of `indicies` -> `indices`.

src/com/xilinx/rapidwright/edif/EDIFPort.java
clavin-xlnx: True, thanks.

PR #1322, opened 2025-11-12
[NetlistBrowser] Cell Instance Schematic Viewer

clavin-xlnx:
This PR augments the `NetlistBrowser` tool with a logical netlist schematic viewer that has a similar flavor to what is found in Vivado.  It is built on a combination of RapidWright's logical netlist, Eclipse Layout Kernel (ELK) and Qt Jambi.

Currently, it only displays one hierarchical cell instance at the top level, but you can descend into lower levels of hierarchy by expanding hierarchical cells.  Here is an example screenshot:

image

Users can navigate the netlist on the left netlist tree browser.  If a user selects a hierarchical cell, it will render the logical netlist of that cell in the right hand pane.  Blue blocks are hierarchical cells (expandable with the `+` button) and yellow blocks are leaf cells.   There is some limited selection capability by selecting cells, ports and nets on the schematic that will highlight in the right hand tree browser the netlist item.  

Example to run:
```
git fetch
git checkout schematic
./gradlew compileJava
git submodule update --init #If this hasn't been done already before
rapidwright NetlistBrowser test/RapidWrightDCP/microblazeAndILA_3pblocks.dcp
```


clavin-xlnx:
> Can netlistbrowser also support recursive search for a given cell name? It could be hierarchical or either the leaf cell.

@coherent17 I have prototyped a selection interface using Python as the interactive prompt (you could also do something similar with [`jshell`](https://docs.oracle.com/en/java/javase/25/jshell/introduction-jshell.html)).  You can setup this branch to run in Python by doing the following (assuming you have Python 3 on your PATH):

```bash
pip install rapidwright
git clone https://github.com/Xilinx/RapidWright -b schematic --recursive
cd RapidWright/
export RAPIDWRIGHT_PATH=`pwd`
export CLASSPATH=$RAPIDWRIGHT_PATH/bin:$RAPIDWRIGHT_PATH/jars/*
# Set JAVA_HOME if your System Java is different from the one on your PATH
./gradlew compileJava
python3
```

Then in the interactive Python interpreter:
```python
import rapidwright
from com.xilinx.rapidwright.design import Design
from com.xilinx.rapidwright.gui import NetlistBrowser
d = Design.readCheckpoint("test/RapidWrightDCP/microblazeAndILA_3pblocks.dcp")
n = d.getNetlist()
NetlistBrowser.select(n.getHierCellInstFromName('u_ila_0/inst/ila_core_inst'))
```
A window should open up with that current instance selected.  
image


If you return to the prompt and run:
```python
NetlistBrowser.select(n.findCellInsts('*capture_qual_ctrl_2_reg[0]').get(0))
```
This will search through the netlist for cell instances that end with the name "capture_qual_ctrl_2_reg[0]" and select the first matching result and display it in the window:

image

Separate from this, I have also added the ability to copy items from the netlist tree by right-clicking and either choosing the display name of the item or the hierarchical name (see screenshot below):

image

.github/workflows/test-rapidwright-wrapper.yml
```diff
@@ -13,11 +13,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Setup JDK 1.11
+      - name: Setup JDK 1.17
```
eddieh-xlnx, line 16: Not resistant to this change -- but is there a reason we should document behind moving to 17 (and why not the latest LTS?)

LICENSE.TXT
```diff
@@ -3237,7 +3239,7 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ********************************* END LICENSE *********************************
 
-******************* google-guava 28 - Apache 2.0 ********************************
+******************* google-guava 28, Google Guava 33.5.0 - Apache 2.0 ********************************
```
eddieh-xlnx, line 3242: Are we pulling in two version of Google Guava?

eddieh-xlnx (approved):
Works for me. Quite snappy, and really good quality for an initial attempt!

.github/workflows/test-rapidwright-wrapper.yml
clavin-xlnx: I switched this when I was trying to diagnose CI failures as Elk is being distributed with Java 17 jar builds.  The other CI test also uses Java 17 so I thought we should be consistent.  I resolved the issue by posting equivalent Java 8 jars of the Elk dependencies and use those if <Java 17 is detected.

LICENSE.TXT
clavin-xlnx: Jython packages a stripped down Google Guava 28 inside of it (in a different name space), with this PR we are adding the full Guava 33.5.0 library dependency.

coherent17:
Hi Chris,

After @sc0210's fixes in #1414, #1416, and #1418, we're now able to read the physically synthesized netlist before entering Vivado using RapidWright's EDIF parser which is quite useful for us.

However, we're seeing relatively high runtime in the netlist browser when the instance hierarchy is relatively deep. In some cases, clicking the expand '+' button can take several minutes to process.

I'm wondering if there's still room for improvement to make this run faster.

Also, if there's quick recursive name seaching throughout the netlist and expand to that hierachy would be quite helpful.

Thanks!

clavin-xlnx:
> Hi Chris,
> 
> After @sc0210's fixes in #1414, #1416, and #1418, we're now able to read the physically synthesized netlist before entering Vivado using RapidWright's EDIF parser which is quite useful for us.
> 
> However, we're seeing relatively high runtime in the netlist browser when the instance hierarchy is relatively deep. In some cases, clicking the expand '+' button can take several minutes to process.
> 
> I'm wondering if there's still room for improvement to make this run faster.
> 
> Also, if there's quick recursive name seaching throughout the netlist and expand to that hierachy would be quite helpful.
> 
> Thanks!

Hi @coherent17 - Please have a look at https://github.com/Xilinx/RapidWright/pull/1420

coherent17:
Hi Chris,

We're able to launch the netlistBrowser right now on this branch with some further fix(Sam will provide another PR to resolve the issue, also seems to have error when generating standalone jar file). The speed has been improved really large. I don't have a A/B test right now, but it actually be faster enough on the Versal design.

However, it seems to have some wire naming are too long, so that the layout are overlapped to each other.

Can we have these fix related to edif parser and netlist browser in next RapidWright release(2026.2)?

Thanks.

PR #1324, opened 2025-11-20
[DesignTools] Parallelize preprocessing methods

eddieh-xlnx:
Built on top/simplification of https://github.com/Xilinx/RapidWright/pull/1316 which parallelizes two preprocessing methods often required prior to routing:
* `makePhysNetNamesConsistent()`
* `createMissingSitePinInsts()`

Differences to #1316:
* Refactor of `DesignTools.getAllRoutedSitePinsFromPhysicalPin()` to lookup net on sitewires directly from the `SiteInst` (an O(1) operation), rather than rely on a set of used sitewires built by iterating over all sitewires.
* Removal of #1316's `getSiteInstToNetSiteWiresMap()` since the only place it will now give a saving is for physical nets living entirely inside an encrypted cell -- arguably a rare operation.
* `createMissingSitePinInsts()` now requires the `Design` to be consistent (made so by `makePhysNetNamesConsistent()`, which is typically called before the former)

### DIFF
--- a/src/com/xilinx/rapidwright/design/DesignTools.java
+++ b/src/com/xilinx/rapidwright/design/DesignTools.java
@@ -2483,7 +2483,6 @@ public class DesignTools {
      * (use {@link #makePhysNetNamesConsistent(Design)} to make this the case).
      * @param design The current design
      */
-    @SuppressWarnings("unchecked")
     public static void createMissingSitePinInsts(Design design) {
         EDIFNetlist netlist = design.getNetlist();
 
@@ -3342,7 +3341,6 @@ public class DesignTools {
      * non-parent Net-s.
      * @param design Design object to be modified in-place.
      */
-    @SuppressWarnings("unchecked")
     public static void makePhysNetNamesConsistent(Design design) {
         final Map<EDIFHierNet, EDIFHierNet> netParentMap = design.getNetlist().getParentNetMap();
 


abutt-amd:
> Results:
> 
> * [Parallelize Router Preprocessing #1316](https://github.com/Xilinx/RapidWright/pull/1316): 15.3 seconds average over 3 runs, on the same DCP as [Parallelize Router Preprocessing #1316 (comment)](https://github.com/Xilinx/RapidWright/pull/1316#issuecomment-3555582048) (same setup, but it was 25s before!?!)
> * This PR: 10.9s average over 3 runs

Cool! Let me give this a try on my very large design and see how the speedup looks.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2454,29 +2477,43 @@ public static List<String> getAllRoutedSitePinsFromPhysicalPin(Cell cell, Net ne
...
     /**
-     * Creates all missing SitePinInsts in a design, except GLOBAL_USEDNET.
-     * See also {@link #createMissingSitePinInsts(Design, Net)}.
+     * Creates all missing SitePinInsts in a design, except GLOBAL_USEDNET. This method is multi-threaded based on the
+     * setting in {@link ParallelismTools}. See also {@link #createMissingSitePinInsts(Design, Net)}.
+     * It is required that the `Design` object only contains physical nets that refer to parent logical nets
+     * (use {@link #makePhysNetNamesConsistent(Design)} to make this the case).
      * @param design The current design
      */
+    @SuppressWarnings("unchecked")
```
abutt-amd: I think we can get rid of this now with the changes to how we are handling futures.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -3305,68 +3342,90 @@ public static void printSiteInstInfo(SiteInst siteInst, PrintStream ps) {
      * non-parent Net-s.
      * @param design Design object to be modified in-place.
      */
+    @SuppressWarnings("unchecked")
```
abutt-amd: Same with this one.

src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -2454,29 +2477,43 @@ public static List<String> getAllRoutedSitePinsFromPhysicalPin(Cell cell, Net ne
...
-                    // will be analyzed in due course, so skip doing so here
-                    continue;
+
+        // Compute (if needed) the physical net pin map outside of parallel loop below
+        // (Note that this also computes the parent net map used in assertions)
+        netlist.getPhysicalNetPinMap();
+
+        int numNets = design.getNets().size();
+        // Experimentally best performing number of jobs, where each job cannot have less than 100 objects
+        int numJobs = Math.min(ParallelismTools.maxParallelism() * 100, numNets / 100);
```
abutt-amd, line 2495: Does ensuring that each job cannot have less than 100 objects important? I guess it might make a slight runtime difference in some edge cases but I'm wondering if there is some other motivation.

eddieh-xlnx: I've just seen a small design with ~160 nets being split up into ~160 jobs of one net in each, which doesn't seem efficient since there is a cost of spinning up each job.

abutt-amd: Ok that makes sense to me, I wasn't really considering designs that small.

Issue #39, opened 2019-06-20
SitePinInst.getConnectedCells() produces unexpected results


jgoeders:
(v2018.3)
I'm trying to use SitePinInst.getConnectedCells() on the following design:

[top_routed.zip](https://github.com/Xilinx/RapidWright/files/3311811/top_routed.zip)

The code below prints the following:

> aes128_0/r2/t0/t3/s4/g3_b2
> aes128_0/r2/t0/t3/s4/out[2]_i_1
> aes128_0/r2/t0/t3/s4/g0_b2
> aes128_0/r2/t0/t3/s4/g1_b2
> aes128_0/r2/t0/t3/s4/g3_b2

One issue is that there appear to be duplicates listed (not a big issue).  The other issue is that I don't see how the D6 pin is connected to the listed cells.  I've included a Vivado screenshot of the design, with the Pin marked.  The cells are highlighted in Purple.  

<img width="1345" alt="vivado_pic" src="https://user-images.githubusercontent.com/1851142/59875900-3745ac80-935f-11e9-9982-6f98ff1cdf06.png">

```
import com.xilinx.rapidwright.design.Cell;
import com.xilinx.rapidwright.design.Design;
import com.xilinx.rapidwright.design.SiteInst;
import com.xilinx.rapidwright.design.SitePinInst;

class bug2 {
    public static void main(String [] args) {
        Design design = Design
                .readCheckpoint("top_routed.dcp");

        SiteInst siteInst = design.getSiteInst("SLICE_X50Y85");
        assert(siteInst != null);
        SitePinInst sitePinInst = siteInst.getSitePinInst("D6");
        for (Cell cell : sitePinInst.getConnectedCells()) {
            System.out.println(cell.getName());
        }        
    }
}
```

clavin-xlnx:
I've just [committed](https://github.com/Xilinx/RapidWright/commit/a54da72f6e7206745d979287691d1b7bebea8577) an alternative to this API, `Design.getConnectedCells(SitePinInst)` that resolves this issue.  It returns a set instead of a list to avoid duplicate situations.  I will deprecate and eventually remove the API on `SitePinInst`.  

The problem was caused by the fact that the site wire `D6` not only drives `A6` on the `D6LUT` but it also connects to pins `WA6` on all 4 of the lower LUTs (A6LUT, B6LUT, ...).  The solution is to also check the connecting cell if it has a pin mapping for that particular physical pin, if it does not, then it is likely not connected.  

[...]

PR #1327, opened 2025-11-21
RWRoute Versal SLR Crossing Test

abutt-amd:
Adds a testcase for Versal SLR crossing support. Also merged in 2025.2.0 so rwroute_versal_slr needs to be updated with 2025.2.0. Requires [RapidWrightDCP #39](https://github.com/eddieh-xlnx/RapidWrightDCP/pull/39).

### DIFF
--- a/test/src/com/xilinx/rapidwright/rwroute/TestRWRoute.java
+++ b/test/src/com/xilinx/rapidwright/rwroute/TestRWRoute.java
@@ -723,4 +723,16 @@ public class TestRWRoute {
 
         }
     }
+
+    @Test
+    public void testRWRouteVersalSLRCrossing() {
+        Path dcp = RapidWrightDCP.getPath("versal_slr_crossing.dcp");
+
+        Design design = Design.readCheckpoint(dcp);
+        design.unrouteDesign();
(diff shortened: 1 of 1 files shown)


test/src/com/xilinx/rapidwright/rwroute/TestRWRoute.java
```diff
@@ -723,4 +723,16 @@ public void testRWRouteSubstituteNetlist(@TempDir Path dir) {
 
         }
     }
+
+    @Test
+    public void testRWRouteVersalSLRCrossing() {
+        Path dcp = RapidWrightDCP.getPath("versal_slr_crossing.dcp");
+
+        Design design = Design.readCheckpoint(dcp);
+        design.unrouteDesign();
```
eddieh-xlnx, line 732: Since you're unrouting the design anyway, does it make sense to have an unrouted (and smaller) DCP as a testcase?

abutt-amd: Chris thought the design might be more generally useful (for other test cases in the future) if it was already routed.

PR #1343, opened 2026-01-05
Partitioner Feature Additions: Mapping generation, net-span partition report, LUTs/partition report, verbose hypergraph reports, LUTCount CLI, mapping constraint partition input

pnewlin-amd:
A series of QoL features and additions to the partitioner branch.

- Added hierarchical mapping output (mapping.txt), if an entire submodule `TOP/Counter` is wholly contained in partition 1 (FPGA_B) then produce only one line `TOP/Counter FPGA_B`
- Added _optional_ human readable hypergraph reports. After a partition a user can easily see which nets have been cut and their names.
- Added a dedicated LUT report .txt file output which shows how many LUTs have been assigned per partition.
- Exposed KaHyPar hypergraph switches to command line interface
- Added partition constraints, this allows the user to specify a mapping constraint which KaHyPar will interpret as a .fix input for assigning some submodule to exactly one partition.  
- Changed partition naming to be FPGA_<...>.
- Added a cells.txt output which displays every partition name. 
- Added command line interface option to specify output location of partition.
- Relocated the LUT counting functionality to a single location (source of truth)
- Added verbose logging to partition process to display caught edge cases and memory usage.
- Enabled KaHyPar verbose timing output.
- Added partition support for .DCP input.
- Added read netlist command line interface tool. Allows user to test memory usage on monolithic netlists.
- Added LUT count command line interface tool. Allows user to provide a netlist and RapidWright will return total LUT count of top level.

### DIFF
--- a/src/com/xilinx/rapidwright/debug/memory/audit/ReadNetlist.java
+++ b/src/com/xilinx/rapidwright/debug/memory/audit/ReadNetlist.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2025, Advanced Micro Devices, Inc.
+ * All rights reserved.
+ *
+ * Author: Perry Newlin
+ *
+ * This file is part of RapidWright.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.xilinx.rapidwright.debug.memory.audit;
+
+import java.io.File;
+
+import com.xilinx.rapidwright.design.Design;
+import com.xilinx.rapidwright.edif.EDIFTools;
+
+/**
+ * Memory audit utility for reading netlists.
+ */
+public class ReadNetlist {
+
+    /**
+     * Prints memory usage.
+     *
+     * @param label The label for this audit point.
+     */
+    private static void memAudit(String label) {
+        Runtime rt = Runtime.getRuntime();
+        long total = rt.totalMemory();
+        long free = rt.freeMemory();
+        long used = total - free;
+        System.out.printf("PARTITIONER DEBUG: %s -> used=%d MB total=%d MB free=%d MB%n",
+                label, used / (1024 * 1024), total / (1024 * 1024), free / (1024 * 1024));
+    }
+
+    /**
+     * Main entry point.
+     *
+     * @param args Command line arguments.
+     */
+    public static void main(String[] args) {
+        if (args.length < 1) {
+            System.err.println("Usage: ReadNetlist <netlist.[edf|dcp]>");
+            return;
+        }
+
+        File input = new File(args[0]);
+        if (!input.exists()) {
+            System.err.printf("ERROR: File not found: %s%n", args[0]);
+            return;
+        }
+
+        memAudit("readnetlist mem usg before read");
+        System.out.println("Reading netlist....");
+        Design design = null;
+        if (args[0].endsWith(".dcp")) {
+            design = Design.readCheckpoint(args[0]);
+        } else {
+            design = new Design(EDIFTools.readEdifFile(args[0]));
+        }
+        memAudit("readnetlist mem usg after read");
+
+        // kept reference to prevent GC until after waiting
+        if (design == null) {
+            System.err.println("ERROR: Failed to load design.");
+        }
+    }
+}
--- a/src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
+++ b/src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
@@ -70,6 +70,46 @@ public final class HierMappingWriter {
         Map<String, Node> children = new LinkedHashMap<>();
         Map<String, Integer> counts = new LinkedHashMap<>();
         Map<String, Integer> leafCounts = new LinkedHashMap<>();
+        String assignedPartition;
+    }
+
+    /**
+     * Recursively assigns the majority partition to each node in the trie.
+     * @param node The current node.
+     */
+    private static void assignPartitions(Node node) {
+        for (Node child : node.children.values()) {
+            assignPartitions(child);
+        }
+        node.assignedPartition = chooseHomePartition(node.counts);
+    }
+
+    /**
+     * Emits mapping lines only when the partition assignment changes from the parent.
+     * @param node The current node.
+     * @param parentPartition The partition assigned to the parent node.
+     * @param currentPath The current hierarchical path.
+     * @param out The list to collect mapping lines.
+     */
+    private static void emitOptimizedMappings(Node node, String parentPartition, String currentPath, List<String> out) {
+        // 1. Recurse to children first (Deepest-First)
+        for (Node child : node.children.values()) {
+            String childPath = currentPath.isEmpty() ? child.name : currentPath + "/" + child.name;
+            emitOptimizedMappings(child, node.assignedPartition, childPath, out);
+        }
+
+        // 2. Check for difference
+        boolean isRoot = parentPartition == null;
+        boolean isDifferent = isRoot || (node.assignedPartition != null && !node.assignedPartition.equals(parentPartition));
+
+        if (isDifferent && node.assignedPartition != null) {
+            // Emit line
+            if (!currentPath.isEmpty()) {
+                if (!logicDiscoveryPolicy.pathHasExcludedSegment(currentPath)) {
+                    out.add(currentPath + " " + node.assignedPartition);
+                }
+            }
+        }
     }
 
     /**
@@ -79,18 +119,25 @@ public final class HierMappingWriter {
      * @param partitionDir The directory containing partition files.
      * @param cellsFile The path to the cells.txt file.
      * @param defaultPartitionName The name of the default partition (optional).
+     * @param optimize Whether to use the optimized differential mapping generation.
      * @return A list of mapping lines.
      */
     public static List<String> buildMappingFromPartitionFiles(String partitionDir
             , String cellsFile
-            , String defaultPartitionName) {
+            , String defaultPartitionName
+            , boolean optimize) {
         List<String> partitions = readPartitionNames(Paths.get(cellsFile));
         Map<String, Set<String>> partToPaths = readPartitionInstanceSets(Paths.get(partitionDir),
                 new LinkedHashSet<>(partitions));
         Node root = buildTrie(partToPaths);
         computeCounts(root);
         List<String> out = new ArrayList<>();
-        selectAndEmitMappings(root, "", out, false);
+        if (optimize) {
+            assignPartitions(root);
+            emitOptimizedMappings(root, null, "", out);
+        } else {
+            selectAndEmitMappings(root, "", out, false);
+        }
 
         if (defaultPartitionName != null && !partitions.contains(defaultPartitionName)) {
             PartitionTools.logPartitionerDebug(null,
@@ -133,7 +180,8 @@ public final class HierMappingWriter {
     public static void write(Path outDir
             , EDIFNetlist netlist
             , Map<Integer, Set<String>> partitions
-            , Map<EDIFHierCellInst, Integer> instLutCountMap) {
+            , Map<EDIFHierCellInst, Integer> instLutCountMap
+            , boolean optimize) {
         // ============================================================================
         // STEP 1
         // Ensure cells.txt has fpga_* names for all partitions based on indices;
@@ -186,7 +234,12 @@ public final class HierMappingWriter {
         Node root = buildTrie(partToPaths);
         computeCounts(root);
         List<String> lines = new ArrayList<>();
-        selectAndEmitMappings(root, "", lines, false);
+        if (optimize) {
+            assignPartitions(root);
+            emitOptimizedMappings(root, null, "", lines);
+        } else {
+            selectAndEmitMappings(root, "", lines, false);
+        }
 
         // ============================================================================
         // STEP 4
@@ -217,7 +270,7 @@ public final class HierMappingWriter {
             String fullSourcePath = topName + "/" + path;
             prefixed.add(fullSourcePath + " " + destination);
         }
-        String topHomePartition = chooseHomePartition(root.counts);
+        String topHomePartition = optimize ? root.assignedPartition : chooseHomePartition(root.counts);
         if (topHomePartition != null) {
             prefixed.add(topName + " " + topHomePartition);
         }
@@ -292,20 +345,27 @@ public final class HierMappingWriter {
      *             [defaultPartitionName].
      */
     public static void main(String[] args) {
-        if (args.length < 3 || args.length > 4) {
+        if (args.length < 3 || args.length > 5) {
             PartitionTools.logPartitionerDebug(null,
                     "usage: <partitionDir> <cells.txt> <output_mapping.txt> " +
-                            "[defaultPartitionName]");
+                            "[defaultPartitionName] [--optimized]");
             return;
         }
         String partitionDir = args[0];
         String cellsFile = args[1];
         String outputFile = args[2];
-        String defaultPartitionName = args.length == 4 ? args[3] : null;
+        String defaultPartitionName = (args.length >= 4 && !args[3].startsWith("--")) ? args[3] : null;
+        boolean optimize = false;
+        for (String arg : args) {
+            if ("--optimized".equals(arg)) {
+                optimize = true;
+            }
+        }
 
         List<String> mapping = buildMappingFromPartitionFiles(partitionDir,
                 cellsFile,
-                defaultPartitionName);
+                defaultPartitionName,
+                optimize);
         writeMappingFile(mapping, outputFile);
     }
 
--- a/src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
+++ b/src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
@@ -1,0 +1,746 @@
+/*
+ *
+ * Copyright (c) 2025, Advanced Micro Devices, Inc.
+ * All rights reserved.
+ *
+ * Author: Perry Newlin
+ *
+ * This file is part of RapidWright.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.xilinx.rapidwright.edif.partition;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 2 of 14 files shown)


src/com/xilinx/rapidwright/debug/memory/audit/ReadNetlist.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.io.File;
+
+import com.xilinx.rapidwright.design.Design;
+import com.xilinx.rapidwright.edif.EDIFTools;
+
+/**
+ * Memory audit utility for reading netlists.
+ */
+public class ReadNetlist {
```
clavin-xlnx, line 33: This class seems to duplicate functionality in the [`CodePerfTracker`](https://github.com/Xilinx/RapidWright/blob/master/src/com/xilinx/rapidwright/tests/CodePerfTracker.java) class.  This class is employed when loading a DCP and, when set with the right options, can produce a similar report.  When calling `Design.readCheckpoint()` there are overload options to provide a custom configured `CodePerfTracker` instance.  Have a look to see if that would meet your needs.

I'm thinking that having something like a `DesignAnalyzer` that can provide an array of analysis options while simply loading a design would be useful, but I'm not sure this is the right PR for it.

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
```diff
@@ -0,0 +1,746 @@
...
+    /**
+     * Trie node representing a hierarchical instance path segment.
+     *
+     * - name: the segment at this level (for example, "stage_16").
+     * - children: sub-segments under this node.
+     * - leafCounts: one count per partition per explicit path)
+     * - counts: aggregated counts over this node’s subtree (leafCounts + all children)
+     *           lets us easily decide if wholly contained or not
+     */
+    private static final class Node {
```
clavin-xlnx, line 68: `Node` is an overloaded term.  Perhaps we could refactor to something more unique such as `HierNode`, or something similar.

src/com/xilinx/rapidwright/util/LUTCount.java
```diff
@@ -0,0 +1,99 @@
...
+import com.xilinx.rapidwright.edif.EDIFCell;
+import com.xilinx.rapidwright.edif.EDIFHierCellInst;
+import com.xilinx.rapidwright.edif.EDIFNetlist;
+import com.xilinx.rapidwright.edif.EDIFTools;
+import com.xilinx.rapidwright.edif.partition.PartitionTools;
+
+/**
+ * CLI utility to count logic LUT usage from EDIF (.edf/.edif) or Vivado DCP (.dcp).
+ */
+public final class LUTCount {
```
clavin-xlnx, line 40: I think the functionality of this class should be absorbed into a `DesignAnalyzer` class.  See my comment on `ReadNetlist.java`.

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
```diff
@@ -0,0 +1,746 @@
...
+     */
+    public static void writeMappingFile(List<String> lines, String outputPath) {
+        try {
+            Path out = Paths.get(outputPath);
+            Path parent = out.getParent();
+            if (parent != null) {
+                Files.createDirectories(parent);
+            }
+            Files.write(out, lines, StandardCharsets.UTF_8);
+        } catch (IOException e) { // more guards the better.
```
clavin-xlnx, line 167: Not sure what you mean by this comment?

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
```diff
@@ -0,0 +1,746 @@
...
+
+    /**
+     * RapidWright partitioner api: writes mapping.txt in outDir.
+     *
+     * @param outDir The output directory.
+     * @param netlist The EDIF netlist.
+     * @param partitions A map of partition IDs to sets of instance names.
+     * @param instLutCountMap A map of instance LUT counts (unused but kept for API consistency).
+     */
+    public static void write(Path outDir
```
clavin-xlnx, line 180: This method is confusingly named as the Javadoc says it writes the mapping file, but there is a method just above it that is called `writeMappingFile()`.

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
```diff
@@ -0,0 +1,746 @@
...
+        return partToPaths;
+    }
+
+    /**
+     * Safe read of file lines with utf-8.
+     *
+     * @param file The file to read.
+     * @return A list of lines from the file.
+     */
+    private static List<String> safeReadLines(Path file) {
```
clavin-xlnx, line 462: There is a similar method in `FileTools`: https://github.com/Xilinx/RapidWright/blob/57438b963079ad74ed29d476a25e7486c5222776/src/com/xilinx/rapidwright/util/FileTools.java#L663

Perhaps we should add an overload there to take a `Path`.

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
```diff
@@ -0,0 +1,746 @@
...
+
+    /**
+     * Case-insensitive search for a file that matches either "<name>" or "<name>.txt"
+     * in a directory.
+     *
+     * @param dir The directory to search in.
+     * @param targetName The name to search for.
+     * @return The path to the matching file, or null if not found.
+     */
+    private static Path findFileCaseInsensitive(Path dir, String targetName) {
```
clavin-xlnx, line 478: Is this method really necessary?

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
clavin-xlnx: Also, let's move it into its own file, there are several methods at the end of this file that could be refactored to live in that class.

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
```diff
@@ -0,0 +1,746 @@
...
+        return null;
+    }
+
+    /**
+     * Trims comments and whitespace; treats lines starting with '#' or '//' as comments.
+     *
+     * @param s The string to trim.
+     * @return The trimmed string, or an empty string if it was a comment.
+     */
+    private static String trimComment(String s) {
```
clavin-xlnx, line 501: This feels like maybe it should be in a more generic file, or perhaps could be replaced by stream filters.

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
```diff
@@ -0,0 +1,746 @@
...
+    }
+
+    /**
+     * Returns lines sorted by path depth, then lexicographically. Useful when post-processing
+     * is needed for readability.
+     *
+     * @param lines The list of lines to sort.
+     * @return A new sorted list of lines.
+     */
+    public static List<String> sortByPathDepth(List<String> lines) {
```
clavin-xlnx, line 773: This is not used outside the class, should we make it `private`?

src/com/xilinx/rapidwright/edif/partition/logicDiscoveryPolicy.java
```diff
@@ -0,0 +1,108 @@
...
+package com.xilinx.rapidwright.edif.partition;
+
+import com.xilinx.rapidwright.edif.EDIFHierCellInst;
+
+/**
+ * logicDiscoveryPolicy is a source-of-truth utility for
+ * discovering logic-related properties in the EDIF/DCP domain.
+ * For example, determining if a cell is a look up table.
+ */
+public final class logicDiscoveryPolicy {
```
clavin-xlnx, line 33: Class names should begin with a capital letter.

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
```diff
@@ -0,0 +1,746 @@
...
+        // append a single top 'home' line at the end for readability.
+        // ============================================================================
+        String topName = null;
+        try {
+            EDIFHierCellInst topHierCellInst = netlist.getTopHierCellInst();
+            topName = (topHierCellInst != null) ? topHierCellInst.getCellName() : null;
+            if (topName == null || topName.isEmpty()) {
+                topName = (topHierCellInst != null) ? topHierCellInst.getInst().getName() : null;
+            }
+        } catch (Exception ignored) {
```
clavin-xlnx, line 256: Having a blanket catch for all exceptions is generally frowned upon.  What is happening here that requires such usage?

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
```diff
@@ -0,0 +1,746 @@
...
+        if (prefixed.size() > 10) {
+            PartitionTools.logPartitionerDebug(null,
+                    "MAPPING_ORDER_DEBUG: post-sort ... (middle lines omitted)");
+            for (int i = prefixed.size() - showLimit; i < prefixed.size(); i++) {
+                PartitionTools.logPartitionerDebug(null,
+                        "MAPPING_ORDER_DEBUG: post-sort[%d] depth=%d line=%s",
+                        i, depthOfPath(prefixed.get(i)), prefixed.get(i));
+            }
+        }
+        java.util.Map<Integer, Integer> depthCounts = new java.util.HashMap<>();
```
clavin-xlnx, line 317: This should just be imported at the top the file rather than in context.

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
```diff
@@ -0,0 +1,746 @@
...
+                Path matched = findFileCaseInsensitive(partitionDir, p);
+                if (matched != null && Files.isRegularFile(matched)) {
+                    readOk = true;
+                    for (String line : safeReadLines(matched)) {
+                        String trimmedLine = trimComment(line);
+                        if (trimmedLine.isEmpty()) {
+                            continue;
+                        }
+                        paths.add(trimmedLine);
+                    }
```
clavin-xlnx, line 440: There is a common pattern occurring here 3 times in a row.  Perhaps we could collapse it down into a single method (or modify the existing methods that are called).  It would be great if we could make it more generic so that it could be used elsewhere.

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
```diff
@@ -0,0 +1,746 @@
...
+        cur.leafCounts.put(partition, cur.leafCounts.getOrDefault(partition, 0) + 1);
+    }
+
+    /**
+     * Splits a hierarchical path on '/' while leaving each segment intact.
+     *
+     * @param path The path to split.
+     * @return A list of path segments.
+     */
+    private static List<String> splitPath(String path) {
```
clavin-xlnx, line 566: Not sure that this deserves to be its own method.

clavin-xlnx: Also, it turns out that it is valid to have `/` as part of a name of a hierarchical cell.  Not sure if your designs will encounter that, we regularly we find this to be the case from IP from Vivado.  We do have methods that will compensate for that (see `EDIFNetlist.getHierObject()`).

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
```diff
@@ -0,0 +1,746 @@
...
+    }
+
+    /**
+     * Computes per-subtree aggregated counts for every node in the trie.
+     *
+     * @param node The root node of the subtree.
+     * @return A map of aggregated counts per partition.
+     */
+    private static Map<String, Integer> computeCounts(Node node) {
+        Map<String, Integer> aggregatedCounts = new LinkedHashMap<>();
```
clavin-xlnx, line 581: I've noticed a lot of `LinkedHashMap` usage, do you need ordered traversal in all those cases?

src/com/xilinx/rapidwright/edif/partition/logicDiscoveryPolicy.java
```diff
@@ -0,0 +1,108 @@
...
+
+    // List of protected names that should be excluded from mapping generation. 
+    // GND and VCC will be re-created locally in the partition / module as needed,
+    // GND, VCC should not be results you see in mapping.txt
+    private static final java.util.LinkedHashSet<String> EXCLUSION_NAMES =
```
clavin-xlnx, line 38: The class should be imported at the top instead of using fully qualified name in context.

src/com/xilinx/rapidwright/edif/partition/logicDiscoveryPolicy.java
```diff
@@ -0,0 +1,108 @@
...
+     * Returns true if a cell type name represents a LUT used as logic.
+     * 
+     * Treat any cell type whose name contains "LUT" as a logic LUT.
+     * substring match to be consistent with legacy policy.
+     * 
+     * TODO : More concrete & fact-checked discovery policy?
+     * @param type_name The cell type name to check
+     * @return True if the type name contains "LUT"
+     */
+    public static boolean isLogicLUT(String type_name) {
```
clavin-xlnx, line 58: Can you use `LUTTools.isCellALUT()` instead?

src/com/xilinx/rapidwright/edif/partition/LUTCountingUtils.java
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Recursively computes LUT count for hierarchical cell instance. Utilize caching 
+     * to avoid recalculating known cell types.
+     * 
+     * @param hierInst Hierarchical cell instance
+     * @param cellTypeCache Cell type cache
+     * @param instMap Instance map
+     * @return LUT count
+     */
+    public static Integer computeLUTCount(EDIFHierCellInst hierInst,
```
clavin-xlnx, line 84: We should probably add a test for this method, let's connect later about the best way to create one.

src/com/xilinx/rapidwright/edif/partition/MtKaHyParPartitioner.java
```diff
@@ -42,67 +42,111 @@ public class MtKaHyParPartitioner implements AbstractPartitioner {
...
+    /**
+     * Executes the MtKaHyPar partitioning tool with configured parameters.
+     * @return Exit code from the partitioner execution
+     */
     @Override
     public Integer runPartitioner() {
+        // If fixed vertices are present, override incompatible presets
+        String effectivePreset = presetType;
+        if (fixedVerticesFile != null) {
+            if ("deterministic".equals(effectivePreset) || "large_k".equals(effectivePreset)) {
```
clavin-xlnx, line 66: We should probably create class member `public static final String`s for the two static strings here.

src/com/xilinx/rapidwright/edif/partition/Partitioner.java
```diff
@@ -47,89 +51,777 @@
  */
 public class Partitioner {
 
+    /**
+     * Logs current memory usage for debugging.
+     * @param label Description label for the audit point
+     */
+    private static void memAudit(String label) {
```
clavin-xlnx, line 58: Again, let's try to use `CodePerfTracker` to avoid code duplication.

src/com/xilinx/rapidwright/edif/partition/Partitioner.java
```diff
@@ -47,89 +51,777 @@
...
+     */
     public static AbstractPartitioner getDefaultPartitioner() {
         MtKaHyParPartitioner p = new MtKaHyParPartitioner();
         return p;
     }
+
+    /**
+     * Prints command-line usage information to console.
+     */
+    private static void printUsage() {
```
clavin-xlnx, line 81: This has enough options that I think we should probably switch this to use the `OptionParser` class.  It will lead to a more maintainable solution and consistency over time.  You can take a look at `PerformanceExplorer` as an example.

src/com/xilinx/rapidwright/edif/partition/Partitioner.java
```diff
@@ -47,89 +51,777 @@
...
         CodePerfTracker t = new CodePerfTracker("Partitioner");
+        
+        boolean generateEdifNets = false; // Human readable hypergraph artifacts
+        Path constraintsFile = null;
+        boolean constraintsDebug = false;
+        boolean skipDetailedReports = false;
+        
+
+
+
```
clavin-xlnx, line 127: Remove extra whitespace.

src/com/xilinx/rapidwright/edif/partition/Partitioner.java
```diff
@@ -47,89 +51,777 @@
...
+        // ============================================================================
+        // ARGUMENT PARSING & INITIALIZATION
```
clavin-xlnx, line 129: It seems like `main()` is broken up into various tasks, but they are all in the same method (its over 700 lines long).  This should be broken up into separate methods.

src/com/xilinx/rapidwright/edif/partition/Partitioner.java
```diff
@@ -47,89 +51,777 @@
...
+        }
+        if (inLower.endsWith(".dcp")) {
+            PartitionTools.logPartitionerDebug(
+                    System.err,
+                    "DCP backfill start -> leafCount=%d",
+                    leafInsts.size());
+            int backfillAttempts = 0;
+            int backfillSuccess = 0;
+            int backfillSkipped = 0;
+            java.util.Map<com.xilinx.rapidwright.edif.EDIFCell, java.lang.Integer> lutCountCache;
```
clavin-xlnx, line 244: Please avoid fully qualified references and instead just import the class at the top of the file.

src/com/xilinx/rapidwright/edif/partition/PartitionLabel.java
```diff
@@ -0,0 +1,62 @@
+/*
```
clavin-xlnx, line 1: Not sure if this functionality needs its own class.  Perhaps refactor to make it a bit more generic and put it in `StringTools`?

src/com/xilinx/rapidwright/edif/partition/PartitionPairCounter.java
```diff
@@ -0,0 +1,164 @@
...
+        List<String> keys = new ArrayList<>(pairCounts.keySet());
+        Collections.sort(keys);
+        List<String> lines = new ArrayList<>(keys.size());
+        for (String pairKey : keys) {
+            int separatorIndex = pairKey.indexOf(',');
+            String partitionLabelA = pairKey.substring(0, separatorIndex);
+            String partitionLabelB = pairKey.substring(separatorIndex + 1);
+            int pairCount = pairCounts.get(pairKey);
+            
+            //format will be "FPGA_A--count--FPGA_B".
```
clavin-xlnx, line 153:
```suggestion
            // format will be "FPGA_A--count--FPGA_B".
```

clavin-xlnx (changes requested):
First round of feedback, please reach out with questions.  I didn't make comments on every occurrence, but hopefully you'll be able to apply the patterns across all the changes.

Thanks for the PR!

src/com/xilinx/rapidwright/edif/partition/Partitioner.java
pnewlin-amd: Good point. Next commit will include `PartitionSolutionProcessor.java` all of the processing and artifact generation (mapping.txt, lut_report.txt, etc.) that occurs after the KaHyPar partition solution is found will be moved into this file. Partitioner.java will utilize these functions.

Going to move netlist coarsening and edge discovery code into `partitiontools.java`.

Partitioner.java is sitting at ~400lines on my side right now, hopefully this next commit makes organization much easier to follow.

src/com/xilinx/rapidwright/edif/partition/HierMappingWriter.java
pnewlin-amd: > There is a similar method in `FileTools`:
> 
> https://github.com/Xilinx/RapidWright/blob/57438b963079ad74ed29d476a25e7486c5222776/src/com/xilinx/rapidwright/util/FileTools.java#L663
> 
> Perhaps we should add an overload there to take a `Path`.

I can include this in next commit? -->

```Java
    /**
     * This is a simple method that will read in a text file and put each line in a
     * string and put all the lines in an ArrayList.  The user is cautioned not
     * to open extremely large files with this method.
     * @param fileName Name of the text file to load.
     * @return An ArrayList containing strings of each line
     */
    public static ArrayList<String> getLinesFromTextFile(Path fileName) {
        return getLinesFromTextFile(fileName.toString());
    }
```

pnewlin-amd:
This feedback has been incorporated into a new RapidWright branch. Goal of future PR is to only have code changes that affect partitioning / `mapping.txt` output, many of the additional artifacts this branch produces are unnecessary and add far too much bloat & code.

PR #1354, opened 2026-03-21
[ArrayBuilder] Avoid failure when reading from placement file

clavin-xlnx:


abutt-amd:
This is to handle the case where no top design is provided right? If so, this change makes sense to me.

clavin-xlnx:
> This is to handle the case where no top design is provided right? If so, this change makes sense to me.

No, a top design is being supplied, but it is when a placement file is being supplied but we don't want to have automatic placements.

abutt-amd (approved):
LGTM!

PR #1360, opened 2026-04-06
Add test case for fixes to getCorrespondingTile when called on certain SLL tiles

abutt-amd:
Some Versal devices contain both SLL and SLL_1 tile name roots, even though both name roots have the same tile type (SLL). These different tile name roots have overlapping coordinates, so there can be an SLL tile and an SLL_1 tile with the same X, Y coordinates. This is a test for a change  to the rapidwright-api that allows SLL tiles to be relocated to SLL_1 tiles and vice versa, creating more opportunities for module relocation.

### DIFF
--- a/test/src/com/xilinx/rapidwright/design/TestModule.java
+++ b/test/src/com/xilinx/rapidwright/design/TestModule.java
@@ -23,11 +23,11 @@
 
 package com.xilinx.rapidwright.design;
 
+import com.xilinx.rapidwright.device.Device;
+import com.xilinx.rapidwright.device.Tile;
+import com.xilinx.rapidwright.examples.AddSubGenerator;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
-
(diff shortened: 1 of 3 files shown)


test/src/com/xilinx/rapidwright/design/TestModule.java
```diff
@@ -56,4 +56,17 @@ public void testModuleSLRRelocate() {
         Assertions.assertTrue(slrAdder.isValidPlacement(slrAdder.getAnchor(), top));
         Assertions.assertFalse(slrAdder.isValidPlacement(noSLRAdder.getAnchor(), top));
     }
+
+    @Test
+    public void testCorrespondingSLLTile() {
```
clavin-xlnx:
```suggestion
    public void testGetCorrespondingTile() {
```
This keeps the name consistent with the API we are testing.

Also, please add a quick comment to describe why we are testing the SLL vs SLL_1 case here.

clavin-xlnx (approved):

PR #1369, opened 2026-05-05
[RWRoute] On Versal, preserve node uphill of sink SPIs

eddieh-xlnx:
... uphill of `SitePinInst.getConnectedNode()` since this only gives the `*_PIN` node. We actually want to preserve its (one and only) uphill node instead -- `IMUX_*` or `BOUNCE_*`:

image

(tooltip is from hovering over the "inner" node which is the `*_PIN`)

Also:
- `Connection.toString()` to emit `<SiteName>.<PinName>` rather than just `<PinName>`
- `PartialRouter.ensureSinkRoutability()` to not `AssertionError` when asked to unpreserve a static net; print an ERROR out instead.

### DIFF
--- a/test/src/com/xilinx/rapidwright/rwroute/TestRWRoute.java
+++ b/test/src/com/xilinx/rapidwright/rwroute/TestRWRoute.java
@@ -765,7 +765,7 @@ public class TestRWRoute {
 
     @ParameterizedTest
     @ValueSource(booleans = {true, false})
-    public void testPartialRouteEnsureSinkRoutabilityStatic(boolean preserveInner) {
+    public void testPartialRouterEnsureSinkRoutabilityStatic(boolean preserveInner) {
         Design design = new Design("top", "xcv80");
 
         Net net = design.createNet("net");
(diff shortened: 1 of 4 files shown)


test/src/com/xilinx/rapidwright/rwroute/TestRWRoute.java
```diff
@@ -762,4 +762,66 @@ public void testRWRouteVersalSLRCrossingStraightIntoFlop(String srcSiteName, Str
...
+        Z_NET.addPIP(device.getNode("INT_X28Y625/OUT_EE1_E_BEG12").getAllUphillPIPs().get(0));
+        Z_NET.addPIP(device.getNode("INT_X29Y626/OUT_SS4_W_BEG6").getAllUphillPIPs().get(0));
+        Z_NET.addPIP(device.getNode("INT_X29Y626/OUT_SS1_W_BEG12").getAllUphillPIPs().get(0));
+        Z_NET.addPIP(device.getNode("INT_X30Y625/OUT_WW2_W_BEG6").getAllUphillPIPs().get(0));
+
+        Net gndNet = design.getGndNet();
+        SiteInst gndSi = design.createSiteInst("SLICE_X95Y621");
+        SitePinInst gndSpi = gndNet.createPin("HX", gndSi);
+        PartialRouter.routeDesignPartialNonTimingDriven(design, Collections.singletonList(gndSpi));
+        Assertions.assertFalse(gndSpi.isRouted());
```
eddieh-xlnx, line 834:
```suggestion

        // Make sure the ground sink is not routed, since INT_X29Y625/BOUNCE_W2
        // is now reserved for signalSpi
        Assertions.assertFalse(gndSpi.isRouted());
```

test/src/com/xilinx/rapidwright/rwroute/TestRWRoute.java
```diff
@@ -762,4 +762,69 @@ public void testRWRouteVersalSLRCrossingStraightIntoFlop(String srcSiteName, Str
...
+            Assertions.assertEquals(1, uphillNodes.size());
+            preserveNode = uphillNodes.get(0);
+        }
+        List<PIP> uphillPIPs = preserveNode.getAllUphillPIPs();
+        gndNet.addPIP(uphillPIPs.get(0));
+
+        List<SitePinInst> pinsToRoute = new ArrayList<>();
+        pinsToRoute.add(dstSpi);
+        PartialRouter.routeDesignPartialNonTimingDriven(design, pinsToRoute);
+        Assertions.assertFalse(dstSpi.isRouted());
```
eddieh-xlnx, line 796:
```suggestion

        // Check that dstSpi does not get routed because its
        // inner or outer node is preserved for the ground net.
        // Check also that an assertion doesn't fire, but this test
        // is not able to verify that an "ERROR" message gets emitted.
        Assertions.assertFalse(dstSpi.isRouted());
```

test/src/com/xilinx/rapidwright/rwroute/TestRWRoute.java
```diff
@@ -762,4 +762,74 @@ public void testRWRouteVersalSLRCrossingStraightIntoFlop(String srcSiteName, Str
...
+    @ValueSource(booleans = {true, false})
+    public void testPartialRouterEnsureSinkRoutabilityStatic(boolean preserveInner) {
+        Design design = new Design("top", "xcv80");
+
+        Net net = design.createNet("net");
+        SiteInst si = design.createSiteInst("SLICE_X95Y621");
+        SitePinInst srcSpi = net.createPin("A_O", si);
+        SitePinInst dstSpi = net.createPin("A_I", si);
+
+        Net gndNet = design.getGndNet();
```
eddieh-xlnx, line 778:
```suggestion
        // Attach either the inner or outer connected node 
        // of the dstSpi above to the gnd net
        Net gndNet = design.getGndNet();
```

clavin-xlnx (approved):
Looks good, I'll wait to hear from @abutt-amd before merging.

PR #1375, opened 2026-06-09
[RWRoute] Check and extend highest base wire for preservedMap

eddieh-xlnx:
Currently, `RWRoute.getHighestBaseWireIndex()` assumes that the first tile seen of a particular `TileTypeEnum` is representative of all tiles in that enum, and that the highest base wire is the same across all of them.

This is not the case for the `xcvp120` device, where the `XPIO_NIBBLE_SC_X121Y0` tile is encountered before `XPIO_NIBBLE_SC_1_X0Y0`, but only the latter has the `XPHY_CORE_3_RST` which has a higher (base) wire index.

Instead of updating the `Net[]` element held by `RouteNodeGraph.preservedMap` non-atomically, make element updates also atomic such that the underlying `Net[]` array can be resized should it be too small. This will certainly have a performance impact for designs with a lot of nets to preserve, which is not measured here.

Also add a critical warning for when global sinks are found to be unrouteable.

### DIFF
--- a/src/com/xilinx/rapidwright/router/VersalClockRouting.java
+++ b/src/com/xilinx/rapidwright/router/VersalClockRouting.java
@@ -674,7 +674,7 @@ public class VersalClockRouting {
                 }
             }
 
-            System.out.printf("CRITICAL WARNING: Unroutable pin " + p.getSitePinName() + " on net " + clk);
+            System.out.println("CRITICAL WARNING: Unroutable pin " + p.getSitePinName() + " on clock net " + clk);
         }
         clk.getPIPs().addAll(allPIPs);
     }
--- a/src/com/xilinx/rapidwright/router/VersalClockRouting.java
+++ b/src/com/xilinx/rapidwright/router/VersalClockRouting.java
@@ -673,6 +673,8 @@ public class VersalClockRouting {
                     q.add(new NodeWithPrevAndCost(downhill, curr, cost));
                 }
             }
+
+            System.out.printf("CRITICAL WARNING: Unroutable pin " + p.getSitePinName() + " on net " + clk);
         }
         clk.getPIPs().addAll(allPIPs);
     }
--- a/src/com/xilinx/rapidwright/rwroute/RouteNodeGraph.java
+++ b/src/com/xilinx/rapidwright/rwroute/RouteNodeGraph.java
@@ -557,25 +557,31 @@ public class RouteNodeGraph {
     }
 
     /*
-     * Return the highest base wire index across all Nodes in this tile
+     * Return the highest base wire index across all Nodes in the given tile.
+     * If its TileTypeEnum has not been seen already, compute this highest index,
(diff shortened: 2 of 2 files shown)


src/com/xilinx/rapidwright/router/VersalClockRouting.java
```diff
@@ -673,6 +673,8 @@ public static void routeNonLCBPins(Net clk, List<SitePinInst> sinks,
                     q.add(new NodeWithPrevAndCost(downhill, curr, cost));
                 }
             }
+
+            System.out.printf("CRITICAL WARNING: Unroutable pin " + p.getSitePinName() + " on net " + clk);
```
clavin-xlnx:
```suggestion
            System.out.println("CRITICAL WARNING: Unroutable pin " + p.getSitePinName() + " on clock net " + clk);
```

clavin-xlnx: Seems like we aren't using the parsing features of printf and we should have a `\n` at the end.

src/com/xilinx/rapidwright/rwroute/RouteNodeGraph.java
```diff
@@ -591,21 +597,26 @@ private Net preserve(Tile tile, int wireIndex, Net net) {
...
         // No need to synchronize access to 'nets' since collisions are not expected
         int tileAddress = tile.getUniqueAddress();
-        Net[] nets = preservedMap.get(tileAddress);
-        if (nets == null) {
-            int baseWireCount = getHighestBaseWireIndex(tile, wireIndex);
-            nets = new Net[baseWireCount];
-            if (!preservedMap.compareAndSet(tileAddress, null, nets)) {
-                // Another thread must have beat us to a compareAndSet, use that result
-                nets = preservedMap.get(tileAddress);
+        final Net[] oldNet = new Net[1];
```
clavin-xlnx, line 599: Is there a reason this needs to be an array?

eddieh-xlnx: It needs to be an array (or a Atomic object) in order to be written to from inside the lambda. I can't think of a better way to return the old value, since by the time the `updateAndGet()` (or `getAndUpdate()`) function returns the same array where the old value have been clobbered.

The alternative would be to make it a class member. However, this would not be thread-safe since many threads may be competing to use the same object as temporary storage.

clavin-xlnx (approved):

src/com/xilinx/rapidwright/rwroute/RouteNodeGraph.java
```diff
@@ -591,21 +597,26 @@ private Net preserve(Tile tile, int wireIndex, Net net) {
         // Assumes that tile/wireIndex describes the base wire on the node
         // No need to synchronize access to 'nets' since collisions are not expected
```
eddieh-xlnx:
```suggestion
```

PR #1377, opened 2026-06-10
[RWRoute] Fix Versal Clock Routing for IOB pin

clavin-xlnx:


eddieh-xlnx (approved):
Great fix -- clock routing completes instantaneously now, as I would expect. Now that you're able to *reuse* existing routing resources, does this impact the QoR on any of the other projects that do partial Versal clock routing?

clavin-xlnx:
> Great fix -- clock routing completes instantaneously now, as I would expect. Now that you're able to _reuse_ existing routing resources, does this impact the QoR on any of the other projects that do partial Versal clock routing?

I'm not sure that this fix impacts QoR substantially.

PR #1380, opened 2026-06-17
Add RegisterInitTools for reading/writing FF INIT values

andrewb1999:
Standalone utility to get/set the INIT property of a placed multi-bit flip-flop register in a post-PnR design checkpoint.

### DIFF
--- a/src/com/xilinx/rapidwright/design/tools/RegisterInitTools.java
+++ b/src/com/xilinx/rapidwright/design/tools/RegisterInitTools.java
@@ -1,0 +1,264 @@
+/*
+ *
+ * Copyright (c) 2025, Advanced Micro Devices, Inc.
+ * All rights reserved.
+ *
+ * Author: Andrew Butt, AMD Advanced Research and Development.
+ *
+ * This file is part of RapidWright.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


src/com/xilinx/rapidwright/design/tools/RegisterInitTools.java
```diff
@@ -0,0 +1,264 @@
...
+import java.math.BigInteger;
+import java.util.Arrays;
+
+/**
+ * Utility methods for updating flip-flop initial values after place and route.
+ * This is useful to change the INIT value of registers without re-running implementation.
+ */
+public class RegisterInitTools {
+
+    private static final String INIT_PROPERTY = "INIT";
```
clavin-xlnx:
```suggestion
    public static final String INIT_PROPERTY = "INIT";
```
Might be useful outside as well.

src/com/xilinx/rapidwright/design/tools/RegisterInitTools.java
```diff
@@ -0,0 +1,264 @@
...
+
+    private static Boolean getInitValue(Cell cell) {
+        if (cell == null) return null;
+        EDIFCellInst cellInst = cell.getEDIFCellInst();
+        if (cellInst == null) return null;
+        EDIFPropertyValue prop = cellInst.getProperty(INIT_PROPERTY);
+        if (prop == null) return null;
+        String value = prop.getValue();
+        if (value.contains("1'b1")) return true;
+        if (value.contains("1'b0")) return false;
```
clavin-xlnx, line 54: Are there examples of values where `1'b1` and `1'b0` have prefix or suffixes?  Just curious why we use `.contains()` here instead of `.equals()`.

andrewb1999: If I remember correctly I saw some cases where the property Vivado would set had additional quotes around it or something like that so I was basically hedging against those cases.

clavin-xlnx (approved):

PR #1395, opened 2026-07-14
Module Metadata: Implicit connections

jakobwenzel:
As discussed previously, I've seen runtimes for generating module metadata.txt files of > 5 hours. Such a simple task should not need this amount of runtime. The core loop is iterating the ports' site pins.

This PR moves generating this information from the TCL script to Java Code. We use the EDIF connectivity to figure out the ports' site pins. 

This has not yet been tested with designs containing encrypted cells. As far as i know, if the module's IO is connected to encrypted cells, we don't have enough information in RapidWright to figure out the site pins. We then still need to write the port->site pin data from TCL. I've added a new parameter to `generate_metadata` to influence this. By default, this parameter is set to `auto`. The new methodl `contains_encrypted_cells` then needs to find out whether to write connections from TCL. It is not yet implemented, as I'm not quite sure how to query Vivado for this. Probably some `get_cells -filter { IS_??? == true }` call?

I also found a rare case where Vivado does not route an output port from BEL to site pin. This leads to unsourced nets when assembling such modules into full designs. I was able to detect and fix these cases in my new code. When writing connections from TCL, the issue remains. There, we don't have enough information in Java to know which BEL should source the port. We would have to try and fix this from the TCL side. I was able to add a check to diagnose this from the Java side.

Depends on https://github.com/Xilinx/RapidWrightDCP/pull/40
Closes #1340


src/com/xilinx/rapidwright/design/MetadataParser.java
```diff
@@ -426,6 +465,156 @@ else if (nextTokens[0].equals(END)) {
...
+                                } catch (NullPointerException e) {
+                                    throw new RuntimeException("Failed to get bel pin for "+spi+". pin name: "+p+", primary name "+primarySitePinName+" at port "+currPort+" with net "+currPortNet, e);
+                                }
+                                return spi;
+                            });
+                })
+                .distinct()
+                .forEach(sitePinInst -> {
+                    currPort.addSitePinInst(sitePinInst);
+                    if (sitePinInst.getNet()!=currPortNet) {
```
clavin-xlnx:
```suggestion
                    if (sitePinInst.getNet() != currPortNet) {
```

src/com/xilinx/rapidwright/design/MetadataParser.java
```diff
@@ -426,6 +465,156 @@ else if (nextTokens[0].equals(END)) {
...
+                return;
+            }
+            if (currPort.getPassThruPortNames().stream().map(p->m.getPort(p)).anyMatch(p->!p.isOutPort())) {
+                //Passthru for an input
+                return;
+            }
+            throw new RuntimeException("no prims are driving output port "+currPort.getName());
+        }
+
+        physicalPins.stream()
```
clavin-xlnx, line 536: This statement is a bit lengthy.  Could it be simplified or sprinkle some comments in there to help guide the reader?

src/com/xilinx/rapidwright/design/MetadataParser.java
```diff
@@ -426,6 +465,156 @@ else if (nextTokens[0].equals(END)) {
...
+        if (currPort.getSitePinInsts().isEmpty() && currPort.isOutPort() && currPort.getPassThruPortNames().isEmpty()) {
+            throw new RuntimeException("in creating implicit module connections, no pin for "+currPort+" is routed to Site Pin");
+        }
+    }
+
+    /**
+     * Handle the rare case where Vivado does not route an output to a Site Pin. We have to choose a Site Pin ourselves.
+     */
+    private String bringOutputToSitePin(Cell cell, Net net, String logicalPin, String pin) {
+
```
clavin-xlnx:
```suggestion
```

tcl/rapidwright.tcl
```diff
@@ -120,7 +120,12 @@ proc write_cache_impl { dcpFile implIdx } {
     exportTest $routedDcpFile
 }
 
-proc generate_metadata { dcpFile includePaths implIdx } {
+proc contains_encrypted_cells { } {
+    #TODO!
+    return "false"
+}
```
clavin-xlnx: One way to test for this:
```
  proc contains_encrypted_cells {} {
      return [expr {
          [llength [
              get_cells -quiet -hierarchical \
                  -filter {IS_DU_WITHIN_ENVELOPE == 1}
          ]] > 0
      }]
  }

```

test/src/com/xilinx/rapidwright/design/TestMetadataParser.java
```diff
@@ -0,0 +1,72 @@
...
+import com.xilinx.rapidwright.support.RapidWrightDCP;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class TestMetadataParser {
+
+    private static WeakReference<Design> moduleDesign = null;
+
+    private static Design loadDesign() {
+        if (moduleDesign!=null) {
```
clavin-xlnx:
```suggestion
        if (moduleDesign != null) {
```

test/src/com/xilinx/rapidwright/design/TestMetadataParser.java
```diff
@@ -0,0 +1,72 @@
...
+            Design existing = moduleDesign.get();
+            if (existing!=null) {
```
clavin-xlnx:
```suggestion
            if (existing != null) {
```

src/com/xilinx/rapidwright/design/MetadataParser.java
jakobwenzel: I refactored this.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.14-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.5-bin.zip
```
clavin-xlnx: In order to update the version of Gradle we use, we have about a 3 week turn-around in order to scan and ensure compatibility with the code.  Is this fixing anything in particular?

jakobwenzel: Sorry, this was added acidentially. Reverting...

tcl/rapidwright.tcl
```diff
@@ -145,6 +155,14 @@ proc generate_metadata { dcpFile includePaths implIdx } {
         puts $md "    period [get_property PERIOD $clk]"
         puts $md "  end clock"
     }
+    if {$includeConnections == "auto"} {
+      set includeConnections [contains_encrypted_cells]
+    }
+    if {$includeConnections == "true"} {
+        puts $md "  connections explicit"
+    } {
```
clavin-xlnx:
```suggestion
    } else {
```

clavin-xlnx (approved):

tcl/rapidwright.tcl
clavin-xlnx: Apparently this is legal Tcl, but let's add the `else` just to be explicit.

jakobwenzel: Sure, let's add it

PR #1396, opened 2026-07-17
Experimental fix for random bin/rapidwright deletion

clavin-xlnx:


eddieh-xlnx:
Here's some analysis:

### The bug:

`master` sets `java.destinationDirectory.fileValue(file('bin'))` — Gradle's Java compiler output directory is `bin/`. But `bin/` also holds two git-tracked launcher scripts, `bin/rapidwright` and `bin/rapidwright.bat`. Gradle's stale-output cleanup treats everything in a registered output directory as its own to delete, so it wipes the launchers. master papered over this with a copy-out/copy-back dance: `preserveBin` → `.preserveBin/` before compile, `restoreScripts` after `mainJar`, `cleanupScripts` to tidy up.

### The fix:

Stop making `bin/` a compiler output at all. Let `compileJava` write to its default `build/classes/java/main`, then create a symlink (Linux/macOS) or junction (Windows) at `bin/com` → `build/classes/java/main/com`. `bin/` still looks like a classpath root to everything outside Gradle, but Gradle no longer owns it.

eddieh-xlnx (approved):
A question I had was: does it make sense for the symlink to be checked in, since it's always the same?

To which the answer was that `git`'s default on Windows is not to use symlinks (unlike Linux) and so it's not a good idea.

In reviewing, this came up to be aware of (but isn't blocking in my eyes):
> `exec {}` at execution time is Gradle-9-incompatible. `Project.exec` inside `doLast` was deprecated in Gradle 8.11 and removed in 9. You have a prior commit (3ad137f6a) explicitly cleaning up Gradle 9 warnings, so this reintroduces one — on Windows only, which is why it doesn't show up in your local build (I confirmed the only deprecation warning here is the pre-existing JVM-version one). Use an injected `ExecOperations`, or just new `ProcessBuilder('cmd', '/c', 'mklink', '/J', …)` and check the exit code.

eddieh-xlnx (approved):

PR #1402, opened 2026-08-12
[Timing] Fixes for unrouted and post-routed designs

eddieh-xlnx:
Fixes so that `TimingAndWirelengthReport`:
- Prints out intra-site timing edges
- Fix `--verbose` for routed DCPs (as opposed to those just routed by RWRoute) by recovering the routing from the net's PIPs
- Fix so that placed-but-unrouted DCPs do not error out
- Plus other cleanups

Added:
- `NetTools.getNodeToDriver()`
- `NetTools.getNodesToSink()`

### DIFF
--- a/src/com/xilinx/rapidwright/design/NetTools.java
+++ b/src/com/xilinx/rapidwright/design/NetTools.java
@@ -562,4 +562,58 @@ public class NetTools {
         }
         return unroutedNets;
     }
+
+    /**
+     * Builds a map from each {@link Node} driven by a PIP of the given net onto the node driving it,
+     * so that the routing reaching any one node can be recovered by walking backwards from it (see
+     * {@link #getNodesToSink(SitePinInst)}). The direction of bidirectional PIPs is taken from
+     * {@link PIP#isReversed()}. Since a node has at most one entry, a multiply-driven node (which
+     * only an illegally routed net can have) retains just the driver of the last such PIP visited;
+     * use {@link #getNodeTrees(Net)} when all drivers matter. Nodes that the net's routing merely
+     * passes through without a PIP terminating on them (e.g. the source pin's node) are absent.
+     *
+     * The net's routing is assumed to be loop-free, so that following these drivers backwards from
+     * any node terminates.
+     *
+     * @param net The net to examine; it need not be fully routed.
+     * @return A map from driven node onto driving node, empty if the net has no PIPs.
+     */
+    public static Map<Node,Node> getNodeToDriver(Net net) {
+        Map<Node, Node> nodeToDriver = new HashMap<>();
+        for (PIP pip : net.getPIPs()) {
+            if (pip.isReversed()) {
+                nodeToDriver.put(pip.getStartNode(), pip.getEndNode());
+            } else {
+                nodeToDriver.put(pip.getEndNode(), pip.getStartNode());
+            }
+        }
+        return nodeToDriver;
+    }
+
+    /**
+     * Recovers the {@link Node} instances that route the net of the given sink pin as far as that
+     * pin, by walking backwards from it through the drivers given by
+     * {@link #getNodeToDriver(Net)}. Since that map is rebuilt on every call, prefer building it
+     * once and walking it directly when more than one sink of the same net is of interest.
+     *
+     * The walk stops at the first node that no PIP drives, which is not included in the result:
+     * for a fully routed sink that is the source pin's node, but for a partially routed one it is
+     * wherever its routing gives out, which this method does not distinguish. The net's routing is
+     * assumed to be loop-free; this walk does not terminate on one that is not.
+     *
+     * @param spi The sink {@link SitePinInst} to walk back from; it must belong to a net.
+     * @return The nodes reaching that pin, ordered sink-first and excluding the node the walk
+     *         stopped at, or an empty list if nothing drives the pin (e.g. an unrouted net).
+     */
+    public static List<Node> getNodesToSink(SitePinInst spi) {
+        Map<Node, Node> nodeToDriver = getNodeToDriver(spi.getNet());
+        Node sinkNode = spi.getConnectedNode();
+        // Walk backwards, stopping at the first node without a driver: either the source node of
+        // the net, or the sink itself when this connection has not been routed to
+        List<Node> nodes = new ArrayList<>();
+        for (Node node = sinkNode, driver; (driver = nodeToDriver.get(node)) != null; node = driver) {
+            nodes.add(node);
+        }
+        return nodes;
+    }
 }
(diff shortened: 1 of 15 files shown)


src/com/xilinx/rapidwright/design/NetTools.java
```diff
@@ -562,4 +562,58 @@ public static List<Net> unrouteTopLevelNetsThatLeavePBlock(Design d, PBlock pBlo
...
+     * Builds a map from each {@link Node} driven by a PIP of the given net onto the node driving it,
+     * so that the routing reaching any one node can be recovered by walking backwards from it (see
+     * {@link #getNodesToSink(SitePinInst)}). The direction of bidirectional PIPs is taken from
+     * {@link PIP#isReversed()}. Since a node has at most one entry, a multiply-driven node (which
+     * only an illegally routed net can have) retains just the driver of the last such PIP visited;
+     * use {@link #getNodeTrees(Net)} when all drivers matter. Nodes that the net's routing merely
+     * passes through without a PIP terminating on them (e.g. the source pin's node) are absent.
+     *
+     * The net's routing is assumed to be loop-free, so that following these drivers backwards from
+     * any node terminates.
```
clavin-xlnx: Not sure that this comment is relevant for this method as the code doesn't perform any loop detection and would return a result in either scenario.

src/com/xilinx/rapidwright/design/NetTools.java
```diff
@@ -562,4 +562,58 @@ public static List<Net> unrouteTopLevelNetsThatLeavePBlock(Design d, PBlock pBlo
...
+     * for a fully routed sink that is the source pin's node, but for a partially routed one it is
+     * wherever its routing gives out, which this method does not distinguish. The net's routing is
+     * assumed to be loop-free; this walk does not terminate on one that is not.
+     *
+     * @param spi The sink {@link SitePinInst} to walk back from; it must belong to a net.
+     * @return The nodes reaching that pin, ordered sink-first and excluding the node the walk
+     *         stopped at, or an empty list if nothing drives the pin (e.g. an unrouted net).
+     */
+    public static List<Node> getNodesToSink(SitePinInst spi) {
+        Map<Node, Node> nodeToDriver = getNodeToDriver(spi.getNet());
```
clavin-xlnx, line 610: This seems expensive, especially if it gets called for static nets often.  Should there be a caching option here?

src/com/xilinx/rapidwright/design/NetTools.java
```diff
@@ -562,4 +562,58 @@ public static List<Net> unrouteTopLevelNetsThatLeavePBlock(Design d, PBlock pBlo
...
+     * The walk stops at the first node that no PIP drives, which is not included in the result:
+     * for a fully routed sink that is the source pin's node, but for a partially routed one it is
+     * wherever its routing gives out, which this method does not distinguish. The net's routing is
+     * assumed to be loop-free; this walk does not terminate on one that is not.
+     *
+     * @param spi The sink {@link SitePinInst} to walk back from; it must belong to a net.
+     * @return The nodes reaching that pin, ordered sink-first and excluding the node the walk
+     *         stopped at, or an empty list if nothing drives the pin (e.g. an unrouted net).
+     */
+    public static List<Node> getNodesToSink(SitePinInst spi) {
```
clavin-xlnx, line 605: What happens if `spi` is an output?

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -1538,7 +1539,20 @@ else if (net.getPIPs().size() > 0)
     }
     
     private void checkForIntrasiteDelay() {
```
clavin-xlnx: Does the name of this method accurately describe what it is doing?  It sounds like a method without side effects, but it appears to update `intrasiteDelay` destructively.

clavin-xlnx: Also, what happens if it gets called multiple times?

src/com/xilinx/rapidwright/timing/TimingModel.java
```diff
@@ -1549,12 +1563,54 @@ private void checkForIntrasiteDelay() {
...
             }
         }
+
+        if (ultraScaleFlopNames.contains(sinkType)) {
+            if (endPinInst != null) {
+                String sourcepin = endPinInst.getName();
+
+                if (!sourcepin.startsWith("CKEN") &&
+                        !sourcepin.startsWith("CLK1") &&
+                        !sourcepin.startsWith("CLK2") &&
```
clavin-xlnx: Could we simplify these two entries by just saying `!sourcepin.startsWith("CLK") &&`

src/com/xilinx/rapidwright/design/NetTools.java
eddieh-xlnx: Sorry, this was sloppy. That comment only applies to the following method.

src/com/xilinx/rapidwright/design/NetTools.java
eddieh-xlnx: This method is currently only called on nets belonging to the critical-path, and only in verbose mode when individual node delays are printed out. This helper method is the best we can do to pull out the path of a *single* sink on a net; the docs say:

>     * ... Since that map [getNodeToDriver(Net)] is rebuilt on every call, prefer building it
>     * once and walking it directly when more than one sink of the same net is of interest.

src/com/xilinx/rapidwright/design/NetTools.java
eddieh-xlnx: Let's throw an exception.

src/com/xilinx/rapidwright/timing/TimingModel.java
eddieh-xlnx: Sorry, this was an artifact of refactoring. I've given it a better name and javadoc now.

clavin-xlnx (approved):
Is there any chance that `accumulateIntrasiteDelays()` could be called multiple times accidentally?  How would we know other than the delay growing larger than expected?

eddieh-xlnx:
> Is there any chance that `accumulateIntrasiteDelays()` could be called multiple times accidentally? How would we know other than the delay growing larger than expected?

I can't say I understand enough of this code to say whether that is reasonable. This PR tries to minimally refactor the timing model enough to have it "work" for placed (rather than just-routed-by-RWRoute) designs.

I do agree it's not very clean code though, so I'll prompt a more *hard-working* power to see if it can be tidied up.

eddieh-xlnx:
> Is there any chance that `accumulateIntrasiteDelays()` could be called multiple times accidentally? How would we know other than the delay growing larger than expected?

Thanks for pushing on this. It turned out to be *load-bearing* ... and there was a bug related to using old data (rather than being called multiple times) that's now fixed in 78b2e29e1.

clavin-xlnx (approved):
Are we safe to rebase this on to the latest changes from `master`?

eddieh-xlnx:
> Are we safe to rebase this on to the latest changes from `master`?

Rebased. Will merge when green.

PR #1412, opened 2026-08-31
[PartialRouter] Only recover arcs between projected pins

eddieh-xlnx:
Replace the PIP walk in `addNetConnectionToRoutingTargets()` and `unpreserveNet()` with a walk of the net's node trees, creating `RouteNode`s and setting previous pointers only for those arcs lying between the nodes that connections were projected onto, and dropping `isExcludedPip()`.

Recovering arcs that Vivado has fixed in place moves to `PartialDFXRouter.setPrevOnFixedArcs()`, which walks PIPs unpruned since a locked path back from a partition pin need not lie within that span.

This means that #1404 can be (and was) reverted, and avoid other non-robustness around non-`INT` nodes that get projected away at the source and sink ends.


clavin-xlnx (approved):
I don't see any issues with the code, but perhaps a test should be added to exercise this case?  Maybe a test that sets `softPreserve=true`, unroutes a net with a routethru and invokes the PartialDFXRouter?

eddieh-xlnx:
> I don't see any issues with the code, but perhaps a test should be added to exercise this case? Maybe a test that sets `softPreserve=true`, unroutes a net with a routethru and invokes the PartialDFXRouter?

We have an internal test for `PartialDFXRouter`, granted with `softPreserve=false`. Unrouting a net and then re-routing it will typically not trigger any unpreserving since a routing solution exists without having to rip up other nets.

PR #1415, opened 2026-09-01
[EDIFParser] Key cell reference map by legal library name

sc0210:
### Problem

`EDIFParser` keys `edifInstCellMap` two different ways for the same library:

- `parseEDIFLibrary()` inserts using `library.getName()`
- `getRefEDIFCell()` looks up using the `libraryref` token read from the file, which is the library's **legal EDIF name**

For a library declared as `(library (rename work_lib "Work Library") ...)` those disagree: `getName()` returns `Work Library`, while `(libraryRef work_lib)` uses `work_lib`. The placeholder `EDIFCell` created for a forward reference is therefore never reconciled with the cell parsed later, and instances are left pointing at an empty `EDIFCell`.

This does not fail at parse time. It surfaces later as a `NullPointerException`, or as a silently incomplete netlist, which makes it awkward to trace back to the parser.

### Change

Use `cache.getLegalEDIFName(library)`, which is already what `parseEDIFCell()` uses for the cell half of the same key.

`ParallelEDIFParser` is unaffected — it keys by `EDIFLibrary` identity rather than by name.

### Testing

New `TestEDIFCellRefLinking` parses a netlist whose cross-library `(cellRef sub (libraryRef ...))` resolves to a cell with known ports, and asserts the ports are present rather than the reference having been left as an empty placeholder. It covers a plain library and a renamed one; the renamed case fails on `master` with a `NullPointerException`.

### DIFF
--- a/src/com/xilinx/rapidwright/edif/EDIFParser.java
+++ b/src/com/xilinx/rapidwright/edif/EDIFParser.java
@@ -152,8 +152,11 @@ public class EDIFParser extends AbstractEDIFParserWorker implements AutoCloseabl
         EDIFLibrary library = parseEdifLibraryHead();
 
         String currToken;
+        // Cell references name their library by its legal EDIF name, which differs
+        // from getName() whenever the library carries a (rename ...)
+        String libraryLegalName = cache.getLegalEDIFName(library);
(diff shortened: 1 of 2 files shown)


clavin-xlnx (approved):
Thanks for the PR!

Issue #1312, opened 2025-11-05
[Report Timing] intersite_delay_terms


coherent17:
Hi RapidWright Experts,

I'm running the `ReportTimingExample` with the ultrascale device, and I'd roughly go through the lightweight timing-model that RapidWright using and get really great coorelation between Vivado. However, since the method is similar to curve fitting, some parameters need to pre-config for different device. I'm wondering if these `intersite_delay_terms` file can also be on-demand downloading instead of open Vivado and dump out again?

Thanks,
Coherent

clavin-xlnx:
I would recommend parsing the timing report file to extract the critical paths for analysis in RapidWright.  That would be the fastest path to what you would like to do.  

There is some timing information stored in some DCPs, but we currently don't process it.  Perhaps in the future that might be possible.

coherent17:
Hi Chris,

I have a follow-up question regarding timing path extraction in RapidWright. I understand that the lightweight timing model is currently only available for UltraScale+ devices. However, our work now focuses on Versal FPGAs for multi-FPGA systems. In these multi-FPGA designs, timing paths reported by Vivado may not accurately represent criticality across the entire design.

Is it possible to extract timing paths from Vivado checkpoints using RapidWright, similar to Vivado's get_timing_paths -from ... -to ... command? Specifically, we need to:

- Query timing paths between specific source and destination points (pins, ports, cells or clocks)
- Extract path delay (cell delay + route delay) <-- this will be really helpful

This capability would significantly enhance our ability to analyze our work on physical synthesis before entering Vivado.

Let me know if you have any suggestion on this or alternative approach you would recommend.

Thanks,
Coherent

clavin-xlnx:
Hi Coherent,

I've made a few different attempts to move in this direction, but have not been successful yet.  The DCP does store a timing graph, but the timing numbers are not stored so they cannot be queried.  Clearly, however, calculating delay accurately in RapidWright is a desirable capability.  I will continue investigating the possibility of trying to make it possible in RapidWright, but for the short term, it is not something I can provide.  You can find/traverse and enumerate timing paths, but we simply won't be able to provide the timing information for them.

coherent17:
Hi Chris,

Thanks for the clarification. After reviewing the timing implementation in RapidWright, I see it consists of two parts: the ```TimingModel``` (for delay calculation) and the ```TimingGraph``` (for connectivity).

[...]

PR #1419, opened 2026-09-02
Add SDF parser, writer, and timing graph annotator

clavin-xlnx:
# Add SDF parser, writer, and timing graph annotator

Adds `com.xilinx.rapidwright.timing.sdf`: a reader and writer for Vivado's Standard Delay Format
output, plus an annotator that applies those delays to a `TimingGraph`.

(see discussion in #1312).

## Motivation

RapidWright's `TimingModel` is a curve-fit estimator that only ships data for UltraScale+, and a DCP
stores timing only in a proprietary binary form we can't read. So today there is no way to get real
delays for a Versal or UltraScale design. Vivado's `write_sdf` exports exactly what's needed —
per-cell `IOPATH` logic delays and per-connection `INTERCONNECT` net delays — and this reads them
back in.

```java
SdfAnnotationReport[] report = new SdfAnnotationReport[1];
TimingManager tm = SdfTools.buildTimingManagerFromSdf(design, sdfPath, null, report);

GraphPath<TimingVertex, TimingEdge> critical = tm.getTimingGraph().getMaxDelayPath();
tm.getTimingGraph().prettyPrintPathDelays(critical);
report[0].printSummary(System.out);
```

The parser and writer are also usable standalone via `SdfTools.readSdf(path)` and
`SdfTools.writeSdf(sdf, path)`, and there is a `SdfTools` entry point registered in
`MainEntrypoint`.

## Design notes

**Byte-exact round-trip.** Re-writing a parsed file reproduces Vivado's bytes exactly, quirks
included: the trailing space after `(CELL `, `(DELAY ` and `(ABSOLUTE ` but not after
`(TIMINGCHECK`, the space in `(SDFVERSION "3.0" )`, the top-level cell's deeper indentation, and the
extra space before each triple that only fast-corner files have. That fidelity is not cosmetic — it
is the cheapest strong evidence that the parser drops nothing, and it is asserted over a corpus of
real Vivado output.

**Strict grammar.** Only the subset `write_sdf` emits is accepted. Anything else throws
`SdfParseException` with `file:line:byteOffset` rather than being silently ignored. That paid off
during testing: production netlists turned out to contain `RECREM` timing checks and edge-qualified
`IOPATH` sources such as `(IOPATH (posedge PRE) Q ...)` from async-reset flip-flops, neither of
which a synthetic corpus had produced. The parser failed loudly at an exact offset instead of
quietly dropping them.

**Parallel.** A cheap byte-scan index locates line-start `(CELL` and `(INTERCONNECT` anchors, then
workers parse exact ranges. Indexing interconnects matters: the top-level cell holds every net delay
and is 50–90% of a typical file, so splitting only at cell boundaries would cap speedup near 2×.

**Device-agnostic.** `new TimingManager(design, false)` builds graph topology without the
UltraScale+-only estimator, which is what makes Versal work at all.

## Changes to existing code

Small and guarded — `useDelayModel` defaults to `true`, so the existing path is unchanged.

| File | Change |
| --- | --- |
| `TimingGraph`, `TimingManager` | Topology-only build path; `finalizeTopology()` / `finalizeTimingGraph()` for graphs extended after construction; a `deferFinalize` constructor flag |
| `TimingGraph` | Fixes `setOrderedTimingVertexLists()` appending without clearing, so it could not be re-run; adds Versal `RAMB18E5`/`RAMB36E5`; guards a latent NPE in `buildSuperGraphPaths()` |
| `VivadoTools` | Adds a `writeSdf(...)` wrapper and the `WRITE_SDF` constant |
| `MainEntrypoint` | Registers `SdfTools` |

## Validation

| Dimension | Result |
| --- | --- |
| Round-trip | Byte-identical on 23 files: Vivado **2025.2 and 2026.1**, UltraScale+ and Versal, both process corners, both `-mode` settings, plain and gzipped |
| Vivado versions | 2026.1's SDF formatting is byte-for-byte identical to 2025.2's; it emits slightly fewer cells, omitting LUT sub-cells whose outputs are unused |
| Scale | Production xcvu19p and xcvu440 netlists up to **6.65 GB / 3.3 M cells / 12 M interconnects**, parsed in ~14 s and round-tripped byte-identically |
| Accuracy | Every arc of Vivado's reported critical path matches to the picosecond |
| Coverage | 99.96% of 3.2 M graph edges annotated on a 582 k-cell production design |
| Tests | 56 tests, plus the full RWRoute suite to confirm the `TimingGraph` changes are behaviour-preserving |

The only difference in a critical-path total was 71 ps, from Vivado's own SDF collapsing
`DOUTADOUT[31:0]` into a single bus-wide arc; on Versal that arc is per-bit and the discrepancy does
not arise. The unresolved 0.04% of names are cells Vivado renamed during `phys_opt_design` that
never reached the EDIF — `SdfAnnotationReport` names them and the hierarchy level they resolved to.

Annotation is best-effort by nature, so nothing is skipped silently: every entry that could not be
applied is counted and sampled, as is every graph delay no entry supplied. Net and logic coverage
are tracked separately, because one edge can need both a net delay and a clock-to-output delay and
either can go missing alone.

## Known limitations

- Annotation is memory-bound: roughly 28 GB for a 582 k-cell design, since the DCP, the graph and
  the SDF model are all resident at once. Parsing alone is far cheaper.
- Full annotation needs a readable EDIF, so DCPs with encrypted netlists are parse and round-trip
  only.
- Byte-exactness is tied to Vivado's current formatting. A per-cell length assertion turns any
  future drift into a loud failure rather than silent corruption.


clavin-xlnx:
@coherent17 - have you been able to try this out?  What improvements or suggestions do you have?

coherent17:
I haven't tried this yet, let me find some time and see what I can do with the sdf parser branch.

PR #1420, opened 2026-09-06
[NetlistBrowser] Speed up the NetlistBrowser schematic viewer

clavin-xlnx:
### Results

  Measured with an interleaved A/B against `master` on the same machine, 1280x1024 window:

  | case | master | this branch |
  |---|---|---|
  | `optical-flow.dcp`, expand into 8,201-instance cell | 42.5 s | **8.4 s** |
  | the next expand below it | 40.6 s | **7.1 s** |
  | `picoblaze_partial.dcp` top (398 insts / 12,324 nets) | 14.5 s | **3.1 s** |
  | worst window freeze (8,201-instance expand) | 42.5 s | **0.66 s** |
  | worst window freeze (`picoblaze_partial` top) | 13.7 s | **0.64 s** |
  | mouse wheel zoom, per notch | 24.7 ms | **1.5 ms** |
  | scene graphics items (8,201-instance cell) | 555,400 | **44,981** |

  ### What changed

  **Fewer graphics items.** Qt Jambi 4.5 tears a `QGraphicsScene` down in time quadratic in the
  item count (measured: 100K items = 3.1 s, 200K = 12.3 s, 400K = 50.7 s), and nothing avoids it
  except creating fewer items. Removed the invisible duplicate "click target" item drawn next to
  almost every visible one — clicks are now resolved by searching a small box around the cursor and
  picking by an explicit priority — and each net edge is now a single `QGraphicsPathItem` polyline
  instead of two line items per segment.

  **Faster layout.** ELK's `layered` algorithm now runs with `THOROUGHNESS=1` and the greedy switch
  off on graphs over 200 nodes (38.5 s → 15.6 s on the largest test cell); smaller graphs keep the
  previous higher-quality settings. Separately, nets with more than `MAX_ROUTED_FANOUT` (100)
  connections — clocks, resets, VCC/GND — are no longer routed; their name is drawn at each pin
  instead. Such nets tie the whole graph into a single connected component, where ELK's orthogonal
  router is quadratic: leaving just `clk` and `reset` unrouted took edge routing from 7.99 s to
  0.63 s.

  **Render only what is on screen.** Above 20,000 items the scene draws only the items intersecting
  the viewport (plus a margin), with a zoom-based level of detail that drops labels and pin lines
  when they are too small to read. Below that threshold everything is drawn as before.

  **Layout off the GUI thread.** ELK reports progress in bursts — 92% of a layout is spent inside
  single steps longer than 100 ms, including one 6.3 s step — so the event loop cannot be kept alive
  from its progress monitor. Layout now runs on a worker thread while the previous schematic stays on
  screen; the status bar shows what is being laid out and <kbd>Esc</kbd> cancels. Graph construction
  stays on the GUI thread because sizing labels needs `QFontMetrics`.

  ### Behavior changes worth reviewing

  - Very high fanout nets appear as a green net name beside each pin rather than as routed wires,
    similar to how Vivado handles VCC/GND.
  - Very large cells may show slightly more edge crossings, from the ELK quality trade above.
  - A pan or zoom gesture redraws once it settles (150 ms) rather than on every step.
  - The zoom-out limit is now relative to the schematic size. Previously a fixed `zoomMin` of 0.05
    meant a large schematic — fitted at 0.001 — could not be zoomed back out far enough to see in
    full once you had zoomed in.

  ### Bugs fixed along the way

  - Net selection: an open polyline in a `QGraphicsPathItem` hit-tests as if closed, so clicks in
    open space selected an unrelated net and clicks on a wire often selected a nearby cell instead.
  - De-selecting a hierarchical cell restored a black border instead of its blue one.
  - `zoomToFit()`'s zoom clamps never took effect, because it re-fitted afterwards and undid them.

  ### Testing

  `./gradlew testJava --tests '*Netlist*' --tests '*EDIF*' --tests '*Gui*'` passes.

  Behaviour was checked with throwaway harnesses on `picoblaze_partial`, `microblazeAndILA_3pblocks`
  and `optical-flow`: click picking checked against a brute-force nearest-net ground truth (12,000
  random clicks plus clicks placed on wires, dispatched through the real `mousePressEvent`), pen
  restore over ~18K select/deselect cycles, selection surviving re-render, scrolling to an off-screen
  object, export containing the whole schematic rather than the visible part, and cancelling or
  superseding a running layout. Renderings were pixel-diffed against the previous behaviour on
  schematics below the viewport-rendering threshold.


clavin-xlnx:
Relevant comment: https://github.com/Xilinx/RapidWright/pull/1322#issuecomment-5566116837

@coherent17 - If we merge this change in now/soon, it will be included in the next release.  Are you satisfied with this version of the PR?

coherent17:
Hi @clavin-xlnx,

For the performance improvement, I'm satisfied with this version.

For future work, I would also like to:
* Add a search method to be used under certain hierarchical instances/modules.
* Add the ability to show the properties on those instances/modules.

PR #1422, opened 2026-09-09
Edif tokenizer max length config and standaloneJar task fix

coherent17:
## Summary
Parsing some EDIFs can fail with `com.xilinx.rapidwright.edif.TokenTooLongException` which exceed `EDIFTokenizer`'s ~4 MB default buffer.

This PR adds a configurable RapidWright parameter `RW_EDIF_MAX_TOKEN_LENGTH` (env var or JVM system property) instead of hardcoding a larger default. The default remains `8192 * 16 * 32`. Invalid values (non-power-of-two, non-positive) warn and fall back to the default.
**Example usage:**
```bash
export RW_EDIF_MAX_TOKEN_LENGTH=134217728 (8192 * 16 * 32 * 32)
```

Need to multiply by 32 so that the DFXed design in Versal Vivado compile can be read by RapidWright. Only found out this crash in Versal DFXed dcp when reading checkpoint dcp + readable edif which generate with Vivado `write_edif` command.

Also includes a small `standaloneJar` fix: use `runtimeClasspath` and exclude JAR signature files for runnable fat JAR builds.


build.gradle
```diff
@@ -192,12 +192,13 @@ task javadocJar(type: Jar) {
 }
 
 task standaloneJar (type: Jar) {
-    duplicatesStrategy = DuplicatesStrategy.WARN
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
```
clavin-xlnx, line 195: Is this relevant for this PR?  If you can explain what these changes are for, perhaps we can consider including them.

coherent17: Currently in master branch, the task ```standaloneJar``` to generate the standalone jar will crash with the message

```
* What went wrong:
Could not determine the dependencies of task ':standaloneJar'.
> Resolving dependency configuration 'implementation' is not allowed as it is defined as 'canBeResolved=false'.
  Instead, a resolvable ('canBeResolved=true') dependency configuration that extends 'implementation' should be resolved
```

Without this exclude change, by changing from ```configurations.implementation``` to ```configurations.runtimeClasspath``` is able to build the standaloneJar, but there's lots of noisy build warnings come from dependencies overlap. With this exclude, the build will be more clean and not that verbose.

clavin-xlnx (approved):

PR #1424, opened 2026-09-16
[DesignTools] Rework populateBlackBox() batching, net merging, and reporting

eddieh-xlnx:
`populateBlackBox()` gains a map overload that populates several black-boxes in one call, so that the shared work — inserting site instances, trimming displaced static source pins, and reconciling nets — happens once for the whole set rather than once per box. The physical work is split into two passes over the boxes so `batchRemoveSitePins()` runs between them, and no static net is trimmed after another box has merged its own static routing in.

Net merging is now restricted to the nets that actually cross a box's boundary — those whose owner sits on the other side of it. Previously every net touched by the call was run through name reconciliation, which also rewrote aliases that lived wholly inside or wholly outside the box and had nothing to do with the insertion. Callers that want those reconciled should call `makePhysNetNamesConsistent()` themselves. The crossings are merged with `makePhysNetNameConsistent()` rather than a second copy of its static-net and parent-lookup logic.

All overloads now return the set of physical nets the call modified, so a caller can route or otherwise follow up on just those nets instead of rescanning the design. Nets merged away are deleted, and so are absent from that set.

Also fixed along the way: a static boundary owner is identified by `EDIFNet.isGND()`/`isVCC()` instead of matching its name against `<const0>`/`<const1>`, which had stranded its pins on a fabricated regular net; a named cell that is not a black box now throws instead of printing to stderr and returning an empty set; and passing the same `Design` for two black boxes is rejected, since its contents are moved rather than copied.

Non-batching usage covered by `TestDesignTools.testPopulateBlackBox()`. Batched usage tests to follow.

Depends on #1409, #1411, #1425 being merged (after a new release) first.


src/com/xilinx/rapidwright/design/DesignTools.java
```diff
@@ -925,174 +926,302 @@ public static void populateBlackBox(Design design, String hierarchicalCellName,
...
      */
-    public static void populateBlackBox(Design design, String hierarchicalCellName, Design cell,
+    public static Set<Net> populateBlackBox(Design design, String hierarchicalCellName, Design cell,
+            boolean keepBoundaryRouting) {
+        return populateBlackBox(design, Collections.singletonMap(hierarchicalCellName, cell),
+                keepBoundaryRouting);
+    }
+
+    /**
+     * NOTE: This method is not fully tested. Populates a set of black boxes in a netlist with the
```
clavin-xlnx: I'm wondering if we should remove the "NOTE: This method is not fully tested." for this method and the one above as it has matured since this was originally written.

PR #1430, opened 2026-09-18
[DesignTools] populateBlackBox() detect SiteInst name clobbering

eddieh-xlnx:
As the test demonstrates, it's possible to have two `SiteInst`s with the same name (but different sites) clobber each other. This can occur, for example, if during the relocation of pre-implemented modules where the `SiteInst` names are not otherwise updated.

Detect and error out on these situations.


eddieh-xlnx:
Superseded by fix behind https://github.com/Xilinx/RapidWright/pull/1432

